PR #1, opened 2015-10-15
alter BMD and Invitro initial migrations

JoshAddington:
Alters the 0001_initial migrations for the Invitro and BMD apps to have respective foreign keys IVExperiment.dose_units and BDM_session.dose_units relate to assessment.DoseUnits instead of animal.DoseUnits. 

This fixes the migration error `django.db.utils.ProgrammingError: relation "animal_doseunits" does not exist` that is encountered when installing HAWC from scratch.


shapiromatron:
thanks @JoshAddington!

PR #2, opened 2015-12-08
Add multiple formats to in-vitro data pivots

shapiromatron:
Design two different format-styles for in-vitro data-pivots- one where data are aggregated by Endpoint, and one where data are aggregated by Endpoint-group.

### DIFF
--- a/project/invitro/exports.py
+++ b/project/invitro/exports.py
@@ -1,9 +1,28 @@
+from copy import copy
 from django.db.models.loading import get_model
 
 from utils.helper import FlatFileExporter
 
 
-class IVEndpointFlatDataPivot(FlatFileExporter):
+def getDose(ser, tag):
+    if ser[tag] != -999:
+        return ser['groups'][ser[tag]]['dose']
+    else:
+        return None
+
+
+def getDoseRange(ser):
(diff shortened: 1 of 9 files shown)


project/invitro/exports.py
```diff
@@ -1,9 +1,28 @@
+from copy import copy
 from django.db.models.loading import get_model
 
 from utils.helper import FlatFileExporter
 
 
-class IVEndpointFlatDataPivot(FlatFileExporter):
```
JoshAddington: invitro.views - line 42 calls IVEndpointFlatDataPivot, should that be changed to DataPivotEndpoint?

shapiromatron: I think it's ok; see line 25 below. I just renamed the class and pulled-out some shared helper functions

JoshAddington: https://github.com/shapiromatron/hawc/blob/master/project/invitro/views.py#L42

shapiromatron: Oh thanks; missed that one, I'll change.

JoshAddington:
I can't test, as I don't have invitro data available, but following the code looks good.  :+1:

PR #8, opened 2016-03-09
Django updates

JoshAddington:
### update deprecations
django.contrib.contenttypes.generic -> django.contrib.contenttypes.fields
reversion ->  reversion.revisions
django.db.models.loading - > django.apps.apps : 
  - get_app(app_name) -> get_app_config(app_name.)models_module
django.utils.log.NullHandler -> logging.NullHandler

Add appConfig to apps with signals


shapiromatron:
had to update the admin templates a bit, both otherwise smooth sailing...

PR #10, opened 2016-03-15
Add metric description

JoshAddington:
Add study quality metric descriptions to risk of bias edit page.


shapiromatron:
@JoshAddington slick, a template filter!

PR #11, opened 2016-03-23
Copy rob approach

JoshAddington:
#### Added
  - 'Copy Approach option to Actions dropdown menu on /study/assessment/[PK]/risk-of-bias/
  - Copy Approach page at /study/assessment/[PK]/risk-of-bias/copy/
  - Study Quality Domain List API endpoint at /study/api/domain/?assessment_id=[PK] for listing the domains and metrics of a given assessment.


shapiromatron:
@JoshAddington looks good; made a few changes to inherit from things rather than rolling your own (used `setHelper` in forms.py and `BaseTable` from hawc.js instead of extending).

Will need to port over to rob module once that's stabilized

PR #12, opened 2016-04-05
on endpoint-filter list, filter by endpoint name

shapiromatron:
Request to filter by by endpoint name, and order results by endpoint name.


JoshAddington:
Rooney had also mentioned being able to sort the results, I believe.

shapiromatron:
ahhh, right. Let me add that really quick

PR #13, opened 2016-04-28
assessment level database dump management command

shapiromatron:
Add `assessment_qs` commands for risk of bias module once the model API has stabilized.


JoshAddington:
So, I tried importing Assessment 57 into a clean DB with just the migrations run, and no Animal Endpoints are being loaded into the DB. Epi and Invitro endpoints are loaded, but not Animal.

shapiromatron:
@JoshAddington any errors?

JoshAddington:
@shapiromatron No errors when running the management command

shapiromatron:
ok, let me update this branch and resubmit

shapiromatron:
@JoshAddington ok, i've made a few tweaks, though they're unrelated to the animal model exports. Can you try again, and see if it works better this time?

JoshAddington:
Spent some time trying to hunt down what's going with the animal endpoints. They're in the .sql file that's generated, but for some reason the entries aren't being added to the animal_endpoint table. Couldn't say why, though.

Are the Risk of Bias reviews supposed to be exported as well?

shapiromatron:
Hmm, I think we're on the danger-zone of "it works on my computer". Both animal endpoints and risk of bias reviews are included in the export, and they load fine for me in the db export. 

Are you seeing any errors when you run the command `psql -d hawc3 -f ~/Desktop/foo.sql`?

JoshAddington:
If I don't run `python manage.py migrate` before the command, then the psql command runs fine, but I get errors running migrate after. If I run migrate first, then I get a bunch of
`multiple primary keys for table "xxx" not allowed`
and 
`insert or update on table "xxx" violates foreign key constraint "xxx", Key (xxx_id)=(ddd) is not present in table "xxx"`
errors

The data for both animal endpoints and risk of bias reviews are in the .sql file.

shapiromatron:
What if you don't run migrate at all?

JoshAddington:
It works if I don't migrate, but wouldn't that lock the models to the current model structure if I can't migrate?

shapiromatron:
Yes, and I think that's fine-as is. The intended goal for this is a static, immutable export, which can be used in other database systems. It'd be nice if I could also add migrations to the export, but I didn't find an introspective way to do this.

JoshAddington:
ok, then it LGTM

PR #14, opened 2016-05-05
Assets reformat

shapiromatron:
Make bundle available on all pages in HAWC. This makes it possible to use modern ES6 webpack, react, etc., anywhere in HAWC.

The implementation creates a new global variable `window.app` on a pages in HAWC; and a startup method can be called to start a new application can be added to `/project/assets/index.js`.

Also made node.js a requirement for HAWC development instance; and added instructions for setup.


JoshAddington:
Nice, this looks goood.

Looking forward to being able to sprinkle React around.

PR #15, opened 2016-05-13
More cleanup options

shapiromatron:
Added ability to clean IV chemical fields in text-cleanup


JoshAddington:
That was surprisingly simple. 👍

PR #16, opened 2016-05-13
Animal ci

shapiromatron:
Calculate confidence intervals for animal datasets in Python instead of JS; that way we can include in data pivot queries. Move the JS calculations to the form-only.


shapiromatron:
@JoshAddington Made a few more updates to this PR. Another request was to be able to sort endpoints by maximum percent change. Please accept before mtg tomorrow if possible

JoshAddington:
What's the best way to test this?

PR #18, opened 2016-06-02
Separate Risk of Bias module & add conflict resolution(final) review

JoshAddington:
- Separates Study Quality from Study module into RiskOfBias module, which includes pulling relevant JS from static/study/js/models.js into static/riskofbias/js/models.js.

- Reworks model structure for  Risk of Bias.
  - Before, each Study had many Risk of Bias reviews, one for each metric to be reviewed.
  - Now
      - the number of reviews to be completed is set at the assessment level.
      - each Study has an author(s) assigned to complete the Risk of Bias review. For each author, a single Risk of Bias review is created, which contains all metrics to be reviewed.

- Adds a required Final Risk of Bias review when reviews are assigned.
  - If the number of reviews for the assessment is 1, then the Final review is the only one assigned.
  - If the number of reviews is more than 1, then the Final review is in addition to that number. 
  - The Final review supplies data for all places in HAWC that use a study's Risk of Bias data.

Adds Views for:
 - Assessment-level Risk of Bias Reviewers list
 - Assessment-level Risk of Bias Reviewers edit
 - Study-level Risk of Bias View All reviews display
 - Study-level Final Review edit


shapiromatron:
@JoshAddington a mighty PR - nice job!

PR #24, opened 2016-06-30
Rob refactor

JoshAddington:
**1. Use react frontend for rob_edit and rob_edit_final; try to use inherited view with different api call for all results**
**2. Use same rendering method for rob_detail and rob_detail_all**
**3. Use react table for modal displays**
Visuals impacted: 
- risk of bias heatmap (full study, cross study metric, and single metric displays)
- data pivot (study detail)

**4. Split study api to use either all or final-only `riskofbiases`**
**5. Remove `qualities` from api**


shapiromatron:
Looks great. One more request before merge:

- If person viewing is not team-member or higher, can the reviewer name not be shown in final review? Example page `/rob/study/3297/` (or in modals)

PR #28, opened 2016-08-10
Rob score cleanup

JoshAddington:
Feature:
======
Feature is accessible at url `assessment/<assessment id>/clean-study-metrics` and linked through Action menu from Assessment detail page and Risk of Bias Requirements page.

Allows editing of a risk of bias review metric across all studies for an assessment. 

Minor edits:
========
Django
-------
 - Moves assessment/api/views.py to assessment/api.py to bring inline with all other apps.

React
------
 - Creates assets/shared/utils/helpers.js to hold most js helpers. React apps then have own utils/helpers.js that add more selective helpers.
 - Adds Select and ArraySelect components to assets/shared.
 - Makes TextCleanup app more reuseable by replacing Endpoint naming with Item.
 - Passes startup functions mounting elements instead of hard coding.
 - Moves .bind(this) on functions


JoshAddington:
Visible items is nice, :+1:

PR #38, opened 2016-09-12
Rob metric update

shapiromatron:
Apologies in advance for reviewer; slightly disorganized PR for mostly RoB changes.

**In RoB heatmap:**
- NR is now yellow, 
- NR and NA can optionally be added/removed from legend

**In RoB barchart:**
- NR is aggregated w/ - 
- NR and NA can optionally be added/removed from legend

**To make these changes, some refactoring was required:**
- RoBLegend was created (was previously an overloaded method on RoBHeatmap)
- Instead of using a d3 linear scale for setting color, we use direct assignment using a object key/value

Could you please review all RoB visualizations to ensure they no edge-cases are broken?


JoshAddington:
> **In RoB barchart:**
> ...
 > - **_NR_** and NA can optionally be added/removed from legend

I'm guessing that you removed the option to add/remove NR from the legend base on it being aggregated with -. Correct?

shapiromatron:
Correct!

JoshAddington:
OK, all the visuals still work as expected. LGTM

PR #40, opened 2016-09-27
added makefile and command for auto-building documentation

shapiromatron:
Added new command `make servedocs` which allows the documentation to be rebuilt and automatically regenerated while source-files change. This should make it easier to write documentation and test that formatting is appropriate.

Also added @JoshAddington to the list of HAWC developers.


JoshAddington:
Everything looks good to me.

PR #41, opened 2016-09-27
Lit refactor

shapiromatron:
Move the literature API/import scripts to a separate package, https://github.com/shapiromatron/litter_getter. 

This should make testing and updating the import utilities easier, and helps to reduce the tight-coupling of the HAWC API w/ other third-party tools


JoshAddington:
Looks like everything is working. :+1:

PR #50, opened 2017-02-10
Copy as new for animal models

shapiromatron:
Added copy as new for animal experiment and animal-group models, per trello card: https://trello.com/c/S4jXMPVY/540-copy-as-new-animal-experiment-and-animal-groups.

Also created a view mixin and applied for all standard copy as new models; investigated implementing for data-pivot and search but determined that they have custom implementations that are too different for this generic mixin.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,5 @@ node_modules
 bundles/
 webpack-stats.json
 .happypack/
+
+bin/*local.sh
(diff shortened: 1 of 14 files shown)


shapiromatron:
@JoshAddington confirmed- your fix works for both cases, and it was needed; I must have only tested the case where template_name was specified - my bad 🤕

PR #53, opened 2017-02-23
New Epi Exposure fields

JoshAddington:
Add new optional, numeric fields for epi exposure metrics [example](https://hawcproject.org/epi/exposure/1350/)
 
- Adds the following optional float fields: N, estimate, variance, lower CI, upper CI, lower range, upper range.
 - Adds the following pull-down lists: “central estimate type”, “variance type”. Options will be identical to epi-result options 
- Fields exported in Excel export
- Fields presented in exposure-detail view if not null

Not sure if lower and upper range need help text and/or what that should be.

https://trello.com/c/rI5dmE9A/551-add-new-optional-numeric-fields-for-epi-exposure-metrics


JoshAddington:
@shapiromatron Should bound_intervals be added to Exposure as well?

shapiromatron:
Yes, might be good to add. Then we can present 3 new columns on the data pivot "exposure estimate", "exposure lower", "exposure upper", or whatever vocabulary is consistent with the result-implementation.

project/epi/models.py
```diff
@@ -867,18 +867,18 @@ def flat_complete_header_row():
...
+            "exposure-variance_type",
             "exposure-lower_ci",
             "exposure-upper_ci",
             "exposure-lower_range",
             "exposure-upper_range",
             "exposure-lower_bound_interval",
             "exposure-upper_bound_interval",
-            "exposure-estimate_type",
-            "exposure-variance_type",
+            "exposure-exposure_distribution",
```
shapiromatron: @JoshAddington had to move `exposure-exposure_distribution` column as the current implementation produced an excel file where data and headers were unsynced.

project/epi/serializers.py
```diff
@@ -81,6 +81,8 @@ class Meta:
 class SimpleExposureSerializer(serializers.ModelSerializer):
     url = serializers.CharField(source='get_absolute_url', read_only=True)
     metric_units = DoseUnitsSerializer()
+    variance_type = serializers.CharField(source='get_variance_type_display', read_only=True)
+    estimate_type = serializers.CharField(source='get_estimate_type_display', read_only=True)
```
shapiromatron: @JoshAddington Had to add these fields to the `SimpleExposureSerializer`, since they were being reported as integer values in Excel export

PR #57, opened 2017-04-24
Copy data across assessments with management command

shapiromatron:
Copy data (bioassay, epi, or in-vitro) and optionally risk of bias across assays. Wrapped in a single transaction in case of failure. If RoB is copied, then metrics/domains are only created if they do not already existing in the target assessment.


shapiromatron:
@JoshAddington I'm not seeing the experiments copy over now, did I inadvertently break this?

JoshAddington:
@shapiromatron Yep, that's what caused the problem.

JoshAddington:
Maybe Metric.metric should be renamed Metric.name, not necessarily in this PR. Most models use name where Metric is using metric. Might be a good idea for consistency.

shapiromatron:
Agreed re: renaming `Metric.metric` to `Metric.name`; I'll add a trello card.

PR #59, opened 2017-05-09
add epimeta copy_across_assessments

JoshAddington:
Closes https://trello.com/c/AZxPhUKD/572-complete-epi-meta-copy-across-assessments

Adds epimeta `copy_across_assessments` ability.

The meta-ness of the data makes the process different from the rest of the studies. The MetaResult has many SingleResults, which link to other studies. The SingleResult.study might not be in the current assessment, so we have to go through SingleResults of all the assessment's Study.MetaProtocol.MetaResults, as well as the SingleResults that link directly to the assessment's Studies.

 If the `single_result.study` is being copied over, then it's in the `cw` crosswalk dict and the new id is assigned, otherwise it's not in the crosswalk dict and stays the same. Same with the `single_result.meta_result`.

### DIFF
--- a/project/study/models.py
+++ b/project/study/models.py
@@ -5,7 +5,6 @@ import collections
 import itertools
 
 from django.db import models, transaction
-from django.db.models import Q
 from django.apps import apps
 from django.core.exceptions import (ValidationError, ObjectDoesNotExist,
                                     MultipleObjectsReturned)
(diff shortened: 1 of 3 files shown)


shapiromatron:
@JoshAddington The SingleResult might not be in the current assessment? Are there actual cases in the database where this occurs?  If so, we should fix this. There shouldn't be any cross-assessment dependencies.

project/study/models.py
```diff
@@ -163,6 +164,14 @@ def copy_across_assessment(cls, studies, assessment):
             for child in children:
                 child.copy_across_assessments(cw)
 
+        # epi-meta single_results contain cross-study data, so we need to finish
+        # copying studies before copying single results.
+        if studies[0].epi_meta:
```
shapiromatron: this only checks if the first study in the list is ``epi_meta`` right? what if the first study isn't epi meta but the second is?

PR #60, opened 2017-05-11
remove ReportTemplate

JoshAddington:
Closes https://trello.com/c/nXyml8h0/573-remove-reporttemplate-from-assessment-and-any-related-libraries

Removes ReportTemplate and GenerateReport (which uses ReportTemplate).


JoshAddington:
@shapiromatron Do we also need to remove the HAWCDocxReport and related classes? That has a dependency on shapiromatron/docxUtils

shapiromatron:
Yes, let's get rid of all Word functionality since none of it is currently exposed on the public HAWC website.

PR #61, opened 2017-05-11
Remove old required environment variables

JoshAddington:
Closes https://trello.com/c/rMXB0PuT/575-remove-old-required-environment-variables-for-hawc
I cloned this to a new folder and setup a new virtualenv to test out.

Removes old environment variables from hawc/settings/  and docs, and moves notes about adding env vars from dev_setup to server_setup in docs.

### DIFF
--- a/docs/source/server_setup.rst
+++ b/docs/source/server_setup.rst
@@ -33,10 +33,9 @@ Required environment variables
 ------------------------------
 
 The following environment variables will need to be set when running HAWC in
-production;
-For linux, you can update your virtual-environment settings in ``$VIRTUAL_ENV/bin/postactivate``::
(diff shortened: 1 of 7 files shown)


docs/source/server_setup.rst
```diff
@@ -33,14 +33,18 @@ Required environment variables
 ------------------------------
 
 The following environment variables will need to be set when running HAWC in
-production; example values are shown for each:
+production.
+ - For linux, you can set these values in ``$VIRTUAL_ENV/bin/postactivate``.
```
shapiromatron: ``postactivate`` only exists when using virtualenvwrapper I think; i think it should just be ``activate``.

PR #62, opened 2017-05-16
2to3

JoshAddington:
Closes https://trello.com/c/UbSrtK6S/574-migrate-hawc-to-python3
 
- Run 2to3 to update python 2 code to python 3.
 - Replace model `__unicode__` methods with `__str__` methods, per https://docs.djangoproject.com/en/1.11/topics/python3/#str-and-unicode-methods
 - Update pydot, bmds, and RISparser to Python 3 compatible versions.
 - Use empty super() calls in child classes.
 - Remove unused code from utils module.


shapiromatron:
wohoo! thanks @JoshAddington, I'll start taking a look!

shapiromatron:
litter_getter fails and needs to be converted; celery also has issues w/ python 3.6. I'm working through litter_getter but not sure what to do w/ celery - perhaps wait until they release a version compatible w/ v3.6?

JoshAddington:
Since litter_getter is the only place that RISparser is used, shouldn't the installation for that package be handling RISparser installation instead of hawc?

shapiromatron:
Yes, good idea.

PR #89, opened 2017-12-06
APPDEV-3840 outline overall confidence row in RoB heat map

jaybuie-cgi:
Complete set of code for outlining the relevant RoB heat map rows.


jaybuie-cgi:
Never mind.  I put this request in the wrong place.

PR #91, opened 2017-12-14
Fix for divide by zero error on About page

shanethacker:
Probably not something y'all would have run into before, but we were getting an error on the About page due to not having some of the data types.

### DIFF
--- a/project/assessment/views.py
+++ b/project/assessment/views.py
@@ -23,6 +23,14 @@ from utils.views import (MessageMixin, LoginRequiredMixin, BaseCreate,
 
 from . import forms, models, tasks
 
+# Checking for denominator counts that are zero when dividing.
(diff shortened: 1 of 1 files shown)


shapiromatron:
Yay, first PR, welcome! Please rename the function and use try/catch, as below, to follow [EAFP](https://docs.python.org/3/glossary.html?highlight=eafp#term-eafp).

```python
def percentage(numerator, denominator):
    # Checking for denominator counts that are zero when dividing.
    try:
        return numerator / float(denominator)
    except ZeroDivisionError:
        return 0
```

shanethacker:
Just recommitted with the suggested changes.

shapiromatron:
@shanethacker accepted! Thanks all!

PR #126, opened 2018-03-30
Added BMD to address issue #125

### LINKED ISSUE #125: Enable BMD as descriptive value for endpoint-group data pivot
BMD and BMDL are currently available as descriptive text columns in animal bioassay data pivots, but only if the pivot is set up with an export style of One row per endpoint/result. We need to enable them for One row per endpoint-group/result-group as well.

shanethacker:
Adds BMD and BMDL as descriptive text to one row per endpoint-group data pivot export types. Fixes #125

### DIFF
--- a/project/animal/exports.py
+++ b/project/animal/exports.py
@@ -146,6 +146,16 @@ class EndpointGroupFlatDataPivot(FlatFileExporter):
             return '|{0}|'.format('|'.join(effs))
         return ''
 
+    @staticmethod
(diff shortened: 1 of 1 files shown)


shanethacker:
No problem!

shapiromatron:
Closing based on discussions offline; users were unable to give a specific example of how they'd use this feature. We'll close but not delete the branch should a use be specified in the future.

PR #132, opened 2018-06-07
new Chemspider API

shapiromatron:
The chemspider API that we currently use is being deprecated; this PR uses the new APIs and gives the same results.

To test this, I got unit-test working again using [py.test](https://docs.pytest.org/en/latest/). There's a test in there that worked both before and after the change.

To use the new PR, you'll need to request a new API token, which goes in the same spot in the settings as the old API token. https://developer.rsc.org/


shapiromatron:
@shanethacker fyi; when you merge into epa hawc you'll need a new chemspider token.

PR #140, opened 2018-11-02
when hovering, add points on crossview

shapiromatron:
Add points dose-response points on crossview when hovering to display mean response values for each dose group. This was requested to better understand the spline curve-fit. The card did not specify to only show points on hover, but after attempting to display all points at all times, it felt too overwhelming and therefore was modified for the hover-only case.

screen shot 2018-11-01 at 21 36 14

Not extensively tested on IE11 or the svg/png rendering upon download; I don't anticipate issues, but it will be good to check after merging and issue hotfix if required (difficult to test locally).

### DIFF
--- a/project/summary/assets/summary/CrossviewPlot.js
+++ b/project/summary/assets/summary/CrossviewPlot.js
@@ -737,7 +737,8 @@ class CrossviewPlot extends D3Visualization {
     draw_visualization() {
         var x = this.x_scale,
             y = this.y_scale,
-            self = this;
+            self = this,
+            paths;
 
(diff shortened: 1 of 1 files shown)


shapiromatron:
@shanethacker I finally was able to look into this; it looks like HAWC prime doesn't have an endpoint_name filter, so that wasn't something I can't replicate. I think it was added here, but it looks correct to me in this commit: https://github.com/teamhero/hawc/commit/de34e26aa4928b25a2e6102aff0714256c183c93#diff-766ee7866f7d1796df41a2caa5e5538c

PR #143, opened 2019-04-30
Fix broken link

nlewycky:
Make the link absolute. As written, this is a relative path so if you're reading https://hawc.readthedocs.io/en/latest/ then it tries to load https://hawc.readthedocs.io/en/latest/hawcproject.org which is 404.


shapiromatron:
Thanks @nlewycky!

Issue #144, opened 2019-08-13
contact us - login-required


shapiromatron:
To prevent spam to the contact-us email, make login required. 

In the future, after codebases have been merged, we could do something more advanced such as adding a captcha, but this should be sufficient to prevent gross spam I've been receiving lately.

PR #145, opened 2019-08-13
make login required to contact; capture prior page

shapiromatron:
Implements #144; added a little more magic to the email so that it reports the prior page the client was currently visiting.


shapiromatron:
@shanethacker fyi; added this one to get the spam out of my inbox hopefully...

PR #182, opened 2020-03-12
Python client

rabstejnek:
Create a REST client in Python.

- [x] Assessment downloads
  - [x] Create
    - [x] Literature review
    - [x] Study evaluation report
    - [x] Animal bioassay data
    - [x] Epidemiology data
    - [x] Epidemiology meta-analysis data
    - [x] In-vitro data
    - [x] Visualizations
  - [x] Purge Legacy
    - [x] Literature review
    - [x] Study evaluation report
    - [x] Animal bioassay data
    - [x] Epidemiology data
    - [x] Epidemiology meta-analysis data
    - [x] In-vitro data
    - [x] Visualizations

### DIFF
--- a/hawc/apps/invitro/api.py
+++ b/hawc/apps/invitro/api.py
@@ -13,9 +13,7 @@ from ..common.views import AssessmentPermissionsMixin
 from . import exports, models, serializers
 
 
-class IVAssessmentViewset(
-    AssessmentPermissionsMixin, LegacyAssessmentAdapterMixin, viewsets.GenericViewSet
-):
+class IVAssessmentViewset(AssessmentPermissionsMixin, LegacyAssessmentAdapterMixin, viewsets.GenericViewSet):
     parent_model = Assessment
     model = models.IVEndpoint
     permission_classes = (AssessmentLevelPermissions,)
@@ -26,15 +24,12 @@ class IVAssessmentViewset(
             return self.model.objects.published(self.assessment)
         return self.model.objects.get_qs(self.assessment)
 
-    @decorators.detail_route(
-        methods=("get",), url_path="full-export", renderer_classes=PandasRenderers
-    )
+    @decorators.detail_route(methods=("get",), url_path="full-export", renderer_classes=PandasRenderers)
     def full_export(self, request, pk):
         self.set_legacy_attr(pk)
         self.object_list = self.get_queryset()
-        export_format = request.GET.get("output", "excel")
         exporter = exports.DataPivotEndpoint(
-            self.object_list, export_format=export_format, filename=f"{self.assessment}-invitro",
+            self.object_list, export_format="excel", filename=f"{self.assessment}-invitro",
         )
         return Response(exporter.build_dataframe())
 
--- a/hawc/apps/invitro/api.py
+++ b/hawc/apps/invitro/api.py
@@ -13,7 +13,9 @@ from ..common.views import AssessmentPermissionsMixin
 from . import exports, models, serializers
 
 
-class IVAssessmentViewset(AssessmentPermissionsMixin, LegacyAssessmentAdapterMixin, viewsets.GenericViewSet):
+class IVAssessmentViewset(
+    AssessmentPermissionsMixin, LegacyAssessmentAdapterMixin, viewsets.GenericViewSet
+):
     parent_model = Assessment
     model = models.IVEndpoint
     permission_classes = (AssessmentLevelPermissions,)
@@ -24,7 +26,9 @@ class IVAssessmentViewset(AssessmentPermissionsMixin, LegacyAssessmentAdapterMix
             return self.model.objects.published(self.assessment)
         return self.model.objects.get_qs(self.assessment)
 
-    @decorators.detail_route(methods=("get",), url_path="full-export", renderer_classes=PandasRenderers)
+    @decorators.detail_route(
+        methods=("get",), url_path="full-export", renderer_classes=PandasRenderers
+    )
     def full_export(self, request, pk):
         self.set_legacy_attr(pk)
         self.object_list = self.get_queryset()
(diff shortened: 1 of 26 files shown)


rabstejnek:
@shapiromatron Ready for review

hawc/client/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+            "search_type": "i",
+            "source": 2,
+            "title": title,
+            "description": description,
+            "search_string": ",".join(str(id_) for id_ in ids),
+        }
+        url = f"{self.root_url}/lit/api/search/"
+        return self.post(url, payload)
+
+    def lit_tags(self, assessment_id: int) -> pd.DataFrame:
```
shapiromatron: does this (and others below) return a dataframe, or does it return a dict, from response.json()?

hawc/client/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+        Returns:
+            Session: a requests.Session object
+        """
+        session = Session()
+
+        url = f"{self.root_url}/user/api/token-auth/"
+        data = {"username": email, "password": password}
+        response = session.post(url, json=data)
+        token = self._handle_hawc_response(response)["token"]
+        print(response)
```
shapiromatron: remove print

hawc/apps/lit/urls.py
```diff
@@ -50,11 +50,6 @@
         views.RefsByTagJSON.as_view(),
         name="refs_json",
     ),
-    url(
-        r"^assessment/(?P<pk>\d+)/references/download/$",
-        views.RefDownloadExcel.as_view(),
-        name="ref_download_excel",
```
shapiromatron: looks like this is also in `reference_list.html` and `reference_visual.html` (ctrl-shift-f for ref_download_excel in vscode)

hawc/apps/lit/views.py
```diff
@@ -76,69 +76,6 @@ def get_form_kwargs(self):
         return kwargs
 
 
-class RefDownloadExcel(BaseList):
```
shapiromatron: good riddance!

hawc/client/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+        session.headers.update(Authorization=f"Token {token}")
+        return session
+
+    def get(self, url: str) -> Optional[Dict]:
```
shapiromatron: perhaps create a "private" method to `_format_to_df()` which takes the response we'll get from all of these and returns a data frame.

Also, these two methods are private, aka they shouldn't be exposed to our public api, so prefix them with a single underscore, eg., `_get()`, `_post()`, this is a pretty standard python convention for "private" stuff

hawc/apps/lit/api.py
```diff
@@ -55,6 +56,27 @@ def reference_tags(self, request, pk):
...
+        methods=("get", "post"), url_path="references-download", renderer_classes=PandasRenderers
+    )
+    def references_download(self, request, pk):
+        """
+        Get all references in an assessment.
+        """
+
+        self.set_legacy_attr(pk)
+
+        self.tags = models.ReferenceFilterTag.get_all_tags(self.assessment.id, json_encode=False)
```
shapiromatron: we can just set this to a local variable; no need to bind to the class (change `self.tags =` to `tags = `)

hawc/apps/invitro/api.py
```diff
@@ -1,6 +1,42 @@
...
+            return self.model.objects.published(self.assessment)
+        return self.model.objects.get_qs(self.assessment)
+
+    @decorators.detail_route(
+        methods=("get",), url_path="full-export", renderer_classes=PandasRenderers
+    )
+    def full_export(self, request, pk):
+        self.set_legacy_attr(pk)
+        self.object_list = self.get_queryset()
+        export_format = request.GET.get("output", "excel")
```
shapiromatron: i don't think we need this; our renderer handles the export_format

hawc/client/__init__.py
rabstejnek: All functions now return a dataframe.

hawc/client/__init__.py
rabstejnek: Done

hawc/apps/lit/urls.py
rabstejnek: Updated reference_list.html and reference_visual.html to use new endpoint.

hawc/client/__init__.py
rabstejnek: get and post methods were made private. I managed to use the dataframe constructor directly on the json, so it did not warrant a separate helper method.

hawc/apps/lit/api.py
rabstejnek: Done

hawc/apps/invitro/api.py
rabstejnek: Understood, I just hardwired 'excel' into the export_format parameter.

rabstejnek:
@shapiromatron Requested changes made and submitted.

hawc/apps/lit/api.py
```diff
@@ -55,6 +56,27 @@ def reference_tags(self, request, pk):
         df = models.ReferenceTags.objects.as_dataframe(instance.id)
         return Response(df)
 
+    @decorators.detail_route(
+        methods=("get", "post"), url_path="references-download", renderer_classes=PandasRenderers
```
shapiromatron: this should just be get, no need to have post as well

shapiromatron: @rabstejnek after making this change, LGTM! Please make this change and then squash+merge. Congrats on your first PR 🎊

rabstejnek: Made the changes, will squash+merge now!

PR #183, opened 2020-03-13
histogram literature by year

shapiromatron:


hawc/apps/lit/api.py
```diff
@@ -55,6 +56,19 @@ def reference_tags(self, request, pk):
         df = models.ReferenceTags.objects.as_dataframe(instance.id)
         return Response(df)
 
+    @decorators.detail_route(methods=("get",))
```
shapiromatron: @rabstejnek this is where histogram code goes

PR #190, opened 2020-03-27
Python client additions

rabstejnek:
- [x] API endpoint and client function for getting public assessments
- [x] Client function for crawling animal endpoints for given assessment id
- [x] Client function for crawling epi endpoints for given assessment id
- [x] Decouple client instantiation and authentication
- [x] Restructure client into its own package with separate setup.py


docs/source/api.rst
```diff
@@ -25,21 +25,12 @@ There's also a HAWC client available:
...
+    client.authenticate(username="username",password="password")
 
-    hero_ids = list(range(5000, 5020))
-    client.lit_import_hero(
-        assessment_id=123,
-        title="example title",
-        description="example description",
-        ids=hero_ids
-    )
+    client.animal.endpoints(assessment_id=123)
```
shapiromatron: let's use `lit.references()`  instead, that will always be a snappy response.

maybe demo both a read and a write:

```python
# as an example, get all references for an assessment
client.lit.references(assessment_id=123)

# as an example, import new references to an assessment
client.lit_import_hero(
    assessment_id=123,
    title="example title",
    description="example description",
    ids=[5000, 5010]
)
```

hawc/apps/assessment/api.py
```diff
@@ -165,6 +165,14 @@ class Assessment(AssessmentViewset):
     serializer_class = serializers.AssessmentSerializer
 
 
+class PublicAssessment(viewsets.ReadOnlyModelViewSet):
```
shapiromatron: add a new `@action(detail=False, methods=('get',))` view to `AssessmentViewset` instead of creating a new viewset

hawc/apps/assessment/serializers.py
```diff
@@ -22,6 +22,12 @@ class Meta:
         fields = "__all__"
 
 
+class PublicAssessmentSerializer(serializers.ModelSerializer):
```
shapiromatron: let's use the existing `AssessmentSerializer` serializer instead of creating a new one

hawc/apps/assessment/urls.py
```diff
@@ -6,6 +6,7 @@
 
 router = DefaultRouter()
 router.register(r"assessment", api.Assessment, basename="assessment")
+router.register(r"public", api.PublicAssessment, basename="public")
```
shapiromatron: we can remove this too since we're using the other assessment

hawc/apps/assessment/api.py
```diff
@@ -165,6 +165,14 @@ class Assessment(AssessmentViewset):
...
+    model = models.Assessment
+    serializer_class = serializers.PublicAssessmentSerializer
+
+    def get_queryset(self):
+        return self.model.objects.get_public_assessments()
```
shapiromatron: we can add the `get_queryset` method here into the new action view we define above

client/hawc_client/__init__.py
```diff
@@ -0,0 +1,306 @@
...
+    pass
+
+
+class HawcServerException(Exception):
+    """An exception occurred on the HAWC server."""
+
+    pass
+
+
+class _Session:
```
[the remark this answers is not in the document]
shapiromatron: Also, for class names, private/public hawc conventions are a little different. I'd remove the leading underscore from class names, and if you want to make them more private, you can add an `__all__`: https://stackoverflow.com/questions/44834/

client/hawc_client/__init__.py
```diff
@@ -0,0 +1,306 @@
...
+            password (str): password to authenticate
+        """
+
+        url = f"{self.root_url}/user/api/token-auth/"
+        data = {"username": email, "password": password}
+        response = self._session.post(url, json=data)
+        token = self._handle_hawc_response(response)["token"]
+        self._session.headers.update(Authorization=f"Token {token}")
+
+    def _page_gen(self, url: str, params: Dict = None) -> Generator:
```
shapiromatron: very nice method, change name to `iter_pages`?

client/hawc_client/__init__.py
```diff
@@ -0,0 +1,306 @@
...
+            Dict: The JSON response, if exists
+        """
+        if response.status_code >= 400 and response.status_code < 500:
+            raise HawcClientException(response.status_code, response.json())
+        elif response.status_code == 500:
+            raise HawcServerException(response)
+        else:
+            return response.json()
+
+    def _get(self, url: str, params: Dict = None) -> Dict:
```
shapiromatron: since we refactored this class, these methods are now called by other external classes, so these should be public not private; change `_get` to `get`; same with other methods

docs/source/api.rst
```diff
@@ -25,21 +25,12 @@ There's also a HAWC client available:
```
shapiromatron: can you also add a new subsection in `developer.rst` that documents the command on how to build the client?

it should be something like:

```bash
python client/setup.py sdist bdist_wheel
```

client/hawc_client/__init__.py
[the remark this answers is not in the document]
rabstejnek: Specified "private" classes in __all__, removed leading underscores, and renamed session class.

hawc/apps/assessment/api.py
rabstejnek: `AssessmentViewset` acts a parent class for several other viewsets, so I added this `@action` to the `Assessment` viewset instead.

PR #194, opened 2020-04-02
add client status bar

shapiromatron:
- add client status bar
- set current package version to 0.1 (still incubating)
- update our dev environment to include `hawc_client`


client/hawc_client/__init__.py
```diff
@@ -109,9 +111,12 @@ def iter_pages(self, url: str, params: Dict = None) -> Generator:
         """
         response_json = self.get(url, params)
         yield response_json["results"]
-        while response_json["next"] is not None:
-            response_json = self.get(response_json["next"])
-            yield response_json["results"]
+        num_pages = math.ceil(response_json["count"] / len(response_json["results"]))
```
rabstejnek, line 114: Is there a case where a user would want to use the generator directly w/o the status bar, and in that case would it be better to put the bar before the list comprehensions in animal.endpoints() and epi.endpoints()?

shapiromatron: if we needed to, we could add a new option on the client to disable status bars, or we could make that a param to this method. However,  I propose not adding it until there's a real use case for it, what do you think?

rabstejnek: Sounds good!

PR #197, opened 2020-04-03
Icf 96 bulk reference tagging resubmit

DookTibs:
Here's the bulk tagging feature in the new codebase, with a few tweaks and a test added.


shapiromatron:
Closing; much of this code will be reused in future body of work which allows an alternative tagging interface, however the jQuery approach will be rewritten in mobx/react

PR #202, opened 2020-04-09
Pubmed search errrors

shapiromatron:
- fix pubmed search error; str vs int in set comparison
- hotfix - fix client json() iterator 
- check for duplicates in data frame upload
- sort reference tag export by keys for stability


shapiromatron:
@rabstejnek fyi, a few changes in this diff you may want to review; could be some merge conflicts

PR #207, opened 2020-04-14
Replace custom exports

rabstejnek:


hawc/apps/common/helper.py
```diff
@@ -204,152 +192,5 @@ def _get_tags(cls, e):
         return returnValue
 
     def build_response(self):
-        header_row = self._get_header_row()
```
shapiromatron, line 207: ohh yeah! this is a good screenshot:
Screen Shot 2020-04-14 at 8 42 47 PM

PR #209, opened 2020-04-17
Rob direction visuals

rabstejnek:


rabstejnek:
Three different visuals were updated to reflect bias directions.

Heatmap visual:
image
image
Details shown on click

Donut visual:
image
Details shown on hover

Risk of bias score bar:
image
On hover
image

Example from public assessment [PFBS(2018)](https://hawcprd.epa.gov/assessment/100000037/)

frontend/riskofbias/constants.js
```diff
@@ -80,15 +80,20 @@ const NA_KEYS = [10, 20],
...
+    BIAS_DIRECTION_COMPACT = {
+        [BIAS_DIRECTION_UNKNOWN]: "-",
+        [BIAS_DIRECTION_UP]: "⯅",
+        [BIAS_DIRECTION_DOWN]: "⯆",
+    },
     BIAS_DIRECTION_SIMPLE = {
         [BIAS_DIRECTION_UNKNOWN]: "?",
-        [BIAS_DIRECTION_UP]: "⬆",
-        [BIAS_DIRECTION_DOWN]: "⬇",
+        [BIAS_DIRECTION_UP]: "🡩",
```
shapiromatron, line 90: these don't render for me; I think they're not defined in all fonts. I think we'll need to revert back to the prior arrows for compatibility

Screen Shot 2020-04-23 at 10 33 03 AM

frontend/summary/summary/RoBHeatmapPlot.js
```diff
@@ -283,6 +287,40 @@ class RoBHeatmapPlot extends D3Visualization {
...
+                d =>
+                    y(d[self.yField]) +
+                    (d.direction == BIAS_DIRECTION_UP ? quarter_width : three_quarter_width)
+            )
+            .attr("text-anchor", "middle")
+            .attr("dy", "3.5px")
+            .attr("class", function(d) {
+                var returnValue = "centeredLabel";
+
+                if (
```
shapiromatron, line 307: instead of all the conditionals, use `d.metric.domain.is_overall_confidence`

rabstejnek:
Override now correctly shows on visuals:
image
image
And as suggested, the arrow we were using was changed back to the old style to solve compatibility issues.

frontend/riskofbias/constants.js
```diff
@@ -80,15 +80,25 @@ const NA_KEYS = [10, 20],
...
     BIAS_DIRECTION_UP = 1,
     BIAS_DIRECTION_DOWN = 2,
+    BIAS_DIRECTION_CHOICES = {
+        [BIAS_DIRECTION_UNKNOWN]: "? (Unknown/not specified)",
+        [BIAS_DIRECTION_UP]: "⬆ (Away from null)",
+        [BIAS_DIRECTION_DOWN]: "⬇ (Towards null)",
+    },
+    BIAS_DIRECTION_COMPACT = {
+        [BIAS_DIRECTION_UNKNOWN]: "",
+        [BIAS_DIRECTION_UP]: "⯅",
```
shapiromatron, line 90: use these instead? ▲ ▼

shapiromatron: if we keep running into issues with this we can use svg-based triangles that will work on anyone's browser

frontend/summary/summary/RoBHeatmapPlot.js
```diff
@@ -283,6 +291,44 @@ class RoBHeatmapPlot extends D3Visualization {
             .style("fill", d => d.score_text_color)
             .text(d => d.score_text);
 
+        this.direction_up = this.cells_group
```
shapiromatron, line 294: i like this approach, nice job. if you wanted you could also filter this.cells_data beforehand to only include cells that have the up/down you want, and then you don't embed hidden nodes to the svg

frontend/riskofbias/constants.js
rabstejnek: Changed the triangles. So now all of our directional characters are from unicode version 4 instead of 7, so I imagine we'll find these in a bunch more font packs.

frontend/summary/summary/RoBHeatmapPlot.js
rabstejnek: Done!

PR #214, opened 2020-05-05
OpenAPI (Swagger API) integration

rabstejnek:
Added endpoint "/openapi" that generates an OpenAPI spec of HAWC's API endpoints.

### DIFF
--- a/hawc/apps/assessment/serializers.py
+++ b/hawc/apps/assessment/serializers.py
@@ -38,6 +38,10 @@ class EffectTagsSerializer(serializers.ModelSerializer):
         # obj is a model-manager in this case; convert to list to serialize
         return list(obj.values("slug", "name"))
 
+    class Meta:
+        model = models.EffectTag
+        fields = "__all__"
+
 
 class DoseUnitsSerializer(serializers.ModelSerializer):
     class Meta:
--- a/hawc/apps/common/serializers.py
+++ b/hawc/apps/common/serializers.py
@@ -3,7 +3,8 @@ from rest_framework import serializers
 
 class UnusedSerializer(serializers.Serializer):
     """
-    Nondescript serializer. Used in instances where a view does not use a serializer but one needs to be set, such as for OpenAPI generation.
+    Nondescript serializer. Used in instances where a view does not use a serializer
+    but one needs to be set, such as for OpenAPI generation.
     """
 
     pass
--- a/hawc/apps/common/serializers.py
+++ b/hawc/apps/common/serializers.py
@@ -1,0 +1,9 @@
+from rest_framework import serializers
+
+
+class UnusedSerializer(serializers.Serializer):
+    """
+    Nondescript serializer. Used in instances where a view does not use a serializer but one needs to be set, such as for OpenAPI generation.
+    """
+
+    pass
--- a/hawc/apps/epi/api.py
+++ b/hawc/apps/epi/api.py
@@ -1,4 +1,3 @@
-from rest_framework import serializers as ser
 from rest_framework import viewsets
 from rest_framework.decorators import action
 from rest_framework.response import Response
@@ -7,6 +6,7 @@ from ..assessment.api import AssessmentLevelPermissions, AssessmentViewset
 from ..assessment.models import Assessment
 from ..common.api import CleanupFieldsBaseViewSet, LegacyAssessmentAdapterMixin
 from ..common.renderers import PandasRenderers
+from ..common.serializers import UnusedSerializer
 from ..common.views import AssessmentPermissionsMixin
 from . import exports, models, serializers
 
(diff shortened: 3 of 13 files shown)


hawc/apps/assessment/serializers.py
```diff
@@ -38,6 +38,10 @@ def to_representation(self, obj):
         # obj is a model-manager in this case; convert to list to serialize
         return list(obj.values("slug", "name"))
 
+    class Meta:
```
rabstejnek, line 41: Meta class was required for OpenAPI, and since the serializing/deserializing methods are overwritten I don't believe this addition would break anything. I ran an API endpoint that used this serializer and I saw no difference.

hawc/apps/epi/api.py
```diff
@@ -16,6 +17,7 @@ class EpiAssessmentViewset(
     parent_model = Assessment
     model = models.Outcome
     permission_classes = (AssessmentLevelPermissions,)
+    serializer_class = ser.Serializer
```
rabstejnek, line 20: serializer_class was required for OpenAPI, so I used the base Serializer class for viewsets that didn't have their serializer_class set. I considered using serializers that best matches the models being used, but I was concerned that would make the code harder to read since people may assume the serializer was actually being used. The tradeoff is a nondescript "response content" portion of the OpenAPI spec.

hawc/main/urls.py
```diff
@@ -3,8 +3,35 @@
...
+from hawc.apps.epi.urls import router as epi_api
+from hawc.apps.epimeta.urls import router as epimeta_api
+from hawc.apps.invitro.urls import router as invitro_api
+from hawc.apps.lit.urls import router as lit_api
+from hawc.apps.mgmt.urls import router as mgmt_api
+from hawc.apps.riskofbias.urls import router as riskofbias_api
+from hawc.apps.study.urls import router as study_api
+from hawc.apps.summary.urls import router as summary_api
+
+open_api_patterns = [
```
rabstejnek, line 22: All of the modular url dispatchers used a router object for their API endpoints, so I just imported them all to the main url dispatcher for the OpenAPI to traverse.

hawc/apps/epi/api.py
shapiromatron: This looks great. Can we subclass the serializer class and put in the common package and call it something like `UnusedSerializer` with a docstring that describes what you wrote above? Then it'd be easy to find all the ones to fix.

hawc/apps/epimeta/api.py
```diff
@@ -1,3 +1,4 @@
+from rest_framework import serializers as ser
 from rest_framework import viewsets
```
shapiromatron, line 2: change line 2 to `from rest_framework import serializers, viewsets` instead of another import, both here and in other files

hawc/main/urls.py
shapiromatron: The include in the urls above can use a string, eg., "hawc.apps.animal.urls"; can these includes use the str version so we don't have to have the imports

hawc/main/urls.py
```diff
@@ -43,6 +70,16 @@
     ),
     url(rf"^admin/{settings.ADMIN_URL_PREFIX}/", admin.site.urls),
     url(r"^selectable/", include("selectable.urls")),
+    url(
+        r"^openapi/$",
+        get_schema_view(
+            title="HAWC OpenAPI",
+            version="0.0.1",
```
shapiromatron, line 77: can we add a `__version__` to `hawc/__init__.py` and use that value here. I think it should be 0.1 since we've used that elsewhere, please confirm

shapiromatron:
Also, could you please add a unit-test that the the schema generates as a admin user, and fails if authenticated as a non-admin?

hawc/main/urls.py
rabstejnek: I tried it with a string parameter, [but it turns out that can only be done with modules.](https://docs.djangoproject.com/en/3.0/ref/urls/#include)

hawc/apps/epi/api.py
rabstejnek: Good idea, done!

hawc/main/urls.py
rabstejnek: I tried it with a string parameter, [but it turns out that can only be done with modules.](https://docs.djangoproject.com/en/3.0/ref/urls/#include)

I did clean up the import a bit, and just used full paths within the code instead of creating a new namespace for all the imports. I think the tradeoff looks a little cleaner this way, let me know what you think.

hawc/main/urls.py
rabstejnek: Yes, it is 0.1 in the setup.cfg. I made the requested changes.

hawc/apps/common/serializers.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import serializers
+
+
+class UnusedSerializer(serializers.Serializer):
+    """
+    Nondescript serializer. Used in instances where a view does not use a serializer but one needs to be set, such as for OpenAPI generation.
```
shapiromatron, line 6: one more comment, then good to merge - can you please wrap the text here, instead of relying on your editor to wrap?

PR #215, opened 2020-05-08
fix assorted edge-case api errors

shapiromatron:
- fix cases where bad literature ids are provided
- fix cases where references have no tags applied
- fix epi meta export to build null columns if needed
- log 500 errors to a file in addition to emails
- massive fixture update w/ real data
- added api tests for summary.DataPivot and summary.Visual, and epi, epimeta, animal, and invitro data exports
- added docs and mechanism to update all exports automatically


tests/apps/test_testenv.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+import pytest
+
+in_ci = os.environ.get("GITHUB_RUN_ID") is not None
+
+
+@pytest.mark.skipif(not in_ci, reason="only run in CI")
```
rabstejnek, line 8: Does having it a directory higher than the other tests ensure it gets run first?

shapiromatron: not sure, but i don't think order matters really, it was more that this isn't really app specific

tests/apps/lit/test_lit_managers.py
```diff
@@ -0,0 +1,18 @@
...
+        # assert empty tag list works
+        df = models.ReferenceTags.objects.as_dataframe(db_keys.assessment_working)
+        assert df.shape == (0, 2)
+        assert df.to_csv(index=False) == "reference_id,tag_id\n"
+
+        # add a tag and make sure it appears
+        models.Reference.objects.get(id=1).tags.set([2])
+        df = models.ReferenceTags.objects.as_dataframe(db_keys.assessment_working)
+        assert df.shape == (1, 2)
+        assert df.to_csv(index=False) == "reference_id,tag_id\n1,2\n"
```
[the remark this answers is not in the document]
shapiromatron: ahh, good catch

PR #216, opened 2020-05-08
Study create API endpoint

rabstejnek:
- [x] API endpoint
- [x] Python client function
- [x] Testing


rabstejnek:
Waiting on updated fixture to add unit tests for endpoint / client.

rabstejnek:
Exception handling and testing done, ready for review!

shapiromatron:
@rabstejnek you're working on this a little more now right? if so I'll hold off on review

rabstejnek:
@shapiromatron yep, I have to make some further changes

hawc/apps/study/api.py
```diff
@@ -47,6 +55,27 @@ def types(self, request):
...
+
+        data = request.data.dict() if isinstance(request.data, QueryDict) else request.data
+        reference_id = tryParseInt(data.pop("reference_id", -1), -1)
+
+        try:
+            reference = Reference.objects.get(id=reference_id)
+        except ObjectDoesNotExist:
+            raise ValidationError(f"Reference ID does not exist.")
+        data["assessment"] = reference.assessment.id
+
```
rabstejnek, line 68: I handled the above this way so that I could check permissions of the reference object. I suppose I could pass the request to the serializer as context and test there, but I'm not sure if thats appropriate for the serailizer. Let me know!

rabstejnek:
@shapiromatron Its ready for review. The best way I found to move the reference validation over to the serializer was through a context argument, since I was struggling to get the reference_ptr field to work.

client/hawc_client/__init__.py
```diff
@@ -511,6 +511,51 @@ def visual_list(self, assessment_id: int) -> pd.DataFrame:
         return pd.DataFrame(response_json)
 
 
+class StudyClient(BaseClient):
+    """
+    Client class for study requests.
+    """
+
+    def create(
+        self, reference_id: int, short_citation: str, full_citation: str, data: Dict = {}
```
shapiromatron, line 520: this is a common python issue. this creates a single dict that is reused anywhere a default  is not provided; so if it's modified it persists even after the method call ends.  To fix:

```python
def create(self, data: Optional[Dict] = None):
    if data is None:
         data = {}
```

tests/apps/study/test_study_api.py
```diff
@@ -56,3 +57,57 @@ def test_study_detail_api(db_keys):
         "url": "/study/1/",
         "rob_response_values": [27, 26, 25, 22, 24, 20],
     }
+
+
+@pytest.mark.django_db
+def test_study_create_api(db_keys):
```
shapiromatron, line 63: for long tests like this, I'm switching to use this pattern:

```
class TestStudyCreateApi:

    def test_permissions():
        ...
   def test_bad_requests():
        ...
   def test_valid_requests():
        ...
```

Breaks things up a little bit

hawc/apps/study/api.py
```diff
@@ -47,6 +55,27 @@ def types(self, request):
...
+        self.check_object_permissions(request, reference)
+
+        serializer_class = self.get_serializer_class()
+        serializer = serializer_class(data=data, context={"reference_id": reference_id})
+        serializer.is_valid(raise_exception=True)
+
+        created_study = models.Study.save_new_from_reference(reference, serializer.validated_data)
```
shapiromatron, line 75: this should go in the `create` method https://www.django-rest-framework.org/api-guide/serializers/#saving-instances

hawc/apps/study/api.py
shapiromatron: I can see why you put it here, but it basically means you have to entirely rewrite the create method.   It'd probably be better to put everything in the serializer, and then if you want you can add this:

```python
def create(self, request):
    # permissions check not here; see serializer validation
    return super().create(request)
```

hawc/apps/study/serializers.py
```diff
@@ -20,6 +22,25 @@ class Meta:
         fields = "__all__"
 
 
+class CreateStudySerializer(StudySerializer):
```
shapiromatron, line 25: this is the same as `SimpleStudySerializer` with a validate method; why not use that instead?

client/hawc_client/__init__.py
rabstejnek: Good to know, thanks! Fixed.

tests/apps/study/test_study_api.py
rabstejnek: Good idea, fixed!

hawc/apps/study/api.py
rabstejnek: Done!

hawc/apps/study/api.py
```diff
@@ -14,7 +14,9 @@
 from . import models, serializers
 
 
-class Study(viewsets.ReadOnlyModelViewSet):
+class Study(
+    viewsets.ReadOnlyModelViewSet, mixins.CreateModelMixin,
```
shapiromatron, line 18: by convention, mixins come before base models, as they should replace default implementations; order should be switched: https://www.ianlewis.org/en/mixins-and-python

hawc/apps/study/api.py
```diff
@@ -53,6 +57,10 @@ def types(self, request):
         study_types = self.model.STUDY_TYPE_FIELDS
         return Response(study_types)
 
+    def create(self, request):
+        # permissions check not here; see serializer validation
```
shapiromatron, line 61: i like this comment; what do you think?

hawc/apps/study/serializers.py
```diff
@@ -29,6 +31,38 @@ class Meta:
...
+            raise exceptions.PermissionDenied("Invalid permission to edit assessment.")
+        return super().validate(data)
+
+    def create(self, validated_data):
+        reference = Reference.objects.get(id=self.initial_data.get("reference_id"))
+        validated_data["assessment"] = reference.assessment.id
+
+        return models.Study.save_new_from_reference(reference, validated_data)
+
+    class Meta:
```
shapiromatron, line 57: do we need meta, or can it just be inherited from simple?  In fact, can we just add these methods to SmipleStudySerializer and remove this subclass?

tests/apps/study/test_study_api.py
```diff
@@ -56,3 +57,77 @@ def test_study_detail_api(db_keys):
...
+        assert response.status_code == 403
+
+        # public shouldn't be able to create
+        client = APIClient()
+        response = client.post(url, data)
+        assert response.status_code == 403
+
+    def test_bad_requests(self, db_keys):
+        # payload needs to include the required short_citation and full_citation
+        url = "/study/api/study/"
```
shapiromatron, line 85: instead of hard coding, use `reverse("study:api:study-list")` here and elsewhere

tests/apps/study/test_study_api.py
```diff
@@ -56,3 +57,77 @@ def test_study_detail_api(db_keys):
...
+        # invalid references will not be successful
+        data = {
+            "reference_id": "invalid",
+            "short_citation": "Short citation.",
+            "full_citation": "Full citation.",
+        }
+
+        response = client.post(url, data)
+        assert response.status_code == 400
+        assert "Reference ID must be a number." == str(response.data["non_field_errors"][0])
```
shapiromatron, line 102: i generally put expected str after the new data; that way i can read it "i'm going to get this thing and compare it to this expected result"; it reads easier for me. Do you  prefer this way?

hawc/apps/study/api.py
rabstejnek: Fixed!

hawc/apps/study/api.py
rabstejnek: Yeah I think its a good addition, it'll be a helpful reminder if we ever have to troubleshoot it.

hawc/apps/study/serializers.py
rabstejnek: The meta is slightly different; the CreateStudySerializer also excludes assessment, since we figure that out from the given reference id. Also the views that use the SimpleStudySerializer do not pass in a reference id, so the validation would be different. I'm sure I can combine them and just test on what keys are in the initial data, but it would lead to one more complicated serializer instead of two less complicated serializers. Would you rather that?

tests/apps/study/test_study_api.py
rabstejnek: I prefer it your way, I just haven't kept convention. I'll fix this and keep it in mind going forward!

hawc/apps/study/serializers.py
rabstejnek: I combinded the serializers, let me know if you like the looks of it!

shapiromatron: LGTM

shapiromatron:
looks great!

PR #217, opened 2020-05-13
Hotfix xlsx renderer

rabstejnek:
I've removed the binary restrictions from the XLSX renderer, which will allow any unicode error responses to be encoded with the base charset.

The django-rest-pandas library [does not place such restrictions](https://github.com/wq/django-rest-pandas/blob/53109d9bd97fac9b82ec44ad54e7c3d8911847c6/rest_pandas/renderers.py#L197), so I thought this was an appropriate fix.

Unit tests were also added for RoB exports using the xlsx format, and they pass successfully!


tests/apps/riskofbias/test_riskofbias_api.py
```diff
@@ -40,6 +40,21 @@ def test_riskofbias_detail(db_keys):
     ]
 
 
+@pytest.mark.django_db
```
shapiromatron, line 43: we need a test on the renderer, not a specific API endpoint: Use this instead:

```python
from rest_framework import status

def test_base_renderer_not_success():
    data = {"error": "Method not allowed"}
    context = {"response": Response(status=status.HTTP_405_METHOD_NOT_ALLOWED)}
    result = renderers.PandasXlsxRenderer().render(data, renderer_context=context)
    assert result == json.dumps(data)

```

shapiromatron: we should also throw an error in the base renderer if status is not ok and the response isn't json, since we don't know how to return

rabstejnek: I originally wrote something very similar to that, but it was not reproducing the server error. It looks like when the view calls ```response.render(); response.rendered_content``` the error is triggered. The view also sets a bunch of properties on that response before the call to render, which I tried to write myself but the code was losing some readability.

rabstejnek: I wrote a test for the xlsx renderer that replicates the server error.

shapiromatron:
@rabstejnek more changes coming or is this one done?

rabstejnek:
@shapiromatron It's ready for rereview. I added a test to the renderer at your request and was able to reproduce the server error, so the case is covered.

rabstejnek:
@shapiromatron I added another test for the case of an OPTIONS request, if you'd like to give it a once over before merge.

PR #219, opened 2020-05-14
Hotfix: catch casting errors on PKs and return appropriate 404

rabstejnek:


rabstejnek:
Making a different approach in another PR.

PR #223, opened 2020-05-19
RoB heatmap footnote updates and visualization improvements

shapiromatron:
Smörgåsbord of assorted risk of bias visualization settings:

- for HAWC PRIME (not EPA); fix N/A to be grey instead of yellow (same as riskofbias/constants.js SCORE_SHADES)
- Used better contrasting colors for all text for visibility
- For multiscore gradients, made spacing larger for higher visibility
- make arrow size on heatmap for more visibility
- use larger asterisk on heatmap for visibility
- add footnotes to heatmap legend when required

Some screenshots

---

New legend items (only appear when displayed on visual): 
Screen Shot 2020-05-19 at 10 54 30 AM

---

Updated contrast and size of items on heatmap:
Screen Shot 2020-05-19 at 12 10 27 PM

---

Updated contrasting color display on tables:
Screen Shot 2020-05-19 at 11 33 51 AM


frontend/summary/summary/RoBHeatmapPlot.js
```diff
@@ -452,13 +451,36 @@ class RoBHeatmapPlot extends D3Visualization {
...
+        const rob_response_values = this.data.aggregation.studies[0].data.rob_response_values,
             options = {
                 dev: this.options.dev || false,
                 collapseNR: false,
-            };
-        this.legend = new RoBLegend(this.svg, this.data.settings, rob_response_values, options);
+            },
+            getFootnoteOptions = () => {
+                let footnotes = [];
+                if (this.cells_data.filter(d => d.robArray.length > 1).length > 0) {
```
rabstejnek, line 464: The asterisk is only applied if the number of unique scores is > 1, so there can be a mismatch of legend to visual if there are multiple scores but they're the same score. I have included a screenshot of an example where it was rescored and the score stayed the same but bias changed.
image

So I suppose I would suggest chaining a uniq() to the end of the filter to be in line with how the asterisk is actually decided. This is a very specific use case and probably won't happen in actual practice, though.
For reference, [where the asterisk is added](https://github.com/shapiromatron/hawc/blob/e996a17a2a3debe20dd3eb42caef0b6d6cd1a50b/frontend/riskofbias/constants.js#L121)

shapiromatron: Fixed.

shapiromatron:
good catch @rabstejnek; I'd argue that this PR has some changes that need to be made here since this is inconsistent logic in the tool; feel free to not always "approve", you found an important logic edge case here that I needed to fix.

PR #246, opened 2020-06-23
Generic view improvements

shapiromatron:


shapiromatron:
Closed; see #249 which captures this work and more

PR #247, opened 2020-06-23
Heatmap TODO stubs

rabstejnek:
Implement all TODO stubs and remove them


tests/apps/epi/test_epi_api.py
```diff
@@ -10,6 +10,25 @@
 
 @pytest.mark.django_db
 class TestEpiAssessmentViewset:
+    def _test_heatmap(self, rewrite_data_files: bool, slug: str, key: int, unpublished: bool):
```
shapiromatron, line 13: this code is almost identical to `test_full_export` - can all 3 tests be consolidated like  `_test_animal_export` in `test_animal_api.py.`

tests/apps/lit/test_lit_api.py
```diff
@@ -99,6 +99,25 @@ def test_reference_year_histogram(self, db_keys):
         resp = c.get(url).json()
         assert resp["data"][0]["type"] == "histogram"
 
+    def test_tag_heatmap(self, rewrite_data_files: str, db_keys):
```
shapiromatron, line 102: like above, can a helper function be written to test `test_references_download` and this function?

tests/apps/animal/test_animal_api.py
```diff
@@ -25,6 +25,25 @@ def _test_animal_export(self, rewrite_data_files: bool, slug: str, key: int):
 
         assert data == json.loads(fn.read_text())
 
+    def _test_heatmap(self, rewrite_data_files: bool, slug: str, key: int, unpublished: bool):
```
shapiromatron, line 28: can you refactor `test_endpoint_export` so we dont have two helper functions that basically do the same thing?

rabstejnek:
# Modal button overhaul
image
image

A regular button shows up for single result, which on click opens up the modal. A dropdown button shows up for multiple results, which on click drops a list of the unique values that the modal targets. A disabled button shows up if there are too many results (more than 10) or no results; there is a mouseover that specifies which it is, and the "too many results" shows an open eye while the "no results" shows a close eye icon.

PR #256, opened 2020-07-06
Data pivot fix

rabstejnek:


hawc/apps/animal/exports.py
```diff
@@ -498,38 +512,56 @@ def _get_data_rows(self):
...
-                        self._get_dose(doses, 1),  # first non-zero dose
-                        self._get_dose(doses, ser["NOEL"]),
-                        self._get_dose(doses, ser["LOEL"]),
-                        self._get_dose(doses, ser["FEL"]),
-                        self._get_dose(doses, len(ser["groups"]) - 1),
-                    ]
-                )
-            else:
-                row.extend([None] * 5)
+            try:
```
shapiromatron, line 529: move this to a new static method and add a few unit tests:

```python
    @staticmethod
    def dose_index_low_high(doses: List[float]) -> Tuple:
        try:
            dose_validity_list = list(map(lambda d: d is not None, doses))
            low_index = dose_validity_list[1:].index(True) + 1
            high_index = len(doses) - 1 - dose_validity_list[1:][::-1].index(True)
            return (doses[low_index], doses[high_index])
        except ValueError:
            return (None, None)
```

tests/apps/animal/test_animal_exports.py
```diff
@@ -53,3 +53,40 @@ def test_get_significance_and_direction():
         ],
     )
     assert resp == ["No", "No", "Yes - ?", "Yes - ↓", "Yes - ↑"]
+
+
+class TestEndpointFlatDataPivot:
+    def test_dose_index_low_high(self):
```
shapiromatron, line 59: nice!!!

hawc/apps/animal/exports.py
```diff
@@ -167,6 +168,8 @@ def _get_dose_units(cls, doses):
 
     @classmethod
     def _get_doses_str(cls, doses):
```
shapiromatron, line 170: would you mind adding annotations here pls? since you just touched it and it's a tiny little function, it'd be easy to add. Thanks

hawc/apps/animal/exports.py
```diff
@@ -454,6 +457,26 @@ def _get_bmd_values(bmd, preferred_units):
...
 
+    @staticmethod
+    def _dose_has_n(dose_group_id: int, groups: List[Dict]) -> bool:
+        for group in groups:
+            if group["dose_group_id"] == dose_group_id:
+                return group["n"] is not None
+        return False
+
+    @staticmethod
+    def _dose_index_low_high(dose_list: List[float]) -> Tuple:
```
shapiromatron, line 468: I guess this is `List[Optional[float]]`, and this returns `Tuple[Optional[float], Optional[float]]` assuming these are valid.  Could you add a docstring to explain what this does?

PR #258, opened 2020-07-10
Literature reference API

rabstejnek:
- [x] DELETE /api/lit/reference/:id/
  - [x] delete a hawc reference
  - [x] also add a django view to allow delete
- [x] PATCH /api/lit/reference/:id/
  - [x] with a payload {title: "this is my title", tags: [1,2,4]}
  - [x] key use case: - replace all tags for a given reference
- [x] Add client functions for new endpoints


rabstejnek:
### There is now a delete option
image
---
### It brings the user to a page to confirm delete.
image
---
### The reference is successfuly deleted and removed from applicable imports. Success message is given and user is directed to lit overview page.
image
---
#### Notes
- I was not able to extend the delete page with the detail template content, I'm assuming because it is generated by JS. Is it necessary to include this info or is it fine as is?
- The explanation of consequences of deleting the reference is pretty basic. Does this need to be expanded, or should the user already know the consequences?

client/hawc_client/__init__.py
```diff
@@ -255,6 +285,48 @@ def references(self, assessment_id: int) -> pd.DataFrame:
...
+        Returns:
+            Dict: JSON representation of the updated reference.
+        """
+        url = f"{self.session.root_url}/lit/api/reference/{reference_id}/"
+        response_json = self.session.patch(url, kwargs).json()
+        return response_json
+
+    def delete_reference(self, reference_id: int) -> None:
+        """
+        Deletes the given reference. This also removes the reference from
```
shapiromatron, line 317: suggested revision; `Deletes the selected reference. This also removes the reference from any searches/imports which may have included the reference. If data was extracted with this reference and it is associated with bioassay, or epi extractions they will also be removed.`

hawc/apps/lit/serializers.py
```diff
@@ -196,3 +196,21 @@ def bulk_create_tags(self):
...
 
             models.Reference.delete_cache(assessment_id)
+
+
+class ReferenceSerializer(serializers.ModelSerializer):
+    tags = serializers.ListField(write_only=True)
+
+    def update(self, instance, validated_data):
+
+        if "tags" in validated_data:
```
shapiromatron, line 206: tags are popped here and then never used; does this mean they're not saved?

hawc/apps/lit/templates/lit/reference_confirm_delete.html
```diff
@@ -0,0 +1,15 @@
...
+{% endblock title %}
+
+{% block breadcrumbs_self %}
+  <li><a href="{{object.get_absolute_url}}">{{object}}</a><span class="divider">/</span></li>
+  <li class="active"> Delete <span class="divider">/</span></li>
+{% endblock breadcrumbs_self %}
+
+{% block content %}
+  {{ block.super }}
+  {% include "hawc/_delete_block.html" with name="reference" notes="This will remove all instances of this reference from this assessment." %}
```
shapiromatron, line 14: Add same as above: `This also removes the reference from any searches/imports which may have included the reference. If data was extracted with this reference and it is associated with bioassay, or epi extractions they will also be removed.`

hawc/apps/lit/views.py
```diff
@@ -440,6 +440,14 @@ class RefEdit(BaseUpdate):
     form_class = forms.ReferenceForm
 
 
+class RefDelete(BaseDelete):
+    success_message = "Reference deleted."
+    model = models.Reference
+
+    def get_success_url(self):
+        return reverse_lazy("lit:overview", kwargs={"pk": self.assessment.pk})
```
shapiromatron, line 448: More recently, I'm using args instead of kwargs b/c it's shorter: `return reverse_lazy("lit:overview", args=(self.assessment.pk,))`

notebooks/client-update-remove-reference.ipynb
```diff
@@ -0,0 +1,278 @@
...
+    {
+     "name": "stdin",
+     "output_type": "stream",
+     "text": [
+      " ········\n"
+     ]
+    }
+   ],
+   "source": [
+    "client = HawcClient('http://127.0.0.1:8000')\n",
```
shapiromatron, line 55: change to `https://hawcproject.org`

client/hawc_client/__init__.py
```diff
@@ -255,6 +285,48 @@ def references(self, assessment_id: int) -> pd.DataFrame:
...
+                title (str): title of the reference
+                abstract (str): reference abstract
+                tags (List[int]): tag IDs to apply to reference;
+                    completely replaces current tags
+
+        Example Usage:
+            updated_reference_json = client.lit.update_reference(
+                reference_id = 1,
+                title = "reference",
+                tags = [1,2,3]
```
shapiromatron, line 305: can you add if this appends or replaces the tags? I think I'd prefer replace, but we should be explicit here

tests/apps/lit/test_lit_api.py
```diff
@@ -5,6 +5,8 @@
 from django.urls import reverse
 from rest_framework.test import APIClient
```
shapiromatron, line 6: I think we also need to add tests to the hawc_client too for these endpoints; I forgot to do so on the bioassay create API, will need to add...

client/hawc_client/__init__.py
shapiromatron: and if it's append by default, then we don't currently have a way to delete the tags I think, is that right?

hawc/apps/lit/serializers.py
rabstejnek: the `instance.tags.set` updates and saves the reference with the given tags, so at that point they're popped since the operation is done / that key is unnecessary in the next steps. I'll add some comments clarifying it

client/hawc_client/__init__.py
rabstejnek: it replaces; I will add a note about that

hawc/apps/lit/serializers.py
shapiromatron: should we decorate w/ `@transaction.atomc()` or something similar since it sounds like there are multiple db operations here?

hawc/apps/lit/serializers.py
```diff
@@ -196,3 +196,24 @@ def bulk_create_tags(self):
...
+
+class ReferenceSerializer(serializers.ModelSerializer):
+    tags = serializers.ListField(write_only=True)
+
+    @transaction.atomic()
+    def update(self, instance, validated_data):
+
+        # updates the reference tags
+        if "tags" in validated_data:
+            instance.tags.set(validated_data.pop("tags"))
```
shapiromatron, line 209: Just realized it could be the case the ids are not from this assessment.  This needs to be a little more involved. It probably belongs in the `validate` method:

```python
tag_ids = validated_data.pop("tags")
tags = ReferenceFilterTag.get_assessment_qs(self.instance.assessment_id).filter(id__in=tag_ids)
if tags.count() != len(tag_ids):
    raise serializer.ValidationError('...')
validated_data[tags] = tags

```

tests/client/test_client.py
```diff
@@ -156,6 +155,23 @@ def test_lit_references(self):
...
+        client.authenticate("pm@pm.com", "pw")
+        updated_title = "client test"
+        response = client.lit.update_reference(self.db_keys.reference_linked, title=updated_title)
+        assert isinstance(response, dict)
+        assert response["title"] == updated_title
+
+    def test_lit_delete_reference(self):
+        client = HawcClient(self.live_server_url)
+        client.authenticate("pm@pm.com", "pw")
+        try:
```
shapiromatron, line 169: just saw this test. Might be better to retrieve and make sure response is 200, then delete and assert that the response is None or empty or something?  And then maybe use the retrieve again and assert it returns a 404

PR #260, opened 2020-07-15
Literature reference API #2

shapiromatron:
Clone of  https://github.com/shapiromatron/hawc/pull/258/commits; this PR didn't update w/ the latest commit on branch


tests/client/test_client.py
```diff
@@ -156,6 +155,29 @@ def test_lit_references(self):
         response = client.lit.references(self.db_keys.assessment_client)
         assert isinstance(response, pd.DataFrame)
 
+    def test_lit_reference(self):
```
shapiromatron, line 158: @rabstejnek minor rewrite of the test (mostly to fix the api), but I think it reads well too, what do you think?

rabstejnek: I like the changes!

PR #262, opened 2020-07-15
HERO reference APIs

rabstejnek:
- [x] POST /api/lit/assessment/:id/update-reference-metadata-from-hero/
  - [x] will update metadata content for all hero references in a project
- [x] POST /api/lit/assessment/:id/replace-hero/
  - [x] with a payload like: {replace: [[1,2],[3,4]]}
  - [x] always update hero metadata w/ a replace
  - [x] throw an exception if in the example above, a reference w/ hero id 2 or 4 already exists in hawc


hawc/apps/lit/api.py
```diff
@@ -166,6 +169,64 @@ def tag_heatmap(self, request, pk):
         export = FlatExport(df=df, filename=f"df-{instance.id}")
         return Response(export)
 
+    @transaction.atomic()
+    @action(
+        detail=True, methods=("patch",), url_path="replace-hero",
+    )
+    def replace_hero(self, request, pk):
+        assessment = self.get_object()
+        body = json.loads(request.body)
```
shapiromatron, line 178: convert all this logic into a serializer

hawc/apps/lit/api.py
```diff
@@ -166,6 +169,64 @@ def tag_heatmap(self, request, pk):
...
+        replace = body.get("replace", list())
+        replace_unzipped = list(zip(*replace))
```
shapiromatron, line 180: Use to tuple unpacking to get the ref and hero lists:

```python
x, y = zip(*[[1,2],[3,4]])
```

hawc/apps/lit/api.py
```diff
@@ -166,6 +169,64 @@ def tag_heatmap(self, request, pk):
...
+
+        # make sure all references are HERO and in assessment
+        matching_references = assessment.references.filter(
+            id__in=ref_ids, identifiers__database=constants.HERO
+        )
+        if matching_references.count() != len(ref_ids):
+            raise exceptions.ValidationError("All references must be from selected assessment.")
+
+        # set hero ref
+        for index, (ref, hero) in enumerate(replace):
```
shapiromatron, line 192: move all the code here into lit/models or lit/managers, and lit tasks.  the API should call biz logic, not have lots of it nested in it

hawc/apps/lit/api.py
```diff
@@ -166,6 +169,64 @@ def tag_heatmap(self, request, pk):
...
+            reference = models.Reference.objects.get(id=ref)
+            hero_identifier = reference.identifiers.get(database=constants.HERO)
+            setattr(hero_identifier, "unique_id", str(hero))
+            hero_identifier.save()
+
+        # update content
+        tasks.update_hero_content.apply(args=[hero_ids])
+
+        # update fields with content
+        for ref in ref_ids:
```
shapiromatron, line 202: so this will work locally, but will fail in prod. The reason is `update_hero_content` is async and will not finish before you're updating the  content.  Locally, we run celery tasks in the same thread, so that's why it looks ok here.

To do this correctly, you'll need convert this to another celery task, and then chain them together: https://docs.celeryproject.org/en/stable/userguide/canvas.html#chains

hawc/apps/lit/models.py
```diff
@@ -815,5 +815,29 @@ def get_hero_id(ref):
                 return int(ident.unique_id)
         return None
 
+    def update_from_hero_content(self):
```
shapiromatron, line 818: duplicates `https://github.com/shapiromatron/hawc/blob/master/hawc/apps/lit/models.py#L590-L604`; pull this code out and into something that works in both places. Something like:

```python
def update_from_hero_content(self, hero_content:Dict):
   ...
```
This would take the json.loaded content from the identifier as an input, and it wouldn't save by default, you could have a save flag if you'd like that's commonly done in djangoland.

hawc/apps/lit/api.py
shapiromatron: I've done it once before here: https://github.com/NIEHS/orio-web/blob/master/project/analysis/tasks.py#L48

hawc/apps/lit/api.py
```diff
@@ -166,6 +169,64 @@ def tag_heatmap(self, request, pk):
         export = FlatExport(df=df, filename=f"df-{instance.id}")
         return Response(export)
 
+    @transaction.atomic()
+    @action(
+        detail=True, methods=("patch",), url_path="replace-hero",
+    )
+    def replace_hero(self, request, pk):
```
shapiromatron, line 176: there need to be some checks here to ensure that the new `hero_ids` area all valid; see `validate_valid_hero_ids`

hawc/apps/lit/api.py
```diff
@@ -166,6 +169,64 @@ def tag_heatmap(self, request, pk):
...
+        references = assessment.references.filter(identifiers__database=constants.HERO)
+        reference_ids = set(references.values_list("id", flat=True))
+        identifiers = models.Identifiers.objects.filter(
+            references__in=reference_ids, database=constants.HERO
+        )
+        hero_ids = identifiers.values_list("unique_id", flat=True)
+        # update content of hero identifiers
+        tasks.update_hero_content.apply(args=[hero_ids])
+        # update fields from content
+        for reference in references:
```
shapiromatron, line 225: same as above, this needs to be chained in tasks, I think you could probably re-use the same new task that you'll be creating above.

hawc/apps/lit/api.py
```diff
@@ -166,6 +169,64 @@ def tag_heatmap(self, request, pk):
...
+            reference = models.Reference.objects.get(id=ref)
+            reference.update_from_hero_content()
+
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @transaction.atomic()
+    @action(
+        detail=True, methods=("patch",), url_path="update-reference-metadata-from-hero",
+    )
+    def update_reference_metadata_from_hero(self, request, pk):
```
shapiromatron, line 212: need to have some sort of strategy to rate-limit these two tasks, so that if someone blasts this API it doesn't schedule a bunch of expensive tasks in the queue

shapiromatron:
@rabstejnek ready for your review; sorry it took me so long to get back to you.

client/hawc_client/__init__.py
```diff
@@ -348,6 +348,40 @@ def delete_reference(self, reference_id: int) -> None:
         url = f"{self.session.root_url}/lit/api/reference/{reference_id}/"
         self.session.delete(url)
 
+    def replace_hero(self, assessment_id: int, replace: List[List[int]]) -> None:
+        """
+        Replace HERO ID associated with each reference with a new HERO ID. Reference
+        fields are updated using the new HERO ID's reference metadata.  This request is
+        throttled; can only be executed once per hour.
```
rabstejnek, line 355: This needs to be updated if we're keeping throttle at 1/min.

client/hawc_client/__init__.py
```diff
@@ -348,6 +348,40 @@ def delete_reference(self, reference_id: int) -> None:
...
+            If the operation is unsuccessful, an error will be raised.
+        """
+        body = {"replace": replace}
+        url = f"{self.session.root_url}/lit/api/assessment/{assessment_id}/replace-hero/"
+        self.session.post(url, body)
+
+    def update_references_from_hero(self, assessment_id: int) -> None:
+        """
+        Updates the fields of all HERO references in an assessment with the most recent metadata
+        from HERO. This request is throttled; can only be executed once per hour.
```
rabstejnek, line 373: This also needs to be updated if we're keeping throttle at 1/min.

hawc/apps/lit/api.py
```diff
@@ -166,6 +171,43 @@ def tag_heatmap(self, request, pk):
...
+        detail=True,
+        throttle_classes=(OncePerMinuteThrottle,),
+        methods=("post",),
+        url_path="replace-hero",
+    )
+    def replace_hero(self, request, pk):
+        """Replace old HERO ID with new HERO ID for selected references
+
+        Expects an input of `{replace: [[1,2],[3,4]]}`, where the first element is the reference ID
+        and the select element in each list is the new HERO ID.
```
rabstejnek, line 185: Typo: should be "second", not "select"

hawc/apps/lit/serializers.py
```diff
@@ -225,3 +230,81 @@ def update(self, instance, validated_data):
...
+            int(id_)
+            for id_ in models.Identifiers.objects.filter(
+                references__in=references_diff, database=constants.HERO
+            ).values_list("unique_id", flat=True)
+        ]
+        hero_counts = Counter(itertools.chain(existing_hero_ids, self.hero_ids))
+        duplicates = [key for key, count in hero_counts.items() if count > 1]
+        if len(duplicates) > 0:
+            raise serializers.ValidationError(
+                f"Duplicate HERO IDs in assessment: {list(duplicates)}"
```
rabstejnek, line 282: I like this approach!

hawc/apps/lit/serializers.py
```diff
@@ -225,3 +230,81 @@ def update(self, instance, validated_data):
...
+        try:
+            _, _, self.fetched_content = models.Identifiers.objects.validate_valid_hero_ids(
+                self.hero_ids
+            )
+        except django.core.exceptions.ValidationError as err:
+            raise serializers.ValidationError(err.args[0])
+
+        return replace
+
+    def execute(self) -> ResultBase:
```
rabstejnek, line 295: Should we mark this with `@transaction.atomic`?

hawc/apps/lit/models.py
```diff
@@ -771,6 +762,27 @@ def delete_cache(cls, assessment_id: int, delete_study_cache: bool = True):
                 assessment_id, delete_reference_cache=False
             )
 
+    @classmethod
+    def update_hero_metadata(cls, assessment_id: int) -> ResultBase:
```
rabstejnek, line 766: Mark with `@transaction.atomic`?

hawc/apps/lit/serializers.py
shapiromatron: it won't work here b/c this method just schedules the tasks, but i'll add to the tasks

shapiromatron:
done & ready to merge @rabstejnek

PR #268, opened 2020-08-04
DTXSID integration

rabstejnek:
DTXSID is the EPA's preferred chemical identifier. The older standard, CASRN, is known to have a number of errors and it's not really clear that it's an open standard (it's owned by the American Chemistry Council). Therefore, we want to add the DTXSID to HAWC as a new chemical identifier. This will not replace the CASRN, but will rather be an additional identifier. eg: https://comptox.epa.gov/dashboard/dsstoxdb/results?search=DTXSID2020006
The goal for this work is to add DTXSID to all cases where a chemical exposure is avaiable.

1. add new table, assessment.DSSToxSubstanceId. The primary key should be the dtxsid, eg., DTXSID2020006. Casrn should be charfield, along w/ created, updated, etc. Also create a field, maybe a TextField, to save results from assessment.tasks.get_dsstox_details . Additional fields can be added, as appropriate.
2. Add M2M relation to DTXSID to assessment.Assessment, and Foreign Key to animal.Experiment, epi.Exposure invitro.IVChemical
3. Add a migration which maps all existing CAS to add DTXSID; migration should load a CSV and there will be a lookup CASRN -> DTXSID
4. Add generic admin view to browse DTXSID w/ search for CAS/DTXSID
5. Modify /assessment/public/?dtxsid=DTXSID2020006 that filters public assessment list with assessments with that id
6. Update APIs to include this new information, and FlatExports used in data pivots and heatmaps


rabstejnek:
[Working diagram for proposed changes](https://dbdiagram.io/d/5f29af9e7543d301bf5de2ab)

hawc/main/urls.py
```diff
@@ -74,6 +74,11 @@
         views.AdminAssessmentSize.as_view(),
         name="admin_assessment_size",
     ),
+    url(
```
shapiromatron, line 77: let's chat about this a bit; I was thinking of just making the standard vanilla admin views which require very little to no programming effort

rabstejnek:
Looks good!

PR #269, opened 2020-08-13
Data pivot fix: hide based on response instead of n

rabstejnek:


shapiromatron:
@rabstejnek make a tweak to catch an edge case and make the check even more promiscuous to fix other cases- look ok?

PR #277, opened 2020-08-24
Job model

rabstejnek:
- [x] Create a new job framework where a long-running task can be executed and results can be downloaded:
- [x] create api endpoint to submit job and poll status completion
- [x] create a diagnosis job type
    - based on kwargs, should be able to pass successfully and also fail; ensure that if an error is thrown in the job is captured
- [x] add unit tests using our diagnosis job type to make sure rest api works as expected
- [x] add a celery task to delete jobs older than 7 days old
- [x] add to our cron list, https://github.com/shapiromatron/hawc/blob/master/hawc/main/celery.py#L22


hawc/main/celery.py
```diff
@@ -28,4 +29,8 @@ def debug_task(self):
         "task": "hawc.apps.lit.tasks.update_pubmed_content",
         "schedule": 3600,
     },
+    "assessment-delete_old_jobs-1-week": {
+        "task": "hawc.apps.assessment.tasks.delete_old_jobs",
+        "schedule": timedelta(weeks=1),
```
rabstejnek, line 34: what do you prefer, time in seconds or the use of `timedelta`? Just so I can make it uniform

hawc/apps/assessment/serializers.py
```diff
@@ -192,3 +192,15 @@ class DatasetSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.Dataset
         fields = "__all__"
+
+
+class JobSerializer(serializers.ModelSerializer):
```
rabstejnek, line 197: should I allow string representations of jobs in POST requests? `{"job":"TEST"}` and `{"job":1}` are both valid?

hawc/apps/assessment/models.py
```diff
@@ -737,10 +737,64 @@ def try_read_df(cls, data, suffix: str, worksheet_name: str = None) -> pd.DataFr
...
+    assessment = models.ForeignKey(
+        Assessment, null=True, blank=True, on_delete=models.CASCADE, related_name="jobs"
+    )
+    status = models.PositiveSmallIntegerField(
+        choices=STATUS_CHOICES, default=PENDING, editable=False
+    )
+    job = models.PositiveSmallIntegerField(choices=JOB_CHOICES, default=TEST)
+
+    kwargs = JSONField(default=dict)
+    result = models.TextField(blank=True, editable=False)
```
rabstejnek, line 768: both `result` and `exception` fields are TextFields instead of JSONFields, my thought is that this way the task return doesn't have to be in JSON form and it can be parsed on a per job basis. thoughts?

rabstejnek:
Should we bother with an admin page, or is it not really necessary since these are only semi permanent? and any other job function to add besides the test one for now?

hawc/apps/assessment/models.py
shapiromatron: i think we can make them JSON fields, if we need to save binary data we'll probably write files and then we can just link to the file in the JSON spec

hawc/apps/assessment/models.py
```diff
@@ -737,10 +737,64 @@ def try_read_df(cls, data, suffix: str, worksheet_name: str = None) -> pd.DataFr
...
+    STATUS_CHOICES = (
+        (1, "PENDING"),
+        (2, "SUCCESS"),
+        (3, "FAILURE"),
+    )
+
+    TEST = 1
+    JOB_CHOICES = ((1, "TEST"),)
+
+    def test(self, fail=False):
```
shapiromatron, line 753: move the test  task outside of the class definition

hawc/apps/assessment/api.py
```diff
@@ -445,3 +445,12 @@ def get(self, request, casrn: str, format=None):
             tasks.get_dsstox_details.delay(casrn)
 
         return Response(data)
+
+
+class JobViewset(viewsets.ModelViewSet):
```
shapiromatron, line 450: if we have an assessment key on each job, can we add the standard assessment level permissions?

hawc/apps/assessment/models.py
```diff
@@ -737,10 +737,64 @@ def try_read_df(cls, data, suffix: str, worksheet_name: str = None) -> pd.DataFr
...
+
+    def set_result(self, result):
+        self.result = result
+        self.status = self.SUCCESS
+
+    def execute(self):
+        run_job.apply_async(task_id=self.task_id)
+
+    def get_func(self):
+        if self.job == self.TEST:
```
shapiromatron, line 789: else, raise an exception

hawc/apps/assessment/serializers.py
shapiromatron: does this work? https://stackoverflow.com/a/28954424

I'm fine with whatever the default behavior is; we can make it easier to use in the client with an enum if we need to

hawc/apps/assessment/signals.py
```diff
@@ -45,3 +46,19 @@ def invalidate_endpoint_cache(sender, instance, **kwargs):
     SerializerHelper.clear_cache(
         apps.get_model("animal", "Endpoint"), {"assessment_id": instance.id}
     )
+
+
+@receiver(pre_save, sender=models.Job)
```
shapiromatron, line 51: do we need this method? I think you can add `default=uuid.uuid4` on the model class and that should do it

https://docs.djangoproject.com/en/3.1/ref/models/fields/#uuidfield

hawc/apps/assessment/tasks.py
```diff
@@ -2,13 +2,34 @@
...
+    # delete jobs where "last_updated" > 1 week old
+    week_old = timezone.now() - timedelta(weeks=1)
+    models.Job.objects.filter(last_updated__lte=week_old).delete()
+
+
+@shared_task(bind=True)
+def run_job(self):
+    job = models.Job.objects.get(pk=self.request.id)
+    try:
+        result = job.get_func()(**job.kwargs)
```
shapiromatron, line 26: can this call `job.execute(**job.kwargs)` instead?

hawc/apps/assessment/models.py
```diff
@@ -737,10 +737,64 @@ def try_read_df(cls, data, suffix: str, worksheet_name: str = None) -> pd.DataFr
...
+
+    def set_exception(self, exception):
+        self.exception = exception
+        self.status = self.FAILURE
+
+    def set_result(self, result):
+        self.result = result
+        self.status = self.SUCCESS
+
+    def execute(self):
```
shapiromatron, line 785: i think the post-save signal should `run_job.apply_async(task_id=self.task_id)`, and then the execute method here can essentially do the get_func behavior.. We should also indicate that execute is expected to return something, I'd probably say it should return a `Dict` since that's we'll save the result field

hawc/main/celery.py
shapiromatron: oh i like timedelta. Can we do this daily instead? And perhaps the pubmed content can be daily too; hourly seems too frequent

hawc/apps/assessment/api.py
rabstejnek: Can we expect each job to have an associated assessment, or can assessment be null?

shapiromatron: ahh good point. sounds good. a uuid is pretty hard to guess, so we shouldn't be worried about permissions, and there's no really sensitive info here anyways

hawc/apps/assessment/admin.py
```diff
@@ -150,6 +150,11 @@ def __init__(self, *args, **kwargs):
         self.list_display_links = []
 
 
+@admin.register(models.Job)
+class JobAdmin(admin.ModelAdmin):
+    list_display = ("task_id", "assessment", "job", "kwargs", "status", "result")
```
shapiromatron, line 155: list_filter status?

hawc/apps/assessment/models.py
```diff
@@ -775,11 +776,97 @@ def try_read_df(cls, data, suffix: str, worksheet_name: str = None) -> pd.DataFr
         return df
 
 
+class Job(models.Model):
+    PENDING = 1
```
shapiromatron, line 780: use `IntChoiceEnum` for status and job: https://github.com/shapiromatron/hawc/blob/master/hawc/apps/vocab/models.py#L16

hawc/apps/assessment/models.py
```diff
@@ -775,11 +776,97 @@ def try_read_df(cls, data, suffix: str, worksheet_name: str = None) -> pd.DataFr
...
+        by the instance's "job" and "kwargs" properties, respectively.
+
+        Returns:
+            Any: Any data returned by the function call.
+            {"data" : <return_value>} MUST be JSON serializable.
+        """
+        func = self.get_func()
+        return func(**self.kwargs)
+
+    def get_func(self) -> Callable:
```
shapiromatron, line 825: do we need get_func? we could just add to execute, it's only used there, unless there's a reason to separate it

hawc/apps/assessment/api.py
```diff
@@ -52,6 +52,30 @@ def get_assessment_from_query(request) -> Optional[models.Assessment]:
     return models.Assessment.objects.filter(pk=assessment_id).first()
 
 
+class AssessmentForeignKeyPermissions(permissions.BasePermission):
```
shapiromatron, line 55: let's add a comment to describe what this is doing. I'd probably call it JobPermissions or something b/c it's really specialized to this use case at this point

hawc/apps/assessment/api.py
```diff
@@ -52,6 +52,30 @@ def get_assessment_from_query(request) -> Optional[models.Assessment]:
...
+        elif view.action == "create":
+            serializer = view.serializer_class(data=request.data)
+            serializer.is_valid(raise_exception=True)
+            assessment = serializer.validated_data.get("assessment")
+            if assessment is None:
+                return bool(request.user and request.user.is_staff)
+            else:
+                return assessment.user_can_edit_object(request.user)
+        else:
+            return True
```
shapiromatron, line 76: return False? safer that way

hawc/apps/assessment/api.py
```diff
@@ -384,6 +408,22 @@ def endpoints(self, request, pk: int = None):
         serializer = serializers.AssessmentEndpointSerializer(instance)
         return Response(serializer.data)
 
+    @action(
+        detail=True, methods=("get", "post"),
+    )
+    def jobs(self, request, pk: int = None):
```
shapiromatron, line 414: nice job

hawc/apps/assessment/api.py
rabstejnek: All the other actions not checked by `has_permission` are object specific, so I have to return True to allow `has_object_permission` to be checked. I suppose I could do `return request.user.is_authenticated` as a first pass to make sure the user is logged in?

shapiromatron: ok, that's fine, but let's add a comment to indicate why we return True

PR #278, opened 2020-08-24
Log and Blog Models

rabstejnek:
### Log
- [x] add methods to add logs sync
- [x] add django admin command to delete logs > 1 yr old
- [x] add admin view for viewing logs
- [x] add api read-only api view for viewing assessment logs

### Blog
- [x] add admin view for CRUD operations
- [x] add beta-tester public view to show published data
- [x] assume content is in myst format


hawc/apps/assessment/models.py
```diff
@@ -737,10 +740,41 @@ def try_read_df(cls, data, suffix: str, worksheet_name: str = None) -> pd.DataFr
...
+    rendered_content = models.TextField(editable=False)
+    published = models.BooleanField(default=False)
+    created = models.DateTimeField(auto_now_add=True)
+    last_updated = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        ordering = ("-created",)
+
+
+@receiver(pre_save, sender=Blog)
```
shapiromatron, line 767: move signals to `hawc/apps/assessment/signals.py`

hawc/apps/assessment/api.py
```diff
@@ -445,3 +456,36 @@ def get(self, request, casrn: str, format=None):
             tasks.get_dsstox_details.delay(casrn)
 
         return Response(data)
+
+
+class BlogViewset(viewsets.ReadOnlyModelViewSet):
```
shapiromatron, line 461: remove blog api endpoint for now; there's no use-case for that one

hawc/apps/assessment/api.py
```diff
@@ -445,3 +456,36 @@ def get(self, request, casrn: str, format=None):
...
+    model = models.Blog
+    serializer_class = serializers.BlogSerializer
+    pagination_class = None
+
+    def get_queryset(self):
+        return self.model.objects.filter(published=True)
+
+
+class LogViewset(viewsets.ReadOnlyModelViewSet):
```
shapiromatron, line 470: some refactoring of this one - would this be easy to do?

- I'd make the list view `/assessment/log/` return system logs if IsAdminUser
- I'd make the detail view `/assessment/log/1/` return logs for assessment 1. 

Alternatively, you could add a new detail action 'logs' on `AssessmentViewset` that returns assessment logs. That's probably easier to do and would require less customization

hawc/apps/assessment/urls.py
```diff
@@ -8,6 +9,8 @@
 router.register(r"assessment", api.Assessment, basename="assessment")
 router.register(r"dashboard", api.AdminDashboardViewset, basename="admin_dashboard")
 router.register(r"dataset", api.DatasetViewset, basename="dataset")
+router.register(r"global/blogs", api.BlogViewset, basename="blogs")
```
shapiromatron, line 12: i think the blog route will be deleted, and we can change the log route to `logs`

hawc/apps/assessment/views.py
```diff
@@ -545,3 +546,15 @@ class AdminAssessmentSize(TemplateView):
...
+
+
+# log / blog
+class BlogList(ListView):
+    model = models.Blog
+
+    def get_queryset(self):
+        return self.model.objects.filter(published=True)
+
+    @method_decorator(beta_tester_required)
```
shapiromatron, line 558: nice!

hawc/apps/assessment/models.py
```diff
@@ -737,10 +740,41 @@ def try_read_df(cls, data, suffix: str, worksheet_name: str = None) -> pd.DataFr
...
+def render_content(sender, instance, *args, **kwargs):
+    instance.rendered_content = to_html(instance.content)
```
shapiromatron, line 769: might want to catch an exception here and set to some value `<h1>Error - myst_parser Parsing error</h1>`, assuming the parser can crash

hawc/apps/assessment/api.py
shapiromatron: Update - looks like you already did the suggestion! I'd just make the first bullet then. I'd use the `ListModelMixin` here with a standard `viewsets.ViewSet`, and then just handle the list view. We probably don't need a detail view for this one, and I'd just set the permission_class to (IsAdminUser,); users can access the assessment logs via a different API

hawc/apps/assessment/admin.py
```diff
@@ -145,3 +147,28 @@ class TimeSpentEditingAdmin(admin.ModelAdmin):
...
+        deleted, _ = queryset.filter(last_updated__lte=year_old).delete()
+        # send a message with number deleted
+        self.message_user(request, f"{deleted} of {queryset.count()} selected logs deleted.")
+
+    delete_gt_year.short_description = "Delete 1 year or older"
+
+
+@admin.register(models.Blog)
+class BlogAdmin(admin.ModelAdmin):
+    list_display = ("subject", "published", "created", "last_updated")
```
shapiromatron, line 172: Add `list_filter = ("published",)` and for `content` for search_fields

hawc/apps/assessment/urls.py
rabstejnek: we are also going to be adding the option for jobs not to be associated with an assessment, so would it be cleaner to keep the `global` namespace for these kind of endpoints? or would you rather the url paths just be `logs` and `jobs`?

shapiromatron: I prefer logs and blogs, since the `dashboard` route is the same kind of thing... but no really strong preference from me if you want to namespace them

rabstejnek: Fair enough! I'll use `logs` and `jobs`

tests/apps/assessment/test_assessment_models.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+from django.test.client import Client
+from django.urls import reverse
+from pytest_django.asserts import assertTemplateUsed
+
+from hawc.apps.assessment.models import Blog
+
+
+@pytest.mark.django_db
+def test_assessment_creation():
```
shapiromatron, line 10: `test_assessment_creation` think this one can be deleted; it's been moved from the dtxid branch

PR #279, opened 2020-08-27
modernize frontend lit application

shapiromatron:
The literature frontend code is some of the earliest code written in HAWC, and it is due for an upgrade to follow more modern conventions. 

- [x] Change jQuery spaghetti to use react/mobx
- [x] Standardize webpack startup to use a new `frontend/index.js` file

This should make it easier to maintain in the future.


rabstejnek (changes requested):
# Tags are not working correctly
## They do not show on reference
image
The reference shows up in a search, but the tags are not displayed. So the tag search works.
## They also do not show up on "Edit tags" page
image
Adding a tag here seems to replace all of the tags for that reference. The reference will now show up in a search for that added tag, but it is still read elsewhere on the frontend as untagged.

@shapiromatron Let me know if I need to clarify this issue more.

shapiromatron:
Thanks @rabstejnek, for comment 1 above, I agree, we should fix that. I think we just need to set a different prop to the renderer. For comment 2, I'm not sure I fully follow, can you provide steps to recreate the issue?

rabstejnek:
tags
@shapiromatron It's probably the same problem, just that tags also aren't showing up on the "Edit tags" page, even when they're set.

shapiromatron:
@rabstejnek updated addressing review. I fixed 1) above. I didn't fix 2). This is because I see the tagging process as a two-step process. By default, "truth" is in the box w/ the red labels. But tags are mutable, so we don't show in the tags below. After tagging is complete, a user can save, and then changes all tags and makes them persist on the server. Let me know if that doesn't make sense and I can explain more.

rabstejnek (approved):
Looks good!

PR #283, opened 2020-08-28
Ordered queryset

shapiromatron:
- [x] Fix non-deterministic behavior in tests and on server where querysets return in non-deterministic order
- [x] Add migration check to alert staff on which visuals may have changed based on prior behavior

Screenshot from 2020-08-28 13-00-25


hawc/apps/riskofbias/models.py
```diff
@@ -355,7 +355,8 @@ def get_dp_export(
         for metric_id in metric_ids:
             for study_id in study_ids:
                 key = (study_id, metric_id)
-                if key in scores_map:
+                if key in scores_map and not isinstance(scores_map[key], str):
```
shapiromatron, line 358: the ordering uncovered a bug

PR #284, opened 2020-08-31
Selenium on GitHub CI

rabstejnek:


tests/conftest.py
```diff
@@ -97,3 +106,44 @@ def rewrite_data_files():
     A test exists in CI to ensure that this flag is set to False on commit.
     """
     return False
+
+
+def _wait_until_webpack_ready(max_wait_sec: int = 60):
```
shapiromatron, line 111: @rabstejnek changed slightly here to wait intelligently rather than sleeping; hopefully this means we'll get less random pipeline failures, though w/ selenium we'll surely get some....

shapiromatron (approved):
Made a few commits, the only major change was replacing the sleep 30 with a healthcheck

PR #285, opened 2020-08-31
hotfix - Cast queryset to list for JSON serializability with celery task messa…

rabstejnek:
…ges on production.


rabstejnek:
Problem originated from #262 . These were the only instances of `values_list` that needed to be cast to list.

PR #287, opened 2020-09-14
Tool integration

rabstejnek:
Bringing in `litter_getter` and `docxUtils` into HAWC.


hawc/services/utils/misc.py
```diff
@@ -0,0 +1,50 @@
+import re
+from typing import Any, List
+
+# First-group is rest of reference; second-group are initials
+# with optional second initial (middle-name) with optional periods
+RE_AUTHOR = re.compile(r"([\-\,\w]+)\s([\s?\w{1,1}\.?]+)$", flags=re.UNICODE)
+RE_TWO_INITIALS = re.compile(r"^([\w])[\,\.]{0,2}\s([\w])[\,\.]{0,2}$", flags=re.UNICODE)
+
+
+def normalize_author(author: str) -> str:
```
shapiromatron, line 10: rename this module to `authors.py`?

hawc/services/utils/misc.py
```diff
@@ -0,0 +1,50 @@
...
+        return str(authors[0])
+    elif n_authors == 2:
+        return "{0} and {1}".format(*authors)
+    elif n_authors == 3:
+        return "{0}, {1}, and {2}".format(*authors)
+    else:  # >3 authors
+        return f"{authors[0]} et al."
+
+
+def try_int(val) -> Any:
```
shapiromatron, line 46: this is the only oddball - move to a new `types.py`?

hawc/tools/docx_utils/__init__.py
```diff
@@ -0,0 +1,4 @@
+__version__ = "0.0.2"
```
shapiromatron, line 1: remove `__version__`

hawc/tools/docx_utils/reports.py
```diff
@@ -0,0 +1,73 @@
+import abc
+import os
+from io import BytesIO
+
+from docx import Document
+from docx.enum.section import WD_ORIENT
+from docx.shared import Inches
+
+
+class DOCXReport(object):
```
shapiromatron, line 10: remove `(object)` parent class for this and other-things in docx_utils, old python2 convention.

Looking forward to adding type annotations to this once we start using it.

hawc/tools/docx_utils/tables.py
```diff
@@ -0,0 +1,217 @@
...
+from docx.oxml.shared import OxmlElement, qn
+from docx.shared import Inches
+
+
+class TableMaker(object):
+    """
+    Helper-object to build tables
+    """
+
+    def __init__(self, colWidths, styles=None, numHeaders=1, tblStyle=None, firstRowCaption=True):
```
shapiromatron, line 11: update signature format, here and elsewhere in this file

pytest.ini
```diff
@@ -1,4 +1,7 @@
 [pytest]
 addopts = --ds=hawc.main.settings.unittest --reuse-db --nomigrations
 python_files = tests.py test_*.py *_tests.py
-testpaths = tests
+testpaths =
```
shapiromatron, line 4: just curious, why is this change required? it looks like it skips integration tests- include?

tests/hawc/services/nih/test_pubmed.py
```diff
@@ -0,0 +1,206 @@
...
+
+from hawc.services.nih import pubmed
+
+PUBMED_API_KEY = os.environ.get("PUBMED_API_KEY")
+
+
+@pytest.mark.skipif(
+    PUBMED_API_KEY is None, reason="Test environment requires $PUBMED_API_KEY to be set."
+)
+class PubMedSearchTests(TestCase):
```
shapiromatron, line 14: we should add `vcr` to these tests

shapiromatron, line 14: drop the TestCase parent class if you rename to `TestPubMed` and do things the "pytest way"

tests/hawc/services/nih/test_pubmed.py
```diff
@@ -0,0 +1,206 @@
...
+    def setUp(self):
+        pubmed.settings.connect(PUBMED_API_KEY)
+        self.term = "science[journal] AND breast cancer AND 2008[pdat]"
+        self.results_list = [19008416, 18927361, 18787170, 18487186, 18239126, 18239125]
+
+    def test_standard_query(self):
+        self.search = pubmed.PubMedSearch(term=self.term)
+        self.search.get_ids_count()
+        self.search.get_ids()
+        self.assertEqual(self.search.request_count, 1)
```
shapiromatron, line 24: change to `assert self.search.request_count == 1`; you'll need to do this for the rest too

tests/hawc/services/utils/test_misc.py
```diff
@@ -0,0 +1,63 @@
+from hawc.services.utils.misc import get_author_short_text, normalize_author
+
+
+class TestNormalizeAuthors:
```
shapiromatron, line 4: this should be `test_authors.py` after the rename

shapiromatron (changes requested):
little stuff, looking great!

pytest.ini
rabstejnek: The integration tests are under frontend now, so they're still run. I had to make this change because after the restructure pytest would run things in a different order and cause problems. If we separate the paths like this pytest won't mix running the different kinds of tests we have, aka it will run all of the hawc tests, then integration tests, then client tests.

shapiromatron: ok, sounds good. thanks!

hawc/services/utils/misc.py
rabstejnek: Its only used once, so I just moved it into `ris.py`

shapiromatron (approved):

PR #288, opened 2020-09-15
Add published field to datasets

rabstejnek:


rabstejnek:
`published` is now an option for datasets.
Published datasets are only visible to users with edit permissions (team members, project managers, and staff), and this visibility applies to the list view on the assessment detail page, the detail view for individual datasets, and API endpoints for datasets.

hawc/apps/assessment/api.py
```diff
@@ -447,6 +447,20 @@ class DatasetViewset(AssessmentViewset):
     serializer_class = serializers.DatasetSerializer
     assessment_filter_args = "assessment_id"
 
+    def check_object_permissions(self, request, obj):
+        if obj.assessment.user_can_edit_object(request.user) or obj.published:
+            return super().check_object_permissions(request, obj)
+        else:
+            raise PermissionDenied
```
shapiromatron, line 454: while either works, we commonly raise an instance of the class, not the class

hawc/apps/assessment/views.py
```diff
@@ -351,25 +356,34 @@ def get_success_url(self):
 
 
 # Dataset views
-class DatasetCreate(BaseCreate):
+class DatasetPublishedMixin:
+    def get_object(self, **kwargs):
+        obj = super().get_object(**kwargs)
+        if obj.assessment.user_can_edit_object(self.request.user) or obj.published:
```
shapiromatron, line 362: this if statement is used in a few places. Should we move it to the dataset instance?

```python
def user can view(self, user):
    return self.assessment.user_can_edit_object(user) or self.published

...

if not obj.user_can_view(self.request.user):
    raise PermissionDenied()
```

hawc/apps/assessment/views.py
```diff
@@ -351,25 +356,34 @@ def get_success_url(self):
...
+            return obj
+        else:
+            raise PermissionDenied
+
+
+class DatasetCreate(DatasetPublishedMixin, BaseCreate):
```
shapiromatron, line 368: we dont need this mixin for any of the create/update/delete views; the base classes check permissions that users can edit content in an assessment. I think you can just move it to the Read view

hawc/apps/assessment/api.py
shapiromatron: (see comment below about moving this logic to instance)

shapiromatron (changes requested):

hawc/apps/assessment/views.py
rabstejnek: I ended up doing 
```
def user_can_view(self, user) -> bool:
        return (
            self.published
            and self.assessment.user_can_view_object(user)
            or self.assessment.user_can_edit_object(user)
        )
```
There is some redundancy with the `user_can_view_object` check and the assessment level permissions, but I figure thats more correct than this method relying on those base permissions already being in place.

shapiromatron: nice job

hawc/apps/assessment/views.py
```diff
@@ -362,6 +367,13 @@ class DatasetCreate(BaseCreate):
 class DatasetRead(BaseDetail):
     model = models.Dataset
 
+    def get_object(self, **kwargs):
+        obj = super().get_object(**kwargs)
+        if obj.user_can_view(self.request.user):
```
shapiromatron, line 372: this is nitpicky, but it's nice to have bailout conditions instead of these conditions. Propose changing this and the api to something like this:

```python
if not obj.user_can_view(self.request.user):
    raise PermissionDenied()
return obj
```

https://www.itamarweiss.com/personal/2018/02/28/return-early-pattern.html

hawc/apps/assessment/api.py
```diff
@@ -447,6 +447,20 @@ class DatasetViewset(AssessmentViewset):
     serializer_class = serializers.DatasetSerializer
     assessment_filter_args = "assessment_id"
 
+    def check_object_permissions(self, request, obj):
+        if obj.user_can_view(request.user):
```
shapiromatron, line 451: see comment below

shapiromatron (changes requested):

shapiromatron (approved):

PR #289, opened 2020-09-17
umls webservices

shapiromatron:
- task A
    - add to hawc.services.nih.umls.py
    - build new client for importing UMLS item by id
    - add new method in vocab.models.Entity to load a umls item
        - determine if we need to add new metadata fields (eg., synonyms, etc)

- task B
    - on the `controlled-vocab-2` branch, we have a fixture of vocab.models.Term data
    - we want to programmatically propose TermEntity relations
    - originally, just add those w/ high confidence


scott-8:
@shapiromatron

PR #292, opened 2020-09-21
Fix tagtree navbar on update

rabstejnek:
## Before
tagtree-update-broken
## After
tagtree-update


rabstejnek:
@shapiromatron I am going to hold off on replacing the old navbar with the new React one until I have more time to hunt down all of the other visuals that use the old one.

shapiromatron (approved):

PR #296, opened 2020-09-29
Migrations to set DTXSID on experiments / IV chemicals

rabstejnek:
Old migrations did not set DTXSIDs on experiments and IV chemicals correctly. New migrations were created to accomplish this, and the old migrations were cleaned up to remove the broken implementation with functionality left unchanged.


shapiromatron:
@rabstejnek see #297 as an alternative proposal to this one.

rabstejnek:
Closing in favor of #297

PR #298, opened 2020-09-30
Hotfix: reference search permissions

rabstejnek:
`AssessmentLevelPermissions` required edit permissions on the reference search API when it should have been read permissions; the subclassed permissions were removed and instead checked locally.


shapiromatron:
This looks great, do you think we should make a new permissions class that just bakes in the `assessment.user_can_view_object(request.user)`, I bet that would be reused in other places too

shapiromatron (changes requested):
Per the comments and our chat yesterday, let's create a new API permissions class to use instead

shapiromatron (approved):

PR #299, opened 2020-09-30
Hotfix: preformatted strings on study table

rabstejnek:
### Before
image

### After
image

### DIFF
--- a/frontend/study/Study.js
+++ b/frontend/study/Study.js
@@ -151,7 +151,7 @@ class Study {
         tbl.add_tbody_tr("Funding source", this.data.funding_source);
         tbl.add_tbody_tr("Study identifier", this.data.study_identifier);
         tbl.add_tbody_tr("Author contacted?", HAWCUtils.booleanCheckbox(this.data.contact_author));
-        tbl.add_tbody_tr_pre("Author contact details", this.data.ask_author);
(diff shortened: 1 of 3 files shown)


rabstejnek:
@shapiromatron  Do you think this inline style fix is appropriate? We could instead make changes to the JS:
https://github.com/shapiromatron/hawc/blob/6d3fee1721509093893beda48f58293a0cdda12c/frontend/study/Study.js#L154
Maybe a third parameter, `preformatted: bool` ?

shapiromatron:
> @shapiromatron Do you think this inline style fix is appropriate? We could instead make changes to the JS:
> https://github.com/shapiromatron/hawc/blob/6d3fee1721509093893beda48f58293a0cdda12c/frontend/study/Study.js#L154
> 
> 
> Maybe a third parameter, `preformatted: bool` ?

Yes, that or just a new `tbl.add_tbody_tr_pre` method; this jQuery stuff will be rewritten in the future

shapiromatron (approved):
@rabstejnek I ended up changing it to an option as you suggested

shapiromatron:
@rabstejnek please review to make sure you like my change; since options already existed with this method, it made sense just to add another one instead of duplicated a complex method

PR #302, opened 2020-10-02
Delete orphaned tags

rabstejnek:
Added an assessment admin action to delete orphaned reference tags.

### DIFF
--- a/hawc/apps/assessment/admin.py
+++ b/hawc/apps/assessment/admin.py
@@ -1,10 +1,10 @@
 from datetime import timedelta
 
+from django.apps import apps
 from django.contrib import admin, messages
 from django.utils import timezone
 from django.utils.html import format_html
 
-from ..lit.models import ReferenceFilterTag, ReferenceTags
 from . import models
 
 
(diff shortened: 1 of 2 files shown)


hawc/apps/assessment/admin.py
```diff
@@ -33,12 +34,30 @@ class AssessmentAdmin(admin.ModelAdmin):
...
     )
 
-    actions = (bust_cache,)
+    actions = (bust_cache, "delete_orphan_tags")
 
     def queryset(self, request):
         qs = super().queryset(request)
         return qs.prefetch_related("project_manager", "team_members", "reviewers")
 
+    def delete_orphan_tags(self, request, queryset):
```
shapiromatron, line 43: most of this logic should be in in a lit manager object (for Reference or ReferenceTag); keep the biz logic outside the admin

hawc/apps/assessment/admin.py
```diff
@@ -33,12 +34,30 @@ class AssessmentAdmin(admin.ModelAdmin):
...
+            )
+            return
+        assessment = queryset.first()
+        root = ReferenceFilterTag.get_assessment_root(assessment.id)
+        # queryset of tag tree associated with assessment
+        filter_tags = ReferenceFilterTag.get_tree(root)
+        # queryset of tags associated with assessment
+        tags = ReferenceTags.objects.filter(content_object__assessment_id=assessment.id)
+        # delete tags that are not in the assessment tag tree
+        deleted, _ = tags.exclude(tag__in=filter_tags).delete()
```
shapiromatron, line 57: we should probably log somewhere what was deleted in case it ever needs to be recovered. Perhaps we can write in the assessment.Log model an item which just lists hte reference-id and tag-id of what was deleted?

shapiromatron (changes requested):
Few minor tweaks, but the good news is we get to use the assessment.Log model you created!

rabstejnek:
New success message:
image

Created log message:
image

shapiromatron (approved):

PR #305, opened 2020-10-04
EHV table query highlighting

rabstejnek:
Text matching the query is now highlighted in the EHV table:
image


frontend/animal/EhvBrowser/Table.js
```diff
@@ -7,6 +7,24 @@ const MAX_ROWS = 150;
 @inject("store")
 @observer
 class Table extends Component {
+    highlight(text) {
```
shapiromatron, line 10: I did this recently, would it make sense to make a `highlightPatternText` method?

https://github.com/shapiromatron/hawc/blob/8f7f74a10b61350bc3227de0edb615b344ee89fd/frontend/assets/shared/components/AutocompleteTerm/index.js#L78

You could even create the regex  outside the function call, so it'd be faster.

shapiromatron (changes requested):
consider using the `boldPatternText` function pattern which uses a regex instead

shapiromatron (approved):
@rabstejnek look ok?

rabstejnek:
@shapiromatron looks good!

PR #306, opened 2020-10-04
Update the term help-text on forms

rabstejnek:


rabstejnek:
image

frontend/vocab/constants.js
```diff
@@ -0,0 +1,6 @@
+const VOCAB_HELP_TEXT = {
```
rabstejnek, line 1: @shapiromatron My thought is that we can control how the namespaces are shown in different contexts with this file, similar to what we do with RoB. Let me know if you'd rather me just use `assessment.get_vocabulary_display()` and I'll remove this.

shapiromatron: yeah, thats what I was thinking originally too, but this is such a simple pattern and easy to track we can use it for now

shapiromatron (approved):

PR #307, opened 2020-10-05
ensure unsupported browser popup presented for IE11

shapiromatron:
Tested to confirm error still reports on IE11: https://gist.github.com/shapiromatron/bae66a34e30c8d6c518676f39592fed4

Had to make a few changes for compatibility:

- explicitly give https:// to external domains, not http
- moved code outside of webpack to ensure code works on <ES6 browsers
- disabled session option to hide warning
- removed `urlify.js` (while cleaning up the base.html)
- moved check to base.html instead of portal.html


rabstejnek:
Working on my end.

IE 11:
image

Chrome:
image

rabstejnek (approved):

PR #309, opened 2020-10-05
Endpoint term API

rabstejnek:
Validation checks:
- Assessment vocab matches term namespace
- Term type matches endpoint field
- Cannot set both term and non-term counterpart: they're mutually exclusive


hawc/apps/animal/models.py
```diff
@@ -1117,11 +1117,19 @@ def get_crumbs(self):
 
     def save(self, *args, **kwargs):
         # ensure our controlled vocabulary terms don't have leading/trailing whitespace
-        self.system = self.system.strip()
-        self.organ = self.organ.strip()
-        self.effect = self.effect.strip()
-        self.effect_subtype = self.effect_subtype.strip()
-        self.name = self.name.strip()
+        self.system = (
```
rabstejnek, line 1120: @shapiromatron I noticed `save` was already overwritten on `Endpoint`, so I just went with that instead of making a pre-save signal. The formatting makes this look a little messy, lmk if you want me to restructure it so that it doesn't use ternary, just so it reads better.

rabstejnek: I changed it back to the way it was; so now you have the leeway of renaming non-terms manually through the form, but the API remains a little more strict in its usage.

hawc/apps/animal/serializers.py
```diff
@@ -295,14 +297,62 @@ def to_representation(self, instance):
 
         return ret
 
+    def _validate_term_and_text(self, data, term_field: str, text_field: str, term_type_str: str):
```
shapiromatron, line 300: Rewrote the field-level validation and made it instance-level since the validation is more than single-dependent. Also made it easier to cache and update things

tests/hawc/apps/animal/test_animal_serializer.py
```diff
@@ -118,3 +124,96 @@ def test_dose_group_failures(self):
         for data in datasets:
             serializer = DosingRegimeSerializer(data=data)
             assert serializer.is_valid() is False
+
+
+@pytest.mark.django_db
+class TestEndpointSerializer:
```
shapiromatron, line 130: I figured out a more direct way to test the serializer, so moved the api tests to the serializer. We may do the same with other tests too since that better isolates what we really want to test.

shapiromatron (approved):
LGTM if it LGTY @rabstejnek :)

PR #314, opened 2020-10-07
logscale axes

shapiromatron:
Regression w/ d3 v5 migration, logscale tick behavior changed:

---

### Error 

Screen Shot 2020-10-07 at 5 52 09 PM

---

### Fix 

Screen Shot 2020-10-07 at 5 52 17 PM

---

In addition, the following changes were made:

- standardized d3.format for numerical formatting data
- removed unused methods
- fixed bug in crossview plot where range styles were not applied (another d3 v5 issue)

### DIFF
--- a/frontend/animal/Barplot.js
+++ b/frontend/animal/Barplot.js
@@ -2,6 +2,7 @@ import $ from "$";
 import _ from "lodash";
 import * as d3 from "d3";
 
+import h from "shared/utils/helpers";
 import D3Plot from "utils/D3Plot";
 
 class Barplot extends D3Plot {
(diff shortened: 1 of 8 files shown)


rabstejnek (approved):
This edge case of domain `[0.1,1]` is a little off, but I don't think its bad enough that we have to take into account any changes:
image
My confusion comes from the fact that the axes have many more ticks than we pass to them, but they have an appropriate number of tick labels, so if you're fine with the striations between the tick labels then it looks good to me.

shapiromatron:
> This edge case of domain `[0.1,1]` has an extra arbitrary tick label, but I don't think its much of a concern:
> image
> My problem comes from the fact that the axes have many more ticks than the number we pass to them, but they have an appropriate number of tick labels, so if you're fine with the striations between the tick labels then it looks good to me.

Even when we recommend 2 ticks we still get 3 here; I think this is just the limit of d3. I dont think this is really going to be a problem in the real world b/c if you're only looking at an interval of 1 dose-unit then you generally dont need to use a logscale, and if a user gets an x-label I think that's should be ok.

PR #322, opened 2020-10-13
cache expensive part not entire page

shapiromatron:
The combination of a cache and login_required made the page-cache dependent on login rules. This caches the expensive part but still does the login and permissions check.


rabstejnek (approved):
Looks good!

PR #323, opened 2020-10-13
API to update endpoint terms

rabstejnek:


hawc/apps/animal/api.py
```diff
@@ -257,6 +259,22 @@ def rob_filter(self, request):
         serializer = self.get_serializer(qs, many=True)
         return Response(serializer.data)
 
+    @action(detail=False, methods=("post",))
+    def update_terms(self, request):
+        # data must have one or more endpoints
+        if len(request.data) == 0:
+            raise DRFValidationError("List of endpoints must be > 1")
+        # check assessment level permissions
+        endpoint = self.model.objects.get(pk=request.data[0].get("id"))
```
shapiromatron, line 268: This permissions check could be abused. You could pass ids from multiple assessments and just the first id on the list would be one you'd have permission to view.  It's probably better to pass the `assessment_id` and use the `get_assessment_id_param, and then do a permissions check.

Later in the update method we can validate that all endpoint ids are from this assessment (by passing the assessment into the update_terms method), but that's a  deeper check than what's done at the api level.

hawc/apps/animal/managers.py
```diff
@@ -297,6 +298,105 @@ def migrate_terms(self, assessment: Assessment) -> pd.DataFrame:
 
         return df
 
+    def update_terms(self, objs: List[Dict]) -> List:
```
shapiromatron, line 301: Add a detailed comment to explain what this method does.

Also pass the assessment_id and add a validation check if the set of endpoint id assessments don't all equal the assessment id.

hawc/apps/animal/managers.py
```diff
@@ -297,6 +298,105 @@ def migrate_terms(self, assessment: Assessment) -> pd.DataFrame:
...
+        term_ids = [id for (_, id) in term_fields_and_ids]
+        terms = Term.objects.filter(pk__in=term_ids)
+        for term in terms:
+            if term.namespace != assessment.vocabulary:
+                raise ValidationError(
+                    f"Assessment vocabulary {assessment.vocabulary} does not match term {term.id} namespace {term.namespace}"
+                )
+        # all terms must match assigned type
+        term_types = VocabularyTermType.as_dict()
+        term_type_to_field = {
```
shapiromatron, line 363: you've used these maps in lots of different methods - can we refactor this and put these maps somewhere shared?

hawc/apps/animal/managers.py
shapiromatron, line 301: add a `transaction.atomic` decorator to this

hawc/apps/animal/managers.py
```diff
@@ -297,6 +298,105 @@ def migrate_terms(self, assessment: Assessment) -> pd.DataFrame:
...
+            "system_term_id": "system",
+            "organ_term_id": "organ",
+            "effect_term_id": "effect",
+            "effect_subtype_term_id": "effect_subtype",
+            "name_term_id": "name",
+        }
+        actual_term_fields_and_ids = {
+            (term_type_to_field[term_types[term.type]], term.pk) for term in terms
+        }
+        for term_field_and_id in term_fields_and_ids:
```
shapiromatron, line 380: so i read over the spec and I think I wasn't clear here, my apologies.

A user can only provide an `endpoint_id` and a `name_term_id` for this api endpoint, and given that valid `name_term_id`, it will automatically change all other fields (system/organ/effect/effect-subtype/name) based on the parent of the name-term specified. 

It might be most efficient to just grab the `Term.ehv_dataframe` and then do queries off that, once you've established all the   name_term_id fields are valid.

Apologies with the imprecision here w/ my spec.

shapiromatron (changes requested):
looking good, but a few more changes to go...

hawc/apps/animal/api.py
rabstejnek: Part of the validation in `update_terms` makes sure all of the endpoints are in the same assessment, so just checking permissions on one of the endpoints should be good enough.

shapiromatron: I agree with you that that check already exists in effect, but it will be something that we will forget a year from now when we just look at the code. A little refactoring will be good to clearly show the separation of concerns - the api can check "can this user change this assessment",  the update method can check "are all ids in this assessment".

hawc/apps/animal/api.py
```diff
@@ -257,6 +259,20 @@ def rob_filter(self, request):
         serializer = self.get_serializer(qs, many=True)
         return Response(serializer.data)
 
+    @action(detail=False, methods=("post",))
+    def update_terms(self, request):
+        # data must have one or more endpoints
+        if len(request.data) == 0:
```
shapiromatron, line 265: this check can now be moved to the `update_terms` method

hawc/apps/animal/managers.py
```diff
@@ -297,6 +284,98 @@ def migrate_terms(self, assessment: Assessment) -> pd.DataFrame:
...
+        term_id_to_type_and_name = {term.id: (term.type, term.name) for term in all_terms}
+
+        # set endpoint terms
+        endpoint_ids = [obj["id"] for obj in objs]
+        endpoints = self.get_queryset().filter(pk__in=endpoint_ids)
+        type_to_text_field = VocabularyTermType.value_to_text_field()
+        type_to_term_field = VocabularyTermType.value_to_term_field()
+        updated_endpoints = []
+        updated_fields = list(type_to_text_field.values()) + list(type_to_term_field.values())
+        for endpoint in endpoints:
```
shapiromatron, line 365: With the term_id, we can set all the other EHV fields too

The `term. ehv_endpoint_name` method returns system, organ effect, etc for these which can also be updated.  It's probably more efficient to in terms of db queries to use the `Term.ehv_dataframe()` method, but you get the idea.

shapiromatron (commented):
few more notes on the WIP; its looking good...

hawc/apps/animal/managers.py
rabstejnek: I believe my implementation is pretty efficient, since I only access terms that I know will be assigned instead of all terms in the EHV. Also in my recursive function I don't retread the same terms if two name terms share part of their tree, so I think its more efficient in that way too. If you think it would be cleaner or read better to use these methods though I can go back and change my implementation to use them.

shapiromatron: The cache you're using looks good, but I think the issue is more in the database requests. If you update 10 different endpoint ids, then it requires 40 independent database queries since there are 4 parents. Building the dataframe is one (big) hit as opposed to 50 little ones.

rabstejnek: Per our talk I swapped in the `ehv_dataframe`, I think it makes the code look cleaner too!

shapiromatron (approved):
nice job!

PR #324, opened 2020-10-14
Upgrade to Django 3.1

rabstejnek:


hawc/apps/assessment/migrations/0024_auto_20201015_1002.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 3.1.2 on 2020-10-15 14:02
+
+from django.db import migrations, models
```
shapiromatron, line 3: rename this and other migration files to `00**_django31` or something similar

requirements/base.txt
```diff
@@ -1,17 +1,16 @@
...
-django-selectable==1.2.1
+django-filter==2.4.0
+django-redis==4.12.1
+django-reversion==3.0.8
+django-selectable==1.3.0
 django-taggit==1.2.0
 django-treebeard==4.3.1
 django-webpack-loader==0.7.0
-djangorestframework==3.11.0
+djangorestframework==3.11.1
```
shapiromatron, line 11: why not 3.12.1? https://pypi.org/project/djangorestframework/#history

hawc/apps/common/api.py
```diff
@@ -1,15 +1,85 @@
...
 from rest_framework.decorators import action
 from rest_framework.response import Response
 from rest_framework.throttling import UserRateThrottle
-from rest_framework_extensions.mixins import ListUpdateModelMixin
 
 from ..assessment.api import DisabledPagination, get_assessment_from_query
 from .helper import try_parse_list_ints
 
 
+class ListUpdateModelMixin:
```
shapiromatron, line 14: move this into it's own file. can you add a comment where you had to change the original code for 3.1 compatibility?

shapiromatron (commented):
a few minor changes

requirements/base.txt
rabstejnek: 3.11.1 provided support for Django 3.1, and 3.12.0 has a much larger [change log](https://www.django-rest-framework.org/community/release-notes/#3120). I just didn't want to fight a two front war in case 3.12.0 breaks anything, but now that everything else is fixed I can upgrade that too.

hawc/apps/common/api.py
shapiromatron: ideally, it woudl be in `hawc/apps/common/api/mixins.py`, but that means you'll need to do a little more refactoring of the api.py. We can import the things in the `api/__init__.py` so all the external imports would still work. do you think that's a good plan

shapiromatron (approved):

PR #325, opened 2020-10-14
hotfix - restore model ordering default when possible

shapiromatron:
A previous change introduced a bug where even if a model had a default ordering, it always default to `id` to prevent non-deterministic query results. This PR introduces a feature to use default ordering where available, and if none is available, default to id.

Prior bug:
https://github.com/shapiromatron/hawc/commit/3ec4783f870cadac282b7e82bdd892ee440f37b5#diff-12d8ff694bfca83829c6d1818f3b0008705913fa5eacb656e41fa9ddc0c5d36cL41


rabstejnek (approved):
Looks good

PR #329, opened 2020-10-30
Fix integration test warnings

rabstejnek:
`django.conf.urls.url` have been replaced with `django.urls.re_path` in Django 3


shapiromatron (approved):
I ended up changing `re_path` to `path` where it was simple, and then after doing that it seemed to unlock more deprecation warnings (achievement unlocked ?!?)

I think I fixed what I could , but could you please review and confirm it works for you?

rabstejnek (commented):
Looks good! The bulk of the remaining integration test warnings are from DRF, but it looks like its been fixed on their repo, so hopefully the next release will remove those.

PR #333, opened 2020-11-09
Hotfix: N/A scores on data pivot

rabstejnek:
`distinct` wasn't working since the studies weren't ordered by id, so there was a chance non-unique study ids were being sent to `get_dp_export`

This was causing a problem [here](https://github.com/shapiromatron/hawc/blob/62e253f33c530b012e1507d8ec0b578ec005323f/hawc/apps/riskofbias/models.py#L358), since the json dump would already exist in the mapping, so the default N/A value was being used.

### DIFF
--- a/hawc/apps/animal/exports.py
+++ b/hawc/apps/animal/exports.py
@@ -196,11 +196,7 @@ class EndpointGroupFlatDataPivot(FlatFileExporter):
         else:
             self.rob_headers, self.rob_data = RiskOfBias.get_dp_export(
                 self.queryset.first().assessment_id,
-                list(
-                    self.queryset.values_list(
-                        "animal_group__experiment__study_id", flat=True
-                    ).distinct()
-                ),
+                set(self.queryset.values_list("animal_group__experiment__study_id", flat=True)),
                 "animal",
             )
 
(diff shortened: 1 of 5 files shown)


hawc/apps/animal/exports.py
```diff
@@ -196,11 +196,7 @@ def _get_header_row(self):
         else:
             self.rob_headers, self.rob_data = RiskOfBias.get_dp_export(
                 self.queryset.first().assessment_id,
-                list(
```
shapiromatron, line 199: I tried to use the ORM on the db side to do the distinct select instead of passing the data over the wire, but I think it's less readable, what do you think?

```python
study_ids = list(
    self.queryset.values_list("animal_group__experiment__study_id", flat=True)
    .order_by("animal_group__experiment__study_id")
    .distinct("animal_group__experiment__study_id")
)
self.rob_headers, self.rob_data = RiskOfBias.get_dp_export(
    self.queryset.first().assessment_id, study_ids, "animal",
)
```

I think I prefer splitting out the study_ids, like this instead for this and other requests

```python
study_ids = set(self.queryset.values_list("animal_group__experiment__study_id", flat=True))
self.rob_headers, self.rob_data = RiskOfBias.get_dp_export(
        self.queryset.first().assessment_id, study_ids, "animal",
)
```

What do you think?

shapiromatron (commented):
Question on splitting up the study_id code from the next request, what do you think?

hawc/apps/animal/exports.py
rabstejnek: I agree on splitting out the study_ids. And as for readability, I think the former is still acceptable; it's pretty obvious what is going on with the method chain

shapiromatron: ok, feel free to make the change and i think this one is ready go go!

PR #339, opened 2020-11-10
server-side browser check

shapiromatron:
Handle IE11 browser check server side with template; this allows us to test and doesn't depend on javascript so can be handled for all browsers.


rabstejnek (approved):
Tested and works!

PR #342, opened 2020-11-16
Header footer

shapiromatron:
- add prime footer and cache
- add epa header/footer and cache


rabstejnek (changes requested):
I made a quick change that uses the same condensed version footer in both EPA and PRIME flavors instead of having a different, larger one in PRIME. I prefer the look of this slimmer one and it simplifies the code a little bit, but feel free to revert the commit if you disagree. Otherwise looks good!

shapiromatron:
@rabstejnek nice suggestion, I like it more too!

rabstejnek (approved):

PR #343, opened 2020-11-18
Update accordion

rabstejnek:
- Fixed accordion components in:
  - Reference search
  - Tag untagged references
  - Animal endpoint list
  - IV endpoint list
  - Outcomes list
  - Results list
- Removed proof of concept vocab view that utilized accordion.


shapiromatron (approved):
accordions looks good; I also refactored to remove the duplicate code and used crispy forms, should make this much more maintainable

PR #346, opened 2020-11-20
Update bootstrap in React input components

rabstejnek:
Update shared React components used as form inputs


rabstejnek:
Along with the shared input changes I also made several form specific changes.

An example is the exploratory heatmap "Figure customization" options, where I put the subgroups of form inputs into bootstrap cards.

image
image

shapiromatron (approved):

PR #347, opened 2020-11-20
Fix django-selectable render problems

rabstejnek:
Changed the order of script imports so that conflicting namespaces don't cause django-selectable buttons to not render. Also added custom scripting to render the selectable inputs using bootstrap stylings.
Reference: https://django-selectable.readthedocs.io/en/latest/advanced.html#using-with-twitter-bootstrap


shapiromatron (approved):
Looks great! I made two tweaks in my commit; the first added a little margin on the bottom, and the second make sure that the javascript was pulled from a URL instead of system path

PR #349, opened 2020-11-23
sanitize referrer

shapiromatron:
Make sure the referrer url can be resolved in the hawc namespace, otherwise return the default


rabstejnek (approved):
Looks good!

PR #361, opened 2020-12-02
bs4 - lit, summary, bmd

shapiromatron:
Primary work:
- migrate lit, summary, and bmd to bs4

Additional work:
- updated test fixture to include other items
- updated usernames and passwords in test fixture and added documentation for use

There was a little scope creep and I made minor revisions in shared code and other modules, but hopefully they are relatively minimal.


rabstejnek (changes requested):
I found just a few things, otherwise looks good!

- Exploratory heatmap tooltips are off; caused by sidebar/header? (seems offset by about that width/height)
- On executive summary (and elsewhere, if this happens elsewhere) hyperlinked headers are hidden by sticky navbar when clicked

PR #364, opened 2020-12-10
add new option to change pubmed and hero reference ids

shapiromatron:
All new options to the reference edit form to add or remove hero or pubmed ids. 

Screen Shot 2020-12-09 at 10 46 23 PM


rabstejnek (approved):
Looks good!

PR #370, opened 2021-01-05
remove all special characters with urlify

rabstejnek:
- No changes are made to `text_hyphen-123` (valid characters)
- The string `here is a space` converts to `here-is-a-space`
- The string `abc "🐕'~([{ }])+- 123` converts to `abc-123`


shapiromatron (approved):
Looks good! 

Since we don't have unit-tests on the JS side I updated the PR description to describe expected behavior.

I also found a straggling hand-rolled slugify, so I replaced with this method

PR #371, opened 2021-01-06
Animal metadata API endpoint

rabstejnek:
Using the API to create new bioassay endpoints requires knowledge of possible options for choice fields as well as foreign key relations for dose units, species, strain, etc.  This PR introduces a new API endpoint to query the hawc database and return all the metadata required to create endpoints.


tests/hawc/apps/animal/test_animal_api.py
```diff
@@ -618,3 +618,29 @@ def test_update_terms_permissions(self, db_keys):
...
+        # public should have access to this metadata
+        client = APIClient()
+        resp = client.get(url)
+        assert resp.status_code == 200
+
+    def test_metadata(self, rewrite_data_files: bool):
+        fn = "api-animal-metadata.json"
+        url = reverse(f"animal:api:metadata-list") + "?format=json"
+        client = APIClient()
+        assert client.login(username="admin@hawcproject.org", password="pw") is True
```
shapiromatron, line 636: no need to login, right?

hawc/apps/animal/actions/model_metadata.py
```diff
@@ -0,0 +1,89 @@
...
+    def dosing_regime_metadata(self):
+        metadata = {}
+        metadata["route_of_exposure"] = models.DosingRegime.ROUTE_EXPOSURE_CHOICES_DICT
+        metadata["positive_control"] = tuple_to_dict(models.DosingRegime.POSITIVE_CONTROL_CHOICES)
+        metadata["negative_control"] = tuple_to_dict(models.DosingRegime.NEGATIVE_CONTROL_CHOICES)
+        return metadata
+
+    def dose_group_metadata(self):
+        metadata = {}
+        dose_units = DoseUnits.objects.all()
```
shapiromatron, line 67: instead of a query and a serializer, I think we can just use [values](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#values)  with a queryset, for this and all the other serializer cases I think

```python
list(DoseUnits.objects.all().values('id', 'name'))
```

hawc/apps/animal/actions/model_metadata.py
```diff
@@ -0,0 +1,89 @@
...
+
+    input_model = NoInput
+
+    def experiment_metadata(self):
+        metadata = {}
+        metadata["type"] = tuple_to_dict(models.Experiment.EXPERIMENT_TYPE_CHOICES)
+        return metadata
+
+    def animal_group_metadata(self):
+        metadata = {}
```
shapiromatron, line 48: formatting opinions so there's no "correct", but could we use a `dict` and define in the consturctor, eg,. 

```python
def experiment_metadata(self):
    return dict(
        type=tuple_to_dict(models.Experiment.EXPERIMENT_TYPE_CHOICES),
    )
```

shapiromatron (changes requested):
Beautiful!

A few minor comments, can we also add to the hawc client as well?

shapiromatron (approved):
LGTM 

The test was failing on my computer however because there was no default ordering for the species, strain, dose-units.

I made a few other more changes that weren't exactly in scope

- added auto-retry for the selenium test
- updated bioassay jupyter notebook to demo metadata query usage
- added Study metdata since that's often needed too...

shapiromatron:
Looping in @DookTibs @jfmccoy-icf, now merged in!

PR #372, opened 2021-01-08
Generic summary tables

rabstejnek:
Created Django models and framework to implement summary tables, and more specifically a generic table.

## Report generation
Created a way to generate a word report for generic tables.
image

## HTML parsing
Quill text can be parsed into a word report.
image
image


hawc/apps/common/table.py
```diff
@@ -0,0 +1,60 @@
...
+        return parser.feed(self.html_text, par)
+
+
+class GenericTable(BaseModel):
+    rows: conint(gt=0)
+    columns: conint(gt=0)
+    cells: List[GenericTableCell]
+
+    @validator("cells")
+    def validate_cells(cls, cells, values):
```
shapiromatron, line 24: Add another to make sure one and only one cell for each position (may need to be in a global validator)

```python
d = set([(d.row, d.column) for d in cells])
if len(d) != tbl.rows * tbl.columns:
   raise ValueError(...)
```

hawc/apps/common/table.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def to_docx(self, par):
+        parser = CellHTMLParser()
+        return parser.feed(self.html_text, par)
+
+
+class GenericTable(BaseModel):
+    rows: conint(gt=0)
+    columns: conint(gt=0)
+    cells: List[GenericTableCell]
```
shapiromatron, line 21: we may also want to add `first_row_header: bool = True` and `second_row_header:bool = False`; we could also add `header_cells: List[GenericTableCell]` to make it more generic, but i'd say 99.99% of tables i see have one or two top header rows.

hawc/apps/common/table.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import List
+
+from docx import Document
+from pydantic import BaseModel, conint, validator
+
+
+class GenericTableCell(BaseModel):
+    row: conint(ge=0)
+    column: conint(ge=0)
+    html_text: str
```
shapiromatron, line 11: may as well add `col_span: int=0` and `row_span: int = 0` as thats something people will always want to do

hawc/apps/common/table.py
```diff
@@ -0,0 +1,60 @@
...
+                    f"Cell row outside of row bounds: row index {cell.row}, number of rows {values['rows']}"
+                )
+            if cell.column >= values["columns"]:
+                raise ValueError(
+                    f"Cell column outside of column bounds: column index {cell.column}, number of columns {values['columns']}"
+                )
+        return cells
+
+    def to_docx(self, docx: Document = None):
+        return Document()
```
shapiromatron, line 37: get rid of this return

hawc/apps/common/table.py
```diff
@@ -0,0 +1,60 @@
+from html.parser import HTMLParser
```
shapiromatron, line 1: should we make a new package for this? I dont think it's django app-specific, maybe `hawc.tools.tables`? I'd add an init and call this code `generic.py` perhaps?

hawc/apps/common/table.py
```diff
@@ -0,0 +1,60 @@
...
+        table = docx.add_table(rows=self.rows, cols=self.columns)
+        for cell in self.cells:
+            table_cell = table.cell(cell.row, cell.column)
+            cell_par = table_cell.add_paragraph()
+            cell.to_docx(cell_par)
+        return docx
+
+
+class CellHTMLParser(HTMLParser):
+    def feed(self, data, par):
```
shapiromatron, line 49: this parser should be in its own file in this new package, as it will be shared for html_text fields

hawc/apps/summary/admin.py
```diff
@@ -46,3 +47,8 @@ class SummaryTextAdmin(TreeAdmin):
         "created",
     )
     form = movenodeform_factory(models.SummaryText)
+
+
+@admin.register(models.SummaryTable)
+class SummaryTableAdmin(VersionAdmin):
```
shapiromatron, line 53: add some basic list attributes like title, created

hawc/apps/summary/models.py
```diff
@@ -149,6 +151,54 @@ def print_node(node, depth):
...
+
+    assessment = models.ForeignKey(Assessment, on_delete=models.CASCADE)
+    title = models.CharField(max_length=128)
+    slug = models.SlugField(
+        verbose_name="URL Name",
+        help_text="The URL (web address) used to describe this object "
+        "(no spaces or special-characters).",
+    )
+    content = models.JSONField(default=dict)
+    table_type = models.PositiveSmallIntegerField(choices=TABLE_CHOICES, default=DUMMY)
```
shapiromatron, line 175: maybe rename DUMMY our Generic table class

hawc/apps/summary/models.py
```diff
@@ -149,6 +151,54 @@ def print_node(node, depth):
...
+        return self.assessment
+
+    def to_docx(self):
+        table = GenericTable.parse_obj(self.content)
+        return table.to_docx()
+
+    def clean(self):
+        # content validation
+        try:
+            if self.table_type == self.DUMMY:
```
shapiromatron, line 189: pull this if statement out into a `get_content_schema_class` method which returns a pydantic class

hawc/apps/summary/models.py
```diff
@@ -149,6 +151,54 @@ def print_node(node, depth):
...
+    assessment = models.ForeignKey(Assessment, on_delete=models.CASCADE)
+    title = models.CharField(max_length=128)
+    slug = models.SlugField(
+        verbose_name="URL Name",
+        help_text="The URL (web address) used to describe this object "
+        "(no spaces or special-characters).",
+    )
+    content = models.JSONField(default=dict)
+    table_type = models.PositiveSmallIntegerField(choices=TABLE_CHOICES, default=DUMMY)
+    created = models.DateTimeField(auto_now_add=True)
```
shapiromatron, line 176: also needs a `published:bool = False` attribute like summary.Visual

hawc/apps/summary/api.py
```diff
@@ -115,3 +115,15 @@ def create(self, request, *args, **kwargs):
...
+
+
+class SummaryTableViewset(AssessmentViewset):
+    assessment_filter_args = "assessment"
+    model = models.SummaryTable
+    serializer_class = serializers.SummaryTableSerializer
+
+    @action(detail=True, methods=("get",), renderer_classes=(DocxRenderer,))
+    def docx(self, request, pk):
+        obj = self.get_object()
```
shapiromatron, line 127: create a new ReportExport class, like https://github.com/shapiromatron/hawc/blob/main/hawc/apps/common/helper.py#L239-L246, that the DocxRenderer can use directly; then you can remove the extra validation from the DocxRenderer class

hawc/apps/summary/api.py
```diff
@@ -115,3 +115,15 @@ def create(self, request, *args, **kwargs):
...
             raise exceptions.PermissionDenied()
         return super().create(request, *args, **kwargs)
+
+
+class SummaryTableViewset(AssessmentViewset):
+    assessment_filter_args = "assessment"
+    model = models.SummaryTable
+    serializer_class = serializers.SummaryTableSerializer
+
+    @action(detail=True, methods=("get",), renderer_classes=(DocxRenderer,))
```
shapiromatron, line 125: add `filter_backends = (InAssessmentFilter, UnpublishedFilter)` like the `VisualViewset`

hawc/apps/common/renderers.py
```diff
@@ -3,13 +3,48 @@
...
+class DocxRenderer(BaseRenderer):
+    """
+    Renders a Document object into a docx file.
+    A tuple of docx, filename can also be given.
+    """
+
+    media_type = "application/vnd.openxmlformats-officedocument.wordprocessingml.document"
+    format = ""
+
+    def validate_tuple(self, docx, filename):
```
shapiromatron, line 23: this validation can be removed after using the new `ReportExport` dataclass

shapiromatron (commented):
Looking great so far! Made some comments; I think it'd be great to move all the table base classes out of the django-hawc entirely so it will be easy to unit-test.

I added some complexity on the generic table too wrt header columns and colspan/rowspan; might be good to add early on since we'll always be doing stuff like that  but if you want to do it later I understand.

we'll also need to add footnotes somehow, still thinking that one through...

hawc/apps/common/table.py
shapiromatron: or maybe it's an optional `conint(ge=2)`?

shapiromatron: might also need validation check to ensure column + col_span < tbl.columns or something similar to that

hawc/tools/tables/generic.py
```diff
@@ -0,0 +1,80 @@
...
+        return parser.feed(self.html_text, block)
+
+
+class GenericTable(BaseModel):
+    rows: conint(gt=0)
+    columns: conint(gt=0)
+    cells: List[GenericTableCell]
+
+    @validator("cells")
+    def validate_cells(cls, cells, values):
```
shapiromatron, line 41: i noticed sometimes that this would run if i didnt specify columns, might add `skip_on_failure=True` to see if that fixes, and maybe a few unit-tests to show good/bad data examples?

hawc/apps/common/renderers.py
```diff
@@ -7,7 +7,27 @@
...
+    """
+    Renders a ReportExport object into a docx file.
+    """
+
+    media_type = "application/vnd.openxmlformats-officedocument.wordprocessingml.document"
+    format = ""
+
+    def render(self, data, accepted_media_type=None, renderer_context=None):
+        # throw error if we don't have a ReportExport
+        if not isinstance(data, ReportExport):
```
shapiromatron, line 23: i added a change upstream in #379 that allows us to short-circuit and return JSON if validation fails; consider bundling into this request (similar to how we handle the pandas renderer)

hawc/apps/summary/api.py
```diff
@@ -115,3 +115,16 @@ def create(self, request, *args, **kwargs):
...
         return super().create(request, *args, **kwargs)
+
+
+class SummaryTableViewset(AssessmentViewset):
+    assessment_filter_args = "assessment"
+    model = models.SummaryTable
+    filter_backends = (InAssessmentFilter, UnpublishedFilter)
+    serializer_class = serializers.SummaryTableSerializer
+
+    @action(detail=True, methods=("get",), renderer_classes=(DocxRenderer,))
```
shapiromatron, line 126: i think `get` is the default verb, so you can remove

hawc/apps/summary/models.py
```diff
@@ -149,6 +153,65 @@ def print_node(node, depth):
                 print_node(node, 2)
 
 
+class SummaryTable(models.Model):
+    objects = managers.SummaryTableManager()
+
+    GENERIC = 0
```
shapiromatron, line 159: wanna try using the django3 enum choices?

https://docs.djangoproject.com/en/3.0/ref/models/fields/#field-choices-enum-types

tests/hawc/tools/tables/__init__.py
```diff
@@ -0,0 +1,16 @@
+def document_compare(doc1, doc2):
```
shapiromatron, line 1: write a test to test this to make sure it can fail :)

shapiromatron (commented):
Looking really good. So good that I started working with it on #379; might be good to review that upstream and pick anything off that you want to put in this PR.   I'm fine collapsing these PRs into one if you think it'd be easier to get in, or keeping separate.

hawc/apps/summary/models.py
```diff
@@ -149,6 +153,65 @@ def print_node(node, depth):
...
+            return GenericTable
+
+    def get_table(self):
+        return self.get_content_schema_class().parse_obj(self.content)
+
+    def to_report(self):
+        table = self.get_table()
+        return ReportExport(docx=table.to_docx(), filename=self.slug)
+
+    def clean(self):
```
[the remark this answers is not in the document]
rabstejnek: Since the `clean` method just makes sure the content is parsable by our pydantic models, I added some parse tests to `test_generic.py` at your recommendation. As for this particular example I think its valid; I just take it as a 2x2 table with empty cells.

hawc/apps/common/helper.py
```diff
@@ -236,6 +237,16 @@ def clear_cache(cls, Model, filters):
         cls.delete_caches(Model, ids)
 
 
+@dataclass(frozen=True)
+class ReportExport:
+    """
+    Document export.
+    """
+
+    docx: Document
```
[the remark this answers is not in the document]
rabstejnek: The `Document` method is imported from `docx`, the `Document` type is imported from `docx.document`. I haven't had any type annotation issues yet, but if they come up going forward I will implement this change.

shapiromatron (approved):

PR #373, opened 2021-01-11
allow unaccented searching of author names

shapiromatron:
Allow search queries to to be entered without [diacritical](https://en.wikipedia.org/wiki/Diacritic) marks, eg a search for "Fred" should return Frédéric

Screen Shot 2021-01-13 at 9 44 18 AM

Enabled unaccent on the postgres db: https://docs.djangoproject.com/en/3.1/ref/contrib/postgres/operations/#unaccentextension


rabstejnek (approved):
Looks good!

PR #377, opened 2021-01-19
HERO url dependent on HERO access

rabstejnek:


hawc/apps/myuser/views.py
```diff
@@ -91,6 +91,10 @@ def get_context_data(self, **kwargs):
         )
         return context
 
+    def form_valid(self, form):
```
rabstejnek, line 94: Session is not easily accessible from a post_save signal, so I update the session variable here instead.

rabstejnek:
The only other place where the url is determined is [here](https://github.com/shapiromatron/hawc/blob/eed910c723bfd716b080386b5348caa347fbd659/frontend/lit/Reference.js#L9).
This is called within [DataPivotExtension](https://github.com/shapiromatron/hawc/blob/eed910c723bfd716b080386b5348caa347fbd659/frontend/summary/dataPivot/DataPivotExtension.js#L76) and [HeatmapDatastore](https://github.com/shapiromatron/hawc/blob/eed910c723bfd716b080386b5348caa347fbd659/frontend/summary/summary/heatmap/HeatmapDatastore.js#L394).
How would you suggest handling it? We can pass in the hero_access variable into these constructors but we'd be complicating these classes with an argument that has a very limited use case in the scope of these visuals.

shapiromatron:
revisit using `window.localStorage` to set flag on user login and profile update. then we can dynamically update the hero url with private if they have access

rabstejnek:
Closing in favor of #398, which uses the `window.localStorage` approach.

PR #378, opened 2021-01-21
Icf 167 create epi api

DookTibs:
draft PR as we discussed...I took a look at how animal API works and reworked a few things based on that, and had some other ideas to make it more streamlined. I still kept the whole "get or create" and the "single or list" ideas b/c I think those are pretty useful.

I am now hooking into forms for validation in e.g. OutcomeWriteSerializer; I thought on the call you had said you had broken out form validation into a separate class and the API and form both talked to that, but it lookked like in the animal api it just called the form.is_valid directly?

The custom_perm_checker/getOrCreatePermissionCheckHelper stuff I mentioned in the email the other day is also gone now; I found a better way to do that I think, passing in the user to the serializer on construction and then the serializer's (or the mixin's) validate method can use that for permission checking. Let me know what you think of that. That certainly made things less complicated than in the email I sent the other day.

So to restate those outstanding questions:
1. Does this general “get/create” approach make sense/seem useful?

2. Is this method of checking permissions (passing in user to the serializer and doing it during validate) acceptable?
 
3. When I actually check the permissions, I’m using user_can_edit_object. But with this get/create approach sometimes you aren’t actually editing, you’re just retrieving, and in those times maybe user_can_view_object is the strictly right one to call? I think it makes more sense to just always call user_can_edit_object but if you feel strongly the other way I can look at a way to do it cleaner.


shapiromatron:
thanks @DookTibs! I'll spend some time digging in... also included @rabstejnek for a review too...

hawc/apps/common/api/mixins.py
```diff
@@ -108,3 +110,113 @@ class LegacyAssessmentAdapterMixin:
...
+    just a little more straightforward -- rather than having to
+    lookup/check/create-if-needed, clients can just hit one endpoint and
+    get back the same object id every time.
+
+    Basic approach taken from https://stackoverflow.com/questions/25026034/django-rest-framework-modelserializer-get-or-create-functionality
+   """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        user = kwargs.get("context", {}).get("user")
```
rabstejnek, line 193: Since viewsets add `request` to the context by default you can access the user that way (`request.user`) instead of passing user in explicitly. That would simplify the `get_serializer` calls a little bit.

hawc/apps/common/api/mixins.py
```diff
@@ -108,3 +110,113 @@ class LegacyAssessmentAdapterMixin:
...
+            if isinstance(validator, UniqueTogetherValidator):
+                self.validators.remove(validator)
+        super().run_validators(value)
+
+    # provides default validation behavior; ensures that the
+    # associated assessment with an object can be edited by the
+    # user in question. If there is no associated assessment this
+    # will fail an assertion and the enclosing serializer must
+    # instead implement its own validation routine and not call this one.
+    def validate(self, data):
```
rabstejnek, line 207: Maybe `data.get("assessment")` can be checked at the viewset permissions level on `POST` requests, that way we're not checking permissions both here and there. @shapiromatron what do you think?

rabstejnek (commented):
Looks good!

shapiromatron (commented):
Looks good overall, full steam ahead. I'd recommend reviewing a few patterns for the animal API to mirror some of the patterns there.

If it's a simple create with a single form (study pop, outcome), you can follow the animal.Experiment patterns:
- https://github.com/shapiromatron/hawc/blob/main/hawc/apps/animal/api.py#L178-L181
- https://github.com/shapiromatron/hawc/blob/main/hawc/apps/animal/serializers.py#L30-L50
- use some of the helper functions we created eg., get_matching_instance

If it's a form + formset (eg., result, comparison set), take a look at the animal.Endpoint pattern:
- https://github.com/shapiromatron/hawc/blob/main/hawc/apps/animal/serializers.py#L334-L400
- https://github.com/shapiromatron/hawc/blob/eed910c723bfd716b080386b5348caa347fbd659/hawc/apps/animal/forms.py#L532-L583 (though if we can just pass the serializer to the form as you did, that's even better)
- https://github.com/shapiromatron/hawc/blob/main/hawc/apps/animal/serializers.py#L234-L245

However, one pattern that you suggest that I'd rather not use is the POST that can create both a single or multiple instance of the object.  We dont use that pattern elsewhere and it's not expected REST behavior. If you want, i'd make a single creation with a POST to /{resource}/, and a multiple creation to a POST /{resource}/multiple/, it'd be easy to just make a second view I think, with less conditional checking and just expects a single instance an a list of instances, respectively.

DookTibs:
@shapiromatron just pushed a bunch more changes, here's the current status and notes:

* API is now functional - all epi stuff should now be creatable/updateable/deletable.

* Includes a publicly visible metadata endpoint (/epi/api/metadata/). I also added an assessment-specific version if you append an assessment id, e.g. /epi/api/metadata/508/; this requires view permission on the assessment in question and includes assessment-specific values like adjustment factors and criteria that are omitted from the public one.

* I have added pytests for all this

* I have NOT yet added support for any of this to the HawcClient

* I have NOT yet deployed this to our AWS dev instance; planning to do this over the next day or so. So I've done a lot of testing but we haven't actually started using this "for real".

* as mentioned on the last call, pretty much all the fields that are tied to choices tuples (ex: StudyPopulation.DESIGN_CHOICES) will accept either the value ("CO") or a case-insensitive readable version ("CoHoRt"). This also works for fields that are tied to other db tables; for instance StudyPopulation.countries can be specified by either the id of the row in the db table, or by the name like "Germany". This feels like a nice feature to me and there is interest in making use of this on our side, but if this feels awful to you I can always rip it out.

* for fields that are tied to another list but allow you to add to said list (for example - when adding adjustment factors to a result); these also will accept either id's or display values. If a display value the api will look up the underlying object; if it fails, it will create them on the fly. The idea here is to make it pretty analogous to the web UI which will let you add new ones, but I want to call this out b/c again I could see this feeling awful, so I can rip it out if you hate it.

* similarly, dtxsid's can be specified that are not yet in HAWC; they will be imported if possible.

* one thing I shelved at the last instance is the FormIntegrationMixin but you'll see remnants of this in the code. Basically serializers could mix this in and during a validate call I would construct a form (so like OutcomeSerializer would instantiate OutcomeForm) and call is_valid on it. I ran into a bunch of issues with this breaking some existing tests, and worse it would break during PATCH calls to the API. For example if "metric" is required for a result, and I PATCHed an existing result with just a "name" update, then the form.is_valid call would complain that metric wasn't supplied. I could probably code around this but didn't want to hold up the rest of things while I worked on that.

Anyway take a look and let me know if there are changes you'd like or if not I can submit it as a real PR. Thx!

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+                            )
+                            criteria_serializer.is_valid(raise_exception=True)
+                            criteria = criteria_serializer.save()
+                            fixed.append(criteria.id)
+                    else:
+                        fixed.append(el)
+
+                request.data[data_key] = fixed
+
+    def perform_create(self, serializer):
```
shapiromatron, line 206: wrap this and all others in a transaction in case we get in a bad state and need to revert https://docs.djangoproject.com/en/3.1/topics/db/transactions/#django.db.transaction.atomic

Please do throughout the api.py

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+        super().perform_create(serializer)
+        self.process_criteria_association(serializer, serializer.data["id"], True)
+
+    def perform_update(self, serializer):
```
shapiromatron, line 210: same re: transactions

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+
+        # ...and save any new ones
+        if len(inserts) > 0:
+            serializer = serializers.SimpleStudyPopulationCriteriaSerializer(
+                data=inserts, many=True
+            )
+            serializer.is_valid(raise_exception=True)
+            serializer.save()
+
+    def get_queryset(self, *args, **kwargs):
```
shapiromatron, line 154: move this to top since it's "standard"

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+    criteria_categories = (
+        ("inclusion_criteria", "I",),
+        ("exclusion_criteria", "E",),
+        ("confounding_criteria", "C",),
+    )
+
+    def process_criteria_association(self, serializer, study_population_id, post_initial_create):
+        # this should probably be rewritten to use the ManyToManyManager on the underlying instance...
+        inserts = []
+        for cc in self.criteria_categories:
```
shapiromatron, line 125: Use tuple unpacking:
```
for (data_key, type_code) in self.criteria_cateogories:
```

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+            data_key = cc[0]
+            type_code = cc[1]
+
+            if data_key in self.request.data:
```
shapiromatron, line 129: do in two passes? first find what we need to delete and what we need to insert? do all deleets and all inserts at once?

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+            serializer = serializers.SimpleStudyPopulationCriteriaSerializer(
+                data=inserts, many=True
+            )
+            serializer.is_valid(raise_exception=True)
+            serializer.save()
+
+    def get_queryset(self, *args, **kwargs):
+        return self.model.objects.all()
+
+    def handle_criteria(self, request, during_create):
```
shapiromatron, line 157: i think the serializer should have a study instance you can just use. `serializer.instance.study`; no need for more db calls

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
 
-class StudyPopulation(AssessmentViewset):
+class Criteria(PermCheckerMixin, AssessmentEditViewset):
+    perm_checker_key = "assessment"
+    assessment_filter_args = "assessment"
+    model = models.Criteria
+    serializer_class = serializers.CriteriaSerializer
+
+
+class StudyPopulation(PermCheckerMixin, viewsets.ModelViewSet):
```
shapiromatron, line 110: why can't we use an AssessmentViewset or AssessmentEditViewset - this adds all the permission checking built-in. 

Same with other Viewsets - I'd rather not use generic DRF stuff again.

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        self.perform_create(serializer)
+
+        instance = self.model.objects.get(id=serializer.data["id"])
+        serializer = self.get_serializer(instance)
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+class Exposure(ReadWriteSerializerMixin, PermCheckerMixin, AssessmentEditViewset):
```
shapiromatron, line 234: do we need PermChecker mixin with the AssessmentEditViewset?

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+
+            if len(cts) == 0:
+                raise ValidationError(f"At least one central tendency is required")
+
+            # allow clients to specify either keys like 2 or readable values like "median" when accessing the API
+            for ct in cts:
+                if "estimate_type" in ct:
+                    probe_ct_estimate_type = ct["estimate_type"]
+                    if type(probe_ct_estimate_type) is str:
+                        converted_estimate_type = find_matching_list_element_by_value(
```
shapiromatron, line 261: it seems like we're doing it wrong; this seems too low low level

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+            ct_serializer = serializers.CentralTendencyPreviewSerializer(data=cts, many=True)
+            try:
+                ct_serializer.is_valid(raise_exception=True)
+            except ValidationError as ve:
+                raise ValidationError({"central_tendencies": ve.detail})
+        else:
+            if not during_update:
+                raise ValidationError(f"At least one central tendency is required")
+
+    def handle_dtxsid(self, request):
```
shapiromatron, line 297: you are nicer than me :) nice job

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+        # client can supply an id, or the name of the doseunits entry (and then we'll look it up for them - or create it if needed)
+        if "metric_units" in request.data:
+            metric_units_probe = request.data["metric_units"]
+
+            if type(metric_units_probe) is str:
+                try:
+                    metric_units = DoseUnits.objects.get(name=metric_units_probe)
+                    request.data["metric_units"] = metric_units.id
+                except ObjectDoesNotExist:
+                    # option 1 - require a pre-existing dose unit
```
shapiromatron, line 324: use option 1

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+        if "central_tendencies" in self.request.data:
+            cts = self.request.data["central_tendencies"]
+            for ct in cts:
+                ct["exposure"] = exposure_id
+
+            ct_serializer = serializers.CentralTendencyWriteSerializer(data=cts, many=True)
+            ct_serializer.is_valid(raise_exception=True)
+            ct_serializer.save()
+
+    def perform_create(self, serializer):
```
shapiromatron, line 368: tranactions

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
 
 
-class Result(AssessmentViewset):
+class GroupResult(PermCheckerMixin, AssessmentEditViewset):
+    perm_checker_key = "group"
+    assessment_filter_args = "result__outcome__assessment"
+    model = models.GroupResult
+    serializer_class = serializers.GroupResultSerializer
+
+    # note - this does not currently validate that a supplied group is part of the assessment.
```
shapiromatron, line 390: i think it should, yes?

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+    perm_checker_key = ["outcome", "comparison_set"]
     assessment_filter_args = "outcome__assessment"
     model = models.Result
     serializer_class = serializers.ResultSerializer
 
+    factor_categories = (("factors_applied", True,), ("factors_considered", False,))
+
+    def process_adjustment_factor_creation(self, serializer, result_id, post_initial_create):
+        inserts = []
+        for fc in self.factor_categories:
```
shapiromatron, line 403: tuple unpacking as above

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+                for adjustment_factor_id in adjustment_factor_ids:
+                    dynamic_obj = {
+                        "included_in_final_model": is_included,
+                        "adjustment_factor": adjustment_factor_id,
+                        "result": result_id,
+                    }
+                    inserts.append(dynamic_obj)
+
+        # ...and save any new ones
+        if len(inserts) > 0:
```
shapiromatron, line 425: same thought as above, do bulk deletes/creates after inspecting data

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+    def handle_adjustment_factors(self, request, during_create):
+        # first - what assessment are we working in?
+        assessment_id = None
+        if during_create:
+            if "outcome" in request.data:
+                outcome_id = request.data["outcome"]
+
+                try:
+                    outcome = models.Outcome.objects.get(id=outcome_id)
+                    assessment_id = outcome.get_assessment().id
```
shapiromatron, line 439: just use outcome.assessment_id

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+        # default behavior except we need to refresh the serializer to get the adjustment factors to show up in the return...
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        self.perform_create(serializer)
+
+        instance = self.model.objects.get(id=serializer.data["id"])
+        serializer = self.get_serializer(instance)
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+    def perform_create(self, serializer):
```
shapiromatron, line 495: transactions

hawc/apps/epi/api.py
```diff
@@ -87,37 +100,425 @@ def result_heatmap(self, request, pk):
...
+        self.process_adjustment_factor_creation(serializer, self.get_object().id, False)
+
+
+class ComparisonSet(PermCheckerMixin, AssessmentEditViewset):
+    perm_checker_key = "study_population"
     assessment_filter_args = "assessment"  # todo: fix
     model = models.ComparisonSet
     serializer_class = serializers.ComparisonSetSerializer
 
+    # note - this does not currently validate that a supplied exposure is a valid one...
```
shapiromatron, line 510: please add

hawc/apps/epi/api.py
```diff
@@ -139,3 +540,21 @@ class ExposureCleanup(CleanupFieldsBaseViewSet):
     serializer_class = serializers.ExposureCleanupFieldsSerializer
     model = models.Exposure
     assessment_filter_args = "study_population__study__assessment"
+
+
+class Metadata(viewsets.ViewSet):
+    def list(self, request):
+        return EpiMetadata.handle_request(request)
+
+    def retrieve(self, request, *args, **kwargs):
```
shapiromatron, line 549: make this dumber; require an assessment. use mixins.RetrieveModelMixin, and just check that assessment.user_can_view_object(request.user). No list view.

shapiromatron (commented):
Lots to discuss here. I can hop on a call if it'd be beneficial.

PR #379, opened 2021-01-21
Evidence Profile Table model and report generation

shapiromatron:
1. Added evidence profile tables to our summary tables.
2. Add frontend CRUD views

## Report generation
A report can be generated with given content.
image


hawc/apps/summary/serializers.py
```diff
@@ -132,5 +132,16 @@ class Meta:
...
 
+    def to_representation(self, instance):
+        ret = super().to_representation(instance)
+        if instance.id:
+            ret["url"] = instance.get_absolute_url()
+        return ret
+
+    def validate(self, data):
+        # check model level validation
+        models.SummaryTable(**data).clean()
```
rabstejnek, line 143: This allows us to use the same validation on both the API/serializers and on forms, all defined in one place (`clean`).
This is necessary since we want validation both on the editing pages where we bypass django forms and use the API, and on admin pages where the django forms are used and `clean` is called.

PR #380, opened 2021-02-04
Generic table editing user-interface

shapiromatron:
Generic tables do not have any schema, but instead allow users to create tables similar to what is done in Excel or Word. Users specify the numbers of rows and columns to present, and can merge rows/columns as needed. They can then edit content in the cell, we're currently using free-text with the Quill library.

**Editing user interface**
user-interface

**Column size adjustments**
column-adjustment

**Full-cell editing with detailed options**
full-cell-editing

**"quick(er)" Inline cell editing**
inline-cell-editing


rabstejnek (commented):
Looks good! My only concern is with header cells; since you put all of the header cells into `thead`, there is no way to define a row header or anything of the like partway down the table. Also the separation of `thead` and `tbody` introduces rendering problems when you try to row span a header cell into the `tbody` (example below). Can we just use `th` and `td` and bypass the separation of `thead` and `tbody` altogether?
> image
Header cell A1 has row span set to 2. This removes the rendering of cell A2 but does not span appropriately.

rabstejnek (approved):

shapiromatron:
> Looks good! My only concern is with header cells; since you put all of the header cells into `thead`, there is no way to define a row header or anything of the like partway down the table. Also the separation of `thead` and `tbody` introduces rendering problems when you try to row span a header cell into the `tbody` (example below). Can we just use `th` and `td` and bypass the separation of `thead` and `tbody` altogether?
> 
> > image
> > Header cell A1 has row span set to 2. This removes the rendering of cell A2 but does not span appropriately.

This makes sense, good thinking Danny. I'm going to keep as is for now, though if users are finding it's not meeting their needs we can re-evaluate.

The reason is this "generic table" isn't really as generic as the html spec; it's designed to create a table with header columns column level headings, or subheadings in a table.   This is probably 95% of all tables in systematic reviews and the primary driver for our use case here. Generally, people may want to create subheadings, which our generic table would allow by just modifying the formatting, but not new headings. The EPT table is our exception to this rule, but that's why we have a separate table format for it...

PR #382, opened 2021-02-16
Evidence Integration Table

rabstejnek:
Created an evidence integration table model and report generation. Also implemented column width settings on generic table reports.

## Evidence integration table

image

An empty cell on the leftmost side of the table was left where an arrow graphic was in the reference material. This may change as we receive more feedback on how to proceed with this table.

## Column widths

image

This is a generic table with column widths set. Report generation now reflects these changes:

image


rabstejnek:
Moved column width and efficiency changes to #390

Closing since this table type is no longer under consideration.

PR #383, opened 2021-02-16
Allow null boolean fields to be blank for form validation

rabstejnek:
By setting `blank=True` as an argument on the `BooleanField`s, form validation allows a `null` value.

image

All the other instances of `BooleanField` where the `null` argument was set were checked using the regex:
```
BooleanField[(](\n|[^)])*null(\n|[^)])*[)]
```
Changes were made appropriately if it was missing the `blank=True` argument.

image


hawc/apps/animal/migrations/0031_auto_20210216_1238.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.1.3 on 2021-02-16 17:38
```
shapiromatron, line 1: instead of a new migration, can you just add `blank=True,` to the django31 migration. Since this doesn't affect the db I try to remove these  IF done correctly, running makemigrations should be a noop.

shapiromatron (commented):
One minor change to try to remove the migrations. Can you also paste that regex in the screenshot into text, just in case we need it.

shapiromatron (approved):
Looks great, I did  a grep to compare your regex:

```bash
grep 'models.BooleanField(' --include models.py -r -A 10 hawc/apps --color > bools.txt
```

Your regex was far cleaner to read, but the good news was in both cases, you got everything!

I also added a required field annotation.

PR #385, opened 2021-02-18
improve container deployment and docs

shapiromatron:
- vendor phantomjs
- redirect all logs to stdout so we can view on docker-compose
- remove supervisord; split django container into web/worker/cron and a one-time sync
- ovehaul deployment instructions


rabstejnek (approved):
Looks good, got it to work!

One remark though, which may not need any action: I had to remove windows style newlines from shell scripts (`wait-for.sh`, `web.sh`, `workers.sh`, `cron.sh`). I'm not sure if this is an issue with the repository files or if git is automatically adding windows style newline for me (`autocrlf` or something of the like). This was the only real problem I had and may be entirely local on my part.

PR #387, opened 2021-02-22
hotfix- default to default score instead of random score in dp export

shapiromatron:
Currently data pivot exports will select any matching rob score instead of the default score.  

Future updates will select the outcome-specific score (which may be a animal-group or endpoint override), but in the meantime, the default score will be shown


rabstejnek (approved):

tests/hawc/apps/riskofbias/test_riskofbias_models.py
```diff
@@ -31,3 +32,34 @@ def test_get_orphan_relations(self):
...
+            ).values_list("metric_id")
+        ) == {metric for _, metric in scores_map.keys()}
+
+        # however, study has more scores than metrics
+        assert RiskOfBiasScore.objects.filter(
+            riskofbias__study=study_id, riskofbias__final=True, riskofbias__active=True,
+        ).count() > len(scores_map.keys())
+
+        # make sure only default scores are used
+        for (_, metric), score in scores_map.items():
```
shapiromatron, line 57: use .values() instead of items() since you're throwing away the key

tests/hawc/apps/riskofbias/test_riskofbias_models.py
```diff
@@ -31,3 +32,34 @@ def test_get_orphan_relations(self):
...
+            expected_score = RiskOfBiasScore.objects.get(
```
shapiromatron, line 58: i'm not sure how many queries this might be, but it could potentially be a lot. is there a single query that can get all the data and then you  could build a metric key:value dict to access? might speed it up,

shapiromatron (commented):
A few minor tweaks to the tests, but pretty close!

tests/hawc/apps/riskofbias/test_riskofbias_models.py
rabstejnek: The key in `scores_map` is a tuple, so I'm only throwing away the first element from the tuple; I still need the second element, `metric`

shapiromatron: ahh, makes sense!

tests/hawc/apps/riskofbias/test_riskofbias_models.py
```diff
@@ -31,3 +32,29 @@ def test_get_orphan_relations(self):
...
+    def test_get_dp_export(self):
+        assessment_id = 1
+        study_id = 1
+        _, scores_map = RiskOfBias.get_dp_export(assessment_id, [study_id], "animal")
+        scores_qs = RiskOfBiasScore.objects.filter(
+            riskofbias__study=study_id, riskofbias__final=True, riskofbias__active=True,
+        )
+
+        # all of the metrics are mapped
+        set(scores_qs.values_list("metric_id")) == {metric for _, metric in scores_map.keys()}
```
shapiromatron: assert?

tests/hawc/apps/riskofbias/test_riskofbias_models.py
```diff
@@ -45,7 +45,7 @@ def test_get_dp_export(self):
         )
 
         # all of the metrics are mapped
-        assert set(scores_qs.values_list("metric_id")) == {
+        assert set(scores_qs.values_list("metric_id", flat=True)) == {
```
shapiromatron, line 48: haha nice! the assert was needed!!!

shapiromatron (commented):
good thing we added that assert :)

PR #392, opened 2021-02-25
delete reference guardrails when linked to study

shapiromatron:
1. display associated study on reference detail pages if one exists
Screen Shot 2021-02-25 at 9 28 42 AM

2. only allow reference delete if study does not exist; if study does exist, show link to the study
Screen Shot 2021-02-25 at 9 29 43 AM

### DIFF
--- a/frontend/lit/Reference.js
+++ b/frontend/lit/Reference.js
@@ -22,6 +22,13 @@ class Reference {
         return `/lit/reference/${this.data.pk}/edit/`;
     }
 
+    get_study_url() {
+        if (this.data.has_study) {
+            return `/study/${this.data.pk}/`;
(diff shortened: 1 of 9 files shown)


rabstejnek (changes requested):
Looks good, but I think the wording remains unclear.

"Please delete the study first, then this reference can be deleted". Deleting the study _does_ delete the reference, so maybe "Please delete the study instead."

Also maybe the study delete can be updated to mention that the underlying reference will be removed too:
image

shapiromatron:
@rabstejnek see the description above; do the proposed changes suffice?

rabstejnek (approved):
Looks good!

PR #398, opened 2021-03-09
public/private HERO reference  links

rabstejnek:
We've long a had a user profile feature which was designed to show the public or private HERO home page for references, depending on a user profile setting. We've now updated the frontend so that if a user selects authenticated hero access, then they'll get links that are on the private hero home page instead of the public. 

By default, public are shown.

The setting is saved in a user's local storage so we dont' have to query the database; all APIs serve the public URL.

Screen Shot 2021-03-12 at 12 15 31 PM

Example:
- public (default): https://hero.epa.gov/hero/index.cfm/reference/details/reference_id/6323927
- private: https://heronet.epa.gov/heronet/index.cfm/reference/details/reference_id/6323927


shapiromatron (commented):
Moved things around a little bit @rabstejnek what do you think?

I also made a common template include to update the hero access setting that's shared on the login portal home page and the user profile detail page. I figure that's better than a custom one for each. The success url for a profile update is the detail page, so it should always be triggered there.

shapiromatron (approved):

rabstejnek:
@shapiromatron Changes look good!

PR #399, opened 2021-03-10
Google Analytics

rabstejnek:
Update analytics so that the key is configurable and can be specified via an environment variable.


shapiromatron (approved):
If this works, then I approve. 

Could you add a fake key in `hawc/compose/example.env` too?

rabstejnek:
image
Implemented the Google Tag Manager script and tested it, and it works!

PR #400, opened 2021-03-10
Optimize endpoint list / animal group detail

rabstejnek:
Before:
image
image

After:
image
image


rabstejnek:
@shapiromatron This is the timing when caching is bypassed and `get_qs_json` directly calls the serializer with the queryset.

image

The actual serializing part is still taking the bulk of the operation time, so a postgres view based approach like we discussed or simplifying the serializer we use here to just grab the necessary data to build the table will likely improve efficiency.

rabstejnek:
Optimized the queryset further:
image
image

The animal group detail page is also included in this PR, since it also uses endpoint's `to_json` and benefits from the prefetching.

Before:
image
image

After:
image
image

shapiromatron (approved):

PR #403, opened 2021-03-12
EPT revisions

shapiromatron:
**Updates based on reviewer feedback:**

- [x] for detailed text, change the ?  icon to an info icon
- [x] schema change:  Add new `hide_content` to schema, default False. This includes human, animal, mechanistic, and integration tab
- [x] schema change: Add new `no_content_text`, which defaults to `No data available`. This includes human, animal, and mechanistic
- [x] Render no_content_text in word
- [x] Render no_content_text in html
- [x] Change word layout if hide_content is true
- [x] Change html layout if hide_content is true

**Screenshots:**
Screen Shot 2021-03-12 at 4 32 37 PM
Demo of the new hover-over icons

Screen Shot 2021-03-23 at 5 55 14 PM
This shows hiding the animal and integration sections, as well as customizing the no data available text.


rabstejnek:
@shapiromatron All of the tasks are implemented, so its ready for a review!

There is one 500 error being thrown when all the content is hidden, since the pydantic model expects at least one cell, so I'll probably just remove that validation from the model so that we don't have any similar problems going forward with these custom tables.

hawc/apps/summary/models.py
```diff
@@ -249,6 +249,8 @@ def clean(self):
             self.get_table()
         except PydanticError as e:
             raise ValidationError({"content": e.json()})
+        except Exception as e:
```
shapiromatron, line 252: Could you change to the specific exception type instead of a catch all?

rabstejnek: Done!

shapiromatron:
@rabstejnek  Force-merging - I cannot approve since I started this work, but I made a few changes after you completed your section.

The only coding thing I changed was altering how you were showing/hiding the fields, if the data was not available. Instead of updating css styles, I just change the class to be "hidden" if we're not showing.

PR #404, opened 2021-03-12
Filterable heatmaps

shapiromatron:
Exploratory heatmaps need a few more features to make them usable in production. Most notably, a few additional fields should be presented in the heatmap datasets, and the frontend needs the option to present only a subset of all available rows. For now, implement the same filter approach as is used in Data Pivot and Crossview, so they can be updated together if needed.

**Backend changes:**

- [x] add overall risk of bias to epi
- [x] add overall risk of bias score to bioassay 
- [x] add calculated fields to bioassay `Treatment Period`, `Animal Description`, and `Animal Description N`
- [x] add to bioassay: https://github.com/shapiromatron/hawc/blob/main/hawc/apps/animal/exports.py#L32-L40
- [x] add calculated animal-group to bioassay: https://github.com/shapiromatron/hawc/blob/main/hawc/apps/animal/exports.py#L11-L30
- [x] add new fields to dashboard views 

**Frontend changes:**

- [x] add option to filter the tabular dataset to only show a subset of the data
- [x] move data filters to new tab pane
- [x] add new fields to dashboard views 

**Filtering options:**
Screen Shot 2021-03-22 at 8 58 12 PM

**Example plot with new calculated fields in table and filter-widgets:**
Picture1


rabstejnek (commented):
Looks good, but I did come across one error when testing a literature summary heatmap. When filtering tags, it threw an error because it tried to call `toString` on `null` in `filters.js`. I made a change locally that seemed to fix it, but I don't know why there was a null value to begin with or if its worth investigating further.
The change I made, may need to be made elsewhere in filters to account for null values:
image

rabstejnek (approved):

PR #409, opened 2021-03-24
RoB score materialized view

rabstejnek:
Created a framework for adding materialized views, and added our first materialized view: `FinalRiskOfBiasScore`

image

image


hawc/apps/materialized/models.py
```diff
@@ -0,0 +1,59 @@
...
+        cache.get(f"refresh-{cls._meta.db_table}", False)
+
+    @classmethod
+    @transaction.atomic
+    def refresh_view(cls):
+        with connection.cursor() as cursor:
+            cursor.execute(f"REFRESH MATERIALIZED VIEW CONCURRENTLY {cls._meta.db_table}")
+
+
+class Score(MaterializedView):
```
rabstejnek, line 28: `MaterializedScore`? Or is the `materialized` namespace enough? I'm curious what kind of naming convention we should use for these going forward.

rabstejnek:
I recently changed the SQL that creates this materialized view because it didn't preserve "original" scores when we joined overrides onto the scores. I have tested these new changes and they appear to work correctly.

shapiromatron (changes requested):

hawc/apps/assessment/views.py
```diff
@@ -365,6 +366,8 @@ def get(self, request, *args, **kwargs):
             raise PermissionDenied()
 
         assessment.bust_cache()
+        refresh_views_by_cache()
```
shapiromatron, line 369: outside the materialized model, i'd be confused by the meaning of views.  Maybe rename to `update_all_mvs()`.

Also, this would only work if our refresh flag is set I think right? It'd be better if this method would just refresh all mvs, no matter what.

hawc/apps/materialized/admin.py
```diff
@@ -0,0 +1,49 @@
+from django.contrib import admin
+
+from ..common.admin import ReadOnlyAdmin
+from . import models
+
+
+class IsOverrideFilter(admin.SimpleListFilter):
```
shapiromatron, line 7: can this a more generic class, maybe moved to common, call `IsNullFilter` Or maybe a generic one with something like this:

```python
class IsOverrideFilter(IsNullFilter):
    qs_filter = "object_id"
```

hawc/apps/materialized/managers.py
```diff
@@ -0,0 +1,20 @@
+from django.db import models
+
+
+class FinalRiskOfBiasScoreQuerySet(models.QuerySet):
+    def prioritized_scores(self):
```
shapiromatron, line 5: i dont think this will work in the way you intend.  The reason is for a given set of RiskOfBiasScore, we'll likely want 10 defaults, 1 animal group override, and 2 endpoint overrides, one score for each metric id. Suggest removing for now.

hawc/apps/materialized/models.py
```diff
@@ -0,0 +1,63 @@
...
+    And since these are views, any foreign keys on these models are not "true" foreign keys, so on_delete
+    should be kept as DO_NOTHING to prevent Django from enforcing foreign key constraints.
+    """
+
+    class Meta:
+        abstract = True
+        managed = False
+
+    @classmethod
+    def set_refresh_view_cache(cls, refresh: bool):
```
shapiromatron, line 22: again, while correct, view by itself is confusing a bit. consider renaming to

- set_refresh_flag
- get_refresh_flag
- refresh_if_flagged

shapiromatron (changes requested):
shaping up, a few more minor revisions

hawc/apps/assessment/views.py
rabstejnek: I believe that calling a celery task as a function does it synchronously, based on [this stackoverflow answer](https://stackoverflow.com/a/55537707) and the following from the [celery docs](https://docs.celeryproject.org/en/stable/userguide/calling.html):
```
calling (__call__)
    Applying an object supporting the calling API (e.g., add(2, 2)) means that
    the task will not be executed by a worker, but in the current process instead
    (a message won’t be sent).
```
If you want to keep synchronous functions separate from our celery tasks though I don't mind just making a separate function that the task calls instead.

shapiromatron: Thanks, this is good to know. 

The way I generally view it, celery tasks are designed more just to implement how we run something, but shouldn't be the only place that contain the business logic. This allows the business logic to be tested and checked or used in other ways w/o the celery infrastructure.  We wont easily be able to write tests for this specific task, but thats my general mindset as to why code is in different places.

Given that, I'd prefer a separate function that the task and other things could call.  Would you agree?

rabstejnek: Yep, I agree with that approach.

hawc/apps/materialized/admin.py
rabstejnek: I made a generic `YesNoFilter` that you can set the inclusion criteria on as a Q object. You kind of have to go out of your way to make a Q object when you inherit from this class, but I think its more flexible and can be used more widely on our other admin pages this way.

hawc/apps/materialized/managers.py
rabstejnek: I went ahead and removed it, but I left the boilerplate in place for custom manager functions / queryset functions down the line

hawc/apps/materialized/admin.py
shapiromatron: that sounds great, using Q object is a great idea b/c that adds a lot of flexibility

hawc/apps/common/admin.py
```diff
@@ -5,3 +5,47 @@ class AllListFieldAdmin(admin.ModelAdmin):
...
+        return False
+
+    def has_change_permission(self, request, obj=None):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return False
+
+
+class YesNoFilter(admin.SimpleListFilter):
```
shapiromatron, line 21: love it!

shapiromatron (approved):

shapiromatron:
@rabstejnek looked great; I made the following changes in my push to this branch:

- added `manage.py refresh_views` command to refresh views
- added documentation to describe management command for refreshing data
- when flushing assessment cache, also trigger materialized view refresh
- add new daily celery task to force refresh all materialized views
- (unrelated) change location of "view all assessments" for staff

PR #222, opened 2020-05-18
Animal bioassay create API

rabstejnek:
- [x] Experiment
- [x] Animal Group / Dosing Regime
- [x] Endpoint

PR #411, opened 2021-03-25
create/update epi API (ICF)

DookTibs:
Update the Epi api to allow for creation and in many cases updating of epi content in HAWC via the API.  This allows data migration from external tools into HAWC programmatically, consistent with recent updates which allow bioassay data to be loaded via API (see #222).

--- 

Discussion points below:

@shapiromatron had some branching/merging issues and just ended up being simpler to make a new branch and a new draft PR for this with the changes stemming from the code review included. We're currently testing this one on the ICF dev instance. I think these changes address most of the notes from your comments, with a couple of followups:

1. **re: IdLookupMixin**
I looked into replacing e.g.
`study_pop = StudyPopulationSerializer() # where StudyPopulationSerializer mixes in IdLookupMixin`
\
with
`study_pop = serializers.PrimaryKeyRelatedField(queryset=models.StudyPopulation.objects)`.
\
This works; I can pass in the number for the studypop during e.g. a comparison set upload. But if I do that, then all I get back in the returned JSON is the id, not the full object. Various existing tests/exports assume the full field set is going to be there. So I just relocated IdLookupMixin into common/serializers as suggested, and didn't replace it.

2. **re: "this will fail if a user isn't a project manager..."**
It's working correctly for me? I tried it with a user who is just a team member, not a project manager, and it is I think correctly granting them permissions and not throwing an exception.

3. **re: explicit permissions unit tests for PermCheckerMixin**
Not sure how to write one of these? I think we get decent coverage via the tests of the individual API endpoints in any case (e.g. TestOutcomeApi::test_permissions).

4. **re: removing find_matching_list_element_by_value**
They're certainly approaching similar territory but I'd argue to keep this; I think it's a useful utility function (I'm using it in a few places, such as in GroupResult::flat_complete_data_row) and I could see wanting to have this functionality even when not as part of a FlexibleChoiceField. FCF does also match by either key or value so not exactly the same.
\
I did rework FCF to use this function so that cuts down on the duplication somewhat. (Also renaming it to the                                        slightly more unwieldy yet more accurate "find_matching_list_element_value_by_value", since it returns a single value from the found element, and not the entire matching element itself.)                                                 

5. **re: transaction wrapping**
I've already got the transaction.atomic decorator on the create/update methods which call perform_create/perform_update. I thought that would accomplish what you're after (and it matches how the animal api approaches it); I can move it to the perform_x level though if you prefer.

6. **re: "why can't we use an AssessmentViewset or AssmentEditViewset..."**
I must not be understanding how to properly use this one b/c I get what seems to be incorrect behavior. Ex:
\
`client = APIClient()`
`client.login(username="reviewer@hawcproject.org", password="pw")`
`response = client.post("/epi/api/study-population/", some_data_object, format="json")`
\
I'd expect this to be permission denied. But if I just have StudyPopulation extend AssessmentEditViewset, eventually AssessmentLevelPermissions.has_permission gets called with a view.action of "create"; this isn't in the "list_actions" (and I don't think it should be), and so that method returns True. I would have expected that method to default to False?


hawc/apps/epi/api.py
```diff
@@ -87,37 +99,410 @@ def result_heatmap(self, request, pk):
...
+    read_serializer_class = serializers.ExposureSerializer
+    write_serializer_class = serializers.ExposureWriteSerializer
+
+    def handle_dtxsid(self, request):
+        """
+        Checks supplied dtxsid if present for existence in the database; attempting to create on-the-fly if necessary.
+        """
+        if "dtxsid" in request.data:
+            dtxsid_probe = request.data["dtxsid"]
+            try:
```
shapiromatron, line 265: ideally this logic would be on `assessment.models.DSSTox` overriding the `get_or_create` method, since it's not epi-specific

hawc/apps/epi/api.py
```diff
@@ -87,37 +99,410 @@ def result_heatmap(self, request, pk):
...
+    def create(self, request, *args, **kwargs):
+        self.handle_dtxsid(request)
+
+        # Default behavior except we need to refresh the serializer to get the central tendencies to show up in the return...
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        self.perform_create(serializer)
+
+        # Refresh serializer instance, as we want to get the central tendencies we created...
+        instance = self.model.objects.get(id=serializer.data["id"])
```
shapiromatron, line 294: can you just do `serializer.instance.refresh_from_db()` here and elsewhere?

hawc/apps/epi/api.py
```diff
@@ -87,37 +99,410 @@ def result_heatmap(self, request, pk):
...
+                            fixed.append(adj_factor.id)
+                        except ObjectDoesNotExist:
+                            # Allow creation of adjustment factors as part of the request
+                            af_serializer = serializers.AdjustmentFactorSerializer(
+                                data={"description": el, "assessment": assessment_id}
+                            )
+                            af_serializer.is_valid(raise_exception=True)
+                            af = af_serializer.save()
+                            fixed.append(af.id)
+                    else:
```
shapiromatron, line 459: any type checking here, or anythign goes as long as it's not a str?

hawc/apps/epi/serializers.py
```diff
@@ -42,6 +50,16 @@ class Meta:
         model = models.Country
         fields = ("code", "name")
 
+    def to_internal_value(self, data):
+        if type(data) is str:
+            try:
+                country = self.Meta.model.objects.get(code__iexact=data)
+                return country
+            except ObjectDoesNotExist:
+                raise serializers.ValidationError(f"Invalid country code '{data}'")
```
shapiromatron, line 59: fix error message; not a country

hawc/apps/epi/serializers.py
```diff
@@ -136,14 +189,24 @@ class Meta:
         fields = "__all__"
 
 
+class ExposureWriteSerializer(serializers.ModelSerializer):
+    central_tendencies = CentralTendencySerializer(many=True, read_only=True)
+    form_integration_class = forms.ExposureForm
```
shapiromatron, line 194: is form_integration_class used anywhere? if not, delete from here and elsewhere

hawc/apps/epi/serializers.py
```diff
@@ -90,41 +114,70 @@ class Meta:
...
     url = serializers.CharField(source="get_absolute_url", read_only=True)
-    design = serializers.CharField(source="get_design_display", read_only=True)
 
     class Meta:
         model = models.StudyPopulation
         fields = "__all__"
 
+    def update(self, instance, validated_data):
+        if "countries" in validated_data:
+            instance.countries.clear()
```
shapiromatron, line 158: instead of clear/add, use set

https://docs.djangoproject.com/en/3.1/ref/models/relations/#django.db.models.fields.related.RelatedManager.set

hawc/apps/epi/serializers.py
```diff
@@ -90,41 +114,70 @@ class Meta:
...
+            instance.countries.add(*validated_data["countries"])
+            # Delete the key so we can call the default update method for all the other fields.
+            del validated_data["countries"]
+
+        return super().update(instance, validated_data)
+
+    def create(self, validated_data):
```
shapiromatron, line 165:
```python
def create(self, validated_data):
        countries = validated_data.pop('countries',None)
        instance = super().create(validated_data)
        if countries:
            instance.countries.set(countries)
        return instance
```

tests/hawc/apps/common/test_common_serializer.py
```diff
@@ -47,3 +54,120 @@ def test_get_matching_instances(db_keys):
     ]:
         with pytest.raises(ValidationError):
             get_matching_instances(Study, data, "study_id")
+
+
+@pytest.mark.django_db
+def test_flexible_choice_field(db_keys):
```
shapiromatron, line 60: nice tests!

tests/hawc/apps/epi/test_epi_api.py
```diff
@@ -71,3 +76,1652 @@ def test_result_heatmap(self, rewrite_data_files: bool, db_keys):
...
+            study_pop_id = resp.json()["id"]
+            study_pop = models.StudyPopulation.objects.get(id=study_pop_id)
+            assert study_pop.name == "updated"
+            assert study_pop_id == just_created_study_id
+
+        def deleted_study_pop_test(resp):
+            nonlocal just_created_study_id
+
+            assert resp.data is None
+            try:
```
shapiromatron, line 223: rewrite this type of thing in your test to 

```python
with pytest.raises(ObjectDoesNotExistError):
    models.StudyPopulation.objects.get(id=just_created_study_id)
```

https://docs.pytest.org/en/stable/reference.html#pytest.raises

tests/hawc/apps/epi/test_epi_api.py
```diff
@@ -71,3 +76,1652 @@ def test_result_heatmap(self, rewrite_data_files: bool, db_keys):
...
+            criteria_id = resp.json()["id"]
+            criteria = models.Criteria.objects.get(id=criteria_id)
+            assert criteria.description == "updated"
+            assert criteria_id == just_created_criteria_id
+
+        def deleted_criteria_test(resp):
+            nonlocal just_created_criteria_id
+
+            assert resp.data is None
+            try:
```
shapiromatron, line 366: pytest.raises here and elsewhere

tests/hawc/apps/epi/test_epi_api.py
```diff
@@ -71,3 +76,1652 @@ def test_result_heatmap(self, rewrite_data_files: bool, db_keys):
...
+
+        if "expected_content" in scenario:
+            assert str(response.data).lower().find(scenario["expected_content"].lower()) != -1
+
+        if "post_request_test" in scenario:
+            scenario["post_request_test"](response)
+
+
+def generic_perm_tester(url, data):
+    print(f">>>>> generic perm test on {url}")
```
shapiromatron, line 1699: remove prints

tests/hawc/apps/epi/test_epi_api.py
```diff
@@ -71,3 +76,1652 @@ def test_result_heatmap(self, rewrite_data_files: bool, db_keys):
...
+    client = APIClient()
+    response = client.post(url, data, format="json")
+    assert response.status_code == 403
+
+
+def generic_get_any(model_class):
+    return model_class.objects.all().first()
+
+
+def generic_merge_overrides(data, overrides):
```
shapiromatron, line 1717: you dont need this. i think you can just call `data.update(overrides)`

shapiromatron (changes requested):
I think we're close here, most of these changes are pretty minor and we've got good test coverage. Do you think you could try to get them working before our call  Monday afternoon? CC @shamsdahnish 

Also, this second PR that was different from the first made it hard for me to qa - in the future, even if it's a little painful, could you try to use the same PR? I can help if you're having git branch issues.

hawc/apps/common/helper.py
```diff
@@ -289,3 +289,51 @@ def build_export(self) -> FlatExport:
 
 
 re_digits = r"\d+"
+
+
+def find_matching_list_element_value_by_value(
```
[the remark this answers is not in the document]
DookTibs: The kwargs are being used in GroupResult::flat_complete_row_data. (I did remove get_p_value_qualifier_display, at some point I think I had it doing something different but yea that is just the built-in behavior so that could go).

hawc/apps/common/api/viewsets.py
```diff
@@ -50,3 +50,71 @@ def fields(self, request, format=None):
...
+
+        for checker_key in checker_keys:
+            if checker_key in serializer.validated_data:
+                things_to_check.append(serializer.validated_data.get(checker_key))
+
+        if append_self_obj:
+            things_to_check.append(self.get_object())
+
+        if len(things_to_check) == 0:
+            raise Exception(f"Improperly configured viewset {self}; needs defined perm_checker_key")
```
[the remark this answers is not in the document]
DookTibs: I don't think I can just guard at the start; I want to ensure that the programmer using it actually provided valid keys and I can't know that til I check them. I have updated the method slightly to have a better error message that reflects this however, and changed it to raise a new subclass of Exception (ValueError doesn't seem right to me here).

hawc/apps/common/helper.py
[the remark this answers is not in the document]
shapiromatron: Did the serializer change here? I'd just write a separate function to handle this, since it's really doing a different thing

hawc/apps/epi/models.py
```diff
@@ -1855,6 +1871,14 @@ def copy_across_assessments(self, cw):
         self.save()
         cw[self.COPY_NAME][old_id] = self.id
 
+    def get_assessment(self):
+        return self.result.get_assessment()
+
+    def get_p_value_qualifier_display(self):
+        return find_matching_list_element_value_by_value(
```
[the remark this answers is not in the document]
DookTibs: yup removed, not needed at this point.

hawc/apps/common/helper.py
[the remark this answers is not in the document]
DookTibs: for me it was failing on the test_epi_api.py::test_full_export method. By the time flat_complete_data_row is called it's just getting a a dictionary of values so for instance a p_test_qualifier of "-" was showing up as "-" and not "n.s". (but maybe that's actually desirable and the test needs to change to match?)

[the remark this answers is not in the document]
shapiromatron: It's because the serializer changed. I f we change the serializer to `ser["p_value_qualifier_display"]` does that do it?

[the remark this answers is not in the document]
DookTibs: yup, that works. OK, renamed/simplified the function.

hawc/apps/epi/api.py
DookTibs: if it's anything other than a string, you'll get normal django checks/errors, for instance if you pass in a bool you'll get:

`[`
`  {`
`    "adjustment_factor": [`
`      "Invalid pk \"True\" - object does not exist."`
`    ]`
`  }`
`]`

tests/hawc/apps/epi/test_epi_api.py
DookTibs: neat, didn't know about that one.

DookTibs:
ok @shapiromatron I think latest push addresses the comments from over the weekend, we can discuss in a bit. Thx!

DookTibs:
I see the from the CI checks that linting is failing on that new exceptions.py; it is passing on my end though (both with my editor plugin, and if I do a manual make lint-py on it. Any ideas why I wouldn't get the error locally?

hawc/apps/common/helper.py
```diff
@@ -289,3 +289,51 @@ def build_export(self) -> FlatExport:
...
+        items (list, tuple): list of items to search
+        lookup_value: the value to search for
+        case_insensitive (bool): perform case-insensitive matching, if lookup_value is a string
+        lookup_index (int): the index in which to look for matches, in each element of items
+        return_index (int): the index of the value to return, if a match is found
+
+    Returns:
+        The value, if a single match can be found.
+
+        Returns None if multiple/no matches are found.
```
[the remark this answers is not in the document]
DookTibs: done. still returns None on no match; throws exception if multiple.

hawc/apps/epi/api.py
DookTibs: moved, there's still a little left in epi/api.py but the bulk of it (the looking up of the substance and such) is now on DSSTox::save so should be usable outside of epi.

hawc/apps/epi/api.py
DookTibs: done; I think I modeled this after what the animal api is doing as it uses similar calls. Swithed to refersh_from_db.

hawc/apps/epi/serializers.py
DookTibs: yup sorry, remnant of FormIntegrationMixin. killed all these.

shapiromatron (approved):

PR #412, opened 2021-03-29
Fix Pubmed parser in 3.9

scott-8:
Python 3.9 removes `getchildren()`.

https://docs.python.org/3.8/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.getchildren


shapiromatron (approved):
LGTM! thanks for the docs link

PR #413, opened 2021-04-01
Use correct endpoint-override riskofibas scores in visuals and other exports

rabstejnek:
# Exploratory Heatmap

A good test candidate: http://127.0.0.1:8000/ani/assessment/100000061/heatmap-endpoints/?unpublished=true

To show the changes we are looking at a study with the following score overrides:

image

## Before
Endpoints used the default RoB scores

image

## After
Endpoints prioritize endpoint overrides -> animal group overrides -> default scores

image

# Data Pivot

To show changes on data pivot we are looking at the following study/metric:

image

Data pivots now show override specific scores:
Screen Shot 2021-05-26 at 6 39 39 PM


hawc/apps/materialized/sql.py
```diff
@@ -0,0 +1,27 @@
+FinalRiskOfBiasScore = (
```
rabstejnek, line 1: We need to separate the SQL statements that create/drop the materialized views from the migrations so that we can use them in commands. I thought it best to keep them in a separate module, and then import them into `models.py` for class specific create/drop, just so that `models.py` doesn't get clogged up with all of these lines. Do you like this approach, or should we just put this inline in `models.py`?
(I haven't replaced migrations with this yet, I'll do that after we find a home for these)

hawc/apps/materialized/management/commands/create_materialized_views.py
```diff
@@ -0,0 +1,28 @@
+from django.apps import apps
```
shapiromatron, line 1: An idea - remove create and delete, and just keep refresh? The refresh will create if needed.  I can't think of a use-case for delete, unless you can?

hawc/apps/materialized/management/commands/refresh_materialized_views.py
```diff
@@ -0,0 +1,28 @@
+from django.apps import apps
```
shapiromatron, line 1: I'd just make them create all and refresh all mvs - if we're calling this for one I think we're calling for all of them.  I think I'd just remove the optional argument to simplify.

hawc/apps/materialized/managers.py
```diff
@@ -1,10 +1,151 @@
...
+            return [
+                (score["metric_id"], score)
+                for score in score_values
+                if endpoint.animal_group.experiment.study_id == score["study_id"]
+                and score["content_type_id"] == content_type_id
+                and endpoint.id == score["object_id"]
+            ]
+
+        for endpoint in endpoints:
+            for metric_id, score in default_scores(endpoint, score_values):
```
shapiromatron, line 68: could be a good use of itertools.chain

https://docs.python.org/3/library/itertools.html#itertools.chain

hawc/apps/materialized/managers.py
```diff
@@ -1,10 +1,151 @@
...
+        results = Result.objects.filter(pk__in=result_ids).select_related(
+            "outcome__study_population"
+        )
+
+        score_values = self.values()
+        result_scores = {result_id: {} for result_id in result_ids}
+
+        for result in results:
+            study_id = result.outcome.study_population.study_id
+            for metric_id, score in self._default_tuples(score_values, study_id):
```
shapiromatron, line 91: again, use of itertools.chain

hawc/apps/materialized/models.py
```diff
@@ -26,6 +27,20 @@ class Meta:
         abstract = True
         managed = False
 
+    @classmethod
+    def sql(cls) -> Tuple[str, str]:
```
shapiromatron, line 31: make a new data type so its clearer what these things are? 

https://docs.python.org/3/library/typing.html#typing.NamedTuple

shapiromatron (changes requested):
Looking great so far. I made a few comments here, but may make a few more after I get a chance to check this out and play with it.

hawc/apps/materialized/management/commands/create_materialized_views.py
rabstejnek: I find it useful as a shorthand for recreating mvs when I'm making changes to the sql during development, but I suppose outside of that the delete command doesn't have much use. I can always do it manually or revert migrations but I thought to add it for completeness / this one use-case.

hawc/apps/materialized/management/commands/refresh_materialized_views.py
rabstejnek: The only problem with combining create / refresh is that indices are also made on create, so subsequent runs of this combined command will duplicate the indices.

hawc/apps/materialized/models.py
```diff
@@ -72,3 +87,80 @@ class FinalRiskOfBiasScore(MaterializedViewModel):
     content_type = models.ForeignKey(ContentType, null=True, on_delete=models.DO_NOTHING)
     object_id = models.IntegerField()
     content_object = GenericForeignKey("content_type", "object_id")
+
+    @classmethod
+    def get_dp_export(cls, assessment_id: int, ids: List[int], data_type: str) -> Tuple[Dict, Dict]:
```
rabstejnek, line 92: The `ids` parameter changes depending on data_type. Maybe it would be better to have separate functions like `get_animal_dp_export`, etc so that we don't have this ambiguous parameter. What do you think?

rabstejnek: A lot of this function can be moved to a private function for reuse, so I don't think it would add too many more lines of code.

shapiromatron (approved):
Really well done. Took me all day to review b/c it was so complex, I made some minor tweaks just as was working through it but it was fine as is.  Here were my changes:

- remove extra args to mgmt commands and built-into load-fixture command since they're always done together (less feature-complete, but less code to manage since we didn't have tests for the extra commands)
- fixed a broken test (changes I made to the db fixture broke this test when it previously worked) - though using the next() iterator command for comparison was confusing; is that the same as an any()?
- added epi.Exposure to the list of override options for epi to be consistent with UI
- performance tweak - always filter list to study-list  first before the itertools.chain() since that's the slow part; I tested on a complex assessment and it seemed quick enough though, nice! 
- code readability tweak - changed a tuple to NamedTuple to make it easier to read and more precise typing

PR #415, opened 2021-04-09
Middleware - Referer checking on safe http methods

shapiromatron:
Add new middleware to validate Referer header on safe requests, per updated security scanning requirements. 

CSRF checking in django middleware is only performed for "unsafe" methods per django [documentation](https://docs.djangoproject.com/en/3.1/ref/csrf/#how-it-works) and [IETF](https://tools.ietf.org/html/rfc7231.html#section-4.2.1).   This PR adds Referer middleware checking on safe requests such as a GET. If the Referer doesn't match the server host name, the middleware returns a valid html page that requests a refresh. This will modify the Referer header to the current domain as expected.

This would only be an real issue if a GET request modified server state and used a Referer header to do so, which isn't a real-world condition for this application.

To test real-world conditions, I made a mock website, and started the django server as normal on port 8000:

```html
<html>
    <body>
        <a href="http://127.0.0.1:8000">GO!</a>
    </body>
</html>
````

Start a webserver via:

```bash
python -m http.server 5000
```

When clicking the link to the django site from an external site, the page load as expected, but the first (small) request issues a refresh:

```text
INFO "GET / HTTP/1.1" 200 79
INFO "GET / HTTP/1.1" 200 23633
```

### DIFF
--- a/hawc/apps/common/middleware.py
+++ b/hawc/apps/common/middleware.py
@@ -23,3 +23,14 @@ class MicrosoftOfficeLinkMiddleware(MiddlewareMixin):
             return HttpResponse(self.RESPONSE_TEXT)
         response = self.get_response(request)
         return response
+
+
+class RemoveRefererMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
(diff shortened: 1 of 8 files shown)


rabstejnek (approved):
Looks good!

PR #417, opened 2021-04-09
vocabulary terms - API for bulk create/update terms

rabstejnek:
## API endpoints
Added new API endpoints to assist in migrating Synaptica data into HAWC.
- `/vocab/api/term/uids/` : Returns a list of id, uid tuples for all terms in HAWC. This helps to map UIDs from Synaptica to their HAWC counterparts.
- `/vocab/api/term/bulk-create/` : Creates terms from a list of serialized data
- `/vocab/api/term/bulk-update/` : Updates terms from a list of serialized data

### _Other changes_
- Removed `Comment` model
- Removed unused views


hawc/apps/common/serializers.py
```diff
@@ -283,3 +286,119 @@ def run_validators(self, value):
...
+        return data
+
+    def values_equal(self, instance, field, value) -> bool:
+        return getattr(instance, field) == value
+
+    def update_field(self, instance, field, value) -> str:
+        setattr(instance, field, value)
+        return field
+
+    def update_fields(self, instance, data) -> Tuple[bool, List]:
```
shapiromatron, line 371: add a docstring to explain return values

hawc/apps/common/serializers.py
```diff
@@ -283,3 +286,119 @@ def run_validators(self, value):
...
+        fields = list(data.keys())
+        fields.remove("id")
+        now = timezone.now()
+        updated_fields = []
+        for field in fields:
+            value = data[field]
+            if not self.values_equal(instance, field, value):
+                set_field = self.update_field(instance, field, value)
+                updated_fields.append(set_field)
+        updated = bool(updated_fields)
```
shapiromatron, line 381: rename updated to requires_update?

hawc/apps/common/serializers.py
```diff
@@ -283,3 +286,119 @@ def run_validators(self, value):
...
+        if updated and hasattr(instance, "last_updated"):
+            instance.last_updated = now
+            updated_fields.append("last_updated")
+        return updated, updated_fields
+
+    def update(self, instances, validated_data):
+        updated_instances = []
+        updated_fields = set()
+        for data in validated_data:
+            instance = next(instance for instance in instances if instance.id == data["id"])
```
shapiromatron, line 391: use a dict:

```python
instances_dict = {instance.id:instance for instance in instances}
```

hawc/apps/vocab/api.py
```diff
@@ -83,3 +85,44 @@ def related_entity(self, request: Request, pk: int = None) -> Response:
...
+    serializer_class = serializers.TermSerializer
+    queryset = models.Term.objects.all()
+
+    @transaction.atomic
+    @action(
+        detail=False, methods=("post",), permission_classes=(IsAdminUser,), url_path="bulk-create"
+    )
+    def bulk_create(self, request: Request) -> Response:
+        # ensure data is list of dicts
+        try:
```
shapiromatron, line 100: can we move this try/except to a helper method for this viewset for both bulk_createa and bulk_update?  and i think it should say "data must be an array of objects" since that's more JSON-specific than dicts in python.

On a related note; is this needed? I think lots of this logic is in the init of the BulkSerializer

hawc/apps/vocab/serializers.py
```diff
@@ -1,9 +1,36 @@
...
 from rest_framework import serializers
 
+from ..common.serializers import BulkSerializer
 from . import models
 
 
+class TermBulkSerializer(BulkSerializer):
+    def values_equal(self, instance, field, value):
+        # equality for deprecated_on is whether it needs to be set
+        if field == "deprecated_on":
```
shapiromatron, line 11: 👏

hawc/apps/common/serializers.py
```diff
@@ -283,3 +286,119 @@ def run_validators(self, value):
...
+                errors.append({})
+
+        if any(errors):
+            raise ValidationError(errors)
+
+        return ret
+
+    def validate_update(self, data):
+        # ids must be in data
+        data_ids = [obj_data.get("id") for obj_data in data if obj_data.get("id") is not None]
```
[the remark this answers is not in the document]
rabstejnek: The second check (all data ids should be in instance) is necessary for our bulk update to run without error. I could instead catch it within the `update` method, but I put it here instead so that all of our validation is in one place.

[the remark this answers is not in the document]
shapiromatron: oh yeah sorry, i wasn't clear. yes, please keep the second check, i'm just thinking you can simplify the first check and use the same data_ids in the second check.

hawc/apps/vocab/api.py
rabstejnek: `ListSerializer` does check if its a list and throws a `ValidationError` if it isn't, but if I were to pass it a list of ints or some other invalid list I think it would result in a 500 error. I'll reconfirm in a moment after some quick testing

shapiromatron: ok, if it's needed keep it for sure, just wanted to make sure. please use a single shared method though for both the update and the create

rabstejnek: Oops, it looks like it was an error of my own creation; when I copied the inherited method verbatim, I didn't take into account that `ValidationError` in `common\serializers.py` was the Django exception, so the inherited method was throwing that instead of the DRF error. Its fixed now and is thrown correctly from the serializer, so I should be able to remove this from `api.py`!

hawc/apps/common/serializers.py
```diff
@@ -283,3 +286,164 @@ def run_validators(self, value):
...
+    "id" must be given write access on serializer for bulk update
+    (eg.  id = IntegerField(required=False))
+    """
+
+    def to_internal_value(self, data):
+        """
+        This is the inherited method from ListSerializer, with a few lines added to work around an error.
+        `exclude_current_instance` on `UniqueValidator` examines the `instance` set on the serializer to exclude it from the uniqueness constraint:
+        https://github.com/encode/django-rest-framework/blob/3875d3284e73ed4d8e36c07d9b70c1b22c9d5998/rest_framework/validators.py#L138
+        However, `instance` in this case is the queryset passed in to the constructor, so it throws an error when it attempts to retrieve the `pk` attribute of the queryset.
```
shapiromatron: minor nit; hard to read b/c the line is so long; no hard rule but I try to keep <100

shapiromatron (approved):
Nice job! Made one minor change to shorten length of docstrings. LGTM!

PR #418, opened 2021-04-16
Assessment permissions cache

shapiromatron:
Build a cache for all assessment setting related to view permissions.  Currently, each view is slowed by querying the assessment, plus multiple many to many relations to check if a specific user has permission to view. This PR implements a cache that's refreshed whenever the assessment is saved, which captures all user ids with permission and other assessment settings in a cache. 

Also update development settings to have a single setting to enable or disable cache in dev mode.

The net result of this PR is a speedup for view permission checking and moving db queries to a cache.

### DIFF
--- a/hawc/apps/assessment/models.py
+++ b/hawc/apps/assessment/models.py
@@ -25,6 +25,7 @@ from ..common.models import IntChoiceEnum, get_private_data_storage
 from ..myuser.models import HAWCUser
 from ..vocab.models import VocabularyNamespace
 from . import jobs, managers
+from .permissions import AssessmentPermissions
 from .tasks import add_time_spent
 
 NOEL_NAME_CHOICES_NOEL = 0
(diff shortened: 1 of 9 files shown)


shapiromatron:
ok, ready for real now @rabstejnek

hawc/apps/assessment/models.py
```diff
@@ -308,83 +309,43 @@ class Meta:
     def __str__(self):
         return f"{self.name} ({self.year})"
 
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
+        AssessmentPermissions.clear_cache(self.id)
```
rabstejnek, line 314: Most of our cache updates are done as signals. This would also let it run on `delete`

rabstejnek (approved):
Just one comment on where we should be updating our caches, if we wanted to try and keep up a pattern of doing it in signals. Otherwise it looks good!

hawc/apps/assessment/models.py
shapiromatron: Oh sorry, I dont have a super-strong opinion on this. I think if the change introduces side-effects or is complex, it may make sense to change to a signal.  I've read some recent opinionated blog posts that suggested sometimes just overriding the save is cleaner than signals but I could see it argued either way.

I'm not worried about cleaning up the cache w/ deletes.

I can move to a signal if you'd prefer, just let me know...

shapiromatron: It seems like just busting a cache based on a single model is kind of model-specific w/ few side effects, which is probably why I put it here...

rabstejnek: I can see it going either way, I was mostly just curious of your thoughts on it. I'm fine with it in its current state!

PR #421, opened 2021-04-19
Miscellaneous error fixes

rabstejnek:
## Bad related field
The related field was `domain_assessment`, when it should have been `domain__assessment`

image
image

## Cannot call assessment method on None object
`get_assessment_from_query` returns either the assessment or none if `assessment_id` does not correspond with an assessment. It works in all contexts where this function is called to just have it raise an exception if the assessment does not exist instead.

image
image

Another case I found:

image
image

## Anonymous user error
This error is thrown when trying to filter by an anonymous user. The fix is to check authentication instead.

image
image
__*Note:__ Should we even be rendering this page as an anonymous user? It serves as a launch page for copying searches, which anonymous users wouldn't be able to do to begin with (the `Searches` dropdown doesn't have any options for anonymous users)

### DIFF
--- a/frontend/riskofbias/constants.js
+++ b/frontend/riskofbias/constants.js
@@ -131,9 +131,11 @@ const NA_KEYS = [10, 20],
     },
     getMultiScoreDisplaySettings = function(scores) {
         // Return visualization/color settings for situations where multiple scores may exist for
-        // a given metric (eg, study-level override settings)
+        // a given metric (eg, study-level override settings).
+        // By default, if multiple scores exist and show he defaults score label if one exists.
+        // If the default score does not exist, present the value of the first score (random).
         let sortedScores = _.orderBy(scores, "score", "desc"),
-            defaultScore = _.find(scores, {is_default: true}),
+            defaultScore = _.find(scores, {is_default: true}) || sortedScores[0],
             shades = _.chain(sortedScores)
                 .map(score => score.score_shade)
                 .uniq()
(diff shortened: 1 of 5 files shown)


rabstejnek:
Two of the errors have the same problem:

image
image

For some reason, there are results in our database that are not constrained by the designated dose response choices:

image

@shapiromatron Is there a chance that this data was imported into HAWC through pure SQL and bypassing the Django ORM? I'm not sure how else the data would have gotten unallowed values.

shapiromatron:
@rabstejnek those have changed with time. Here's a screenshot from https://github.com/shapiromatron/hawc/pull/154/files#diff-f00e1d242b23d1fc71d5edb7bfb5574cd4af52eeb39e8fed08065a6a21e89ecbL654

Screen Shot 2021-04-19 at 3 14 09 PM

I will find the remapping we did in the archives... looks like we need to apply here...

rabstejnek:
@shapiromatron The error is coming from bad `DOSE_RESPONSE_CHOICES` keys, which haven't been changed: https://github.com/shapiromatron/hawc/blame/7093a1c28e8119bddac13c0104441d991aeaf920/hawc/apps/epi/models.py#L1225

The good news is that once I updated all the invalid keys with valid keys the export works, so it doesn't look like the error goes any deeper than this.

shapiromatron:
History is wide and deep, it is :) https://github.com/teamhero/hawc/blob/master/project/epi/models.py#L1287-L1296

Maybe we missed a mapping here?  Is everything ready with this PR except this migration?

shapiromatron (approved):

shapiromatron:
Added one more error fix. In some cases when building rob heatmaps, a user may not want to present the default score and only show override scores for a given study/metric combination, like this:

Screen Shot 2021-04-23 at 4 05 02 PM

Previously, this would result in an error because the `defaultScore` in a multi-score display would be undefined; now we fallback to the first score.

PR #422, opened 2021-04-20
risk of bias/study evaluation tabular export timestamps

rabstejnek:
In the tabular exports, add `rob-created` and `rob-last_updated` timestamps for when rob responses were last modified for each study. This is useful for QA/QC purposes.


hawc/apps/riskofbias/exports.py
```diff
@@ -23,17 +24,16 @@ def _get_data_rows(self):
             row.extend(Study.flat_complete_data_row(ser))
 
             try:
-                scores = [
```
shapiromatron, line 26: this could be problematic to maintain if we add new fields. Could we use `common.helper.df_move_column` instead?

hawc/apps/riskofbias/exports.py
```diff
@@ -13,6 +13,7 @@ def _get_header_row(self):
         header = []
         header.extend(Study.flat_complete_header_row())
         header.extend(models.RiskOfBiasScore.flat_complete_header_row())
```
shapiromatron, line 15: why not just add these two items to the end of `models.RiskOfBiasScore.flat_complete_header_row`?

shapiromatron (commented):
Such strange legacy code. One question on the header override

shapiromatron (approved):
@rabstejnek looks fine to me. I left a comment that it was hard to QA the code just based on the prior implementation you were working off of. I made a large change to how this is implemented and reduces the duplicative logic code in two different classes. 

Could you give it a quick scan to let me know if you're happy with my changes?

rabstejnek:
I checked out the changes and I'm happy with it!

PR #423, opened 2021-04-21
hotfix - fix tests in CI/CD

shapiromatron:
We inadvertently had a unit-test which used an external webservice which is periodically down. This update temporarily disables that test, and in the future we'll use [vcr](https://vcrpy.readthedocs.io/en/latest/) to cache the response when it becomes available in the future.

At a future date the commented-test will be uncommented and a cached response will be saved.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,6 +35,7 @@ jobs:
     - name: install
       run: |
         mv ci-webpack-stats.json webpack-stats.json
+        pip install -U pip wheel
         pip install -r requirements/dev.txt
     - name: lint
       run: |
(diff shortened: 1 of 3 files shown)


shapiromatron:
@DookTibs this more of an fyi; it looks like that dsstox creation test crashes our test suite if actor webservices is down; we'll be able to re-enable this test once the service is up and we can cache the response

rabstejnek (approved):

DookTibs (approved):
cool, thx for the heads up. sorry if that was causing problems!

shapiromatron:
> cool, thx for the heads up. sorry if that was causing problems!

NBD; its a quick fix once we figured it out!

PR #426, opened 2021-04-23
Paginated lit.Reference API

rabstejnek:
Create new REST API for paginated API for lit.Reference mirror the current TagsJSON API.  The following
routes are possible:

- /lit/api/assessment/:assessment_id:/references/?tag_id=:tag_id:
- /lit/api/assessment/:assessment_id:/references/?tag_id=:tag_id:&search_id=:search_id:
- /lit/api/assessment/:assessment_id:/references/?tag_id=untagged

The following frontend changes were made:

- paginate references on lit tagtree modal reference browsing
- create Paginator React component

To test, view any literature Tagtree and click a node w/ > 50 references. It should return much quicker than before and the list will have pagination at the bottom:
Screen Shot 2021-05-27 at 3 19 11 PM

Additional unrelated changes:

- found fixed a few cases where instance.assessment.id was used instead of instance.assessment_id to prevent extra queries
- wrap html on long lit.Search import strings
- filter openapi warnings on unit tests
- remove @action(methods=("get",)) since that is the default

In the future, additional changes will be made to migrate to these APIs.


shapiromatron (approved):
@rabstejnek looked good, but i went ahead and implemented using it on the frontend so it doesn't become stale; please take a look at my changes and let me know what you think...

rabstejnek:
@shapiromatron I checked it out on the front end, it looks good!

PR #435, opened 2021-05-04
hawc fixture updates + resource static content page

shapiromatron:
Assorted updates for easing acceptance of the hawc website to fit into the epa web ecosystem

- rename some content in fixtures to be more reflective on "real" content in hawc instead of test data
- add a new `resources` static content page and add to navbar

bonus
- update to `Django==3.1.10` (unrelated; [security update](https://www.djangoproject.com/weblog/2021/may/06/security-releases/))


rabstejnek (approved):
Tests are passing locally, resource page works; looks good!

PR #440, opened 2021-05-13
Sort order for domain/metric

rabstejnek:
Added a `sort_order` field to `RiskOfBiasDomain` and `RiskOfBiasMetric`, and updated the ordering on these models to use this field.

**Move up/down views:**
Screen Shot 2021-06-08 at 3 35 51 PM

**Detail view:**
Screen Shot 2021-06-08 at 12 04 20 PM


shapiromatron (changes requested):
A few changes based on earlier discussions:

1. require the sort order to have a min value of 1 instead of 0
2. move the current "edit" page into this sort, and rename "Sort requirements" to "Update requirements"

shapiromatron (approved):

PR #446, opened 2021-05-27
adding epi create/update API support to client and tests

DookTibs:
@shapiromatron Re: adding epi API support to the client. I've done for just the top level "study population" for now. As the ticket specified I've largely modeled this after the work done on the animal client.

Just wanted to touch base as I have diverged a bit on the tests; animal api has a big "test_animal_create" method where it creates experiments, animal groups, etc. I'd like to test the individual create/read/update/delete methods I'm adding in epi so I was planning to structure it more like "test_epi_study_population", and a separate "test_epi_outcomes", "test_epi_results", etc.

If that looks ok I can move ahead with the rest, this went pretty fast, and if there are changes I'd rather know at this point :-) Thx!


shapiromatron:
@DookTibs I think it'd be better to keep it a single function. The reason is when we use these client tests, the database transaction rollbacks are confusing and a lot of time we're building off parents that were previously created. 

Consistent with animal, I would focus on writing API examples for the create only. There are generally other ways the read will be done, and the update will infrequently be used. If we have time/hours remaining, we can focus on those other endpoints.

DookTibs:
@shapiromatron OK, rewrote the tests to keep more in line. This should now be set; is there a way on Github to change a draft PR to a real one?

shapiromatron:
> @shapiromatron OK, rewrote the tests to keep more in line. This should now be set; is there a way on Github to change a draft PR to a real one?

Do you see this ready for review button? I think you have to push it, assuming it's visible...
Screen Shot 2021-06-05 at 10 58 33 AM

PR #448, opened 2021-06-01
add epi models to admin

shapiromatron:
Add new epi admin views just like other models to allow browsing and search.

Screen Shot 2021-06-01 at 3 40 48 PM


rabstejnek (approved):
Looks good!

PR #449, opened 2021-06-03
sortable exploratory heatmap axes

shapiromatron:
Make exploratory heatmap axes sortable. Previously all items in the heatmap were sorted alphabetically and a user could not customize. Now they can be sorted. We attempted also adding the ability to include/exclude, but do to the way the filtering is currently implemented, this added too much additional complexity.

**Default case (alphabetical)**
Screen Shot 2021-06-03 at 5 22 42 PM

**Custom case**
Screen Shot 2021-06-03 at 5 23 04 PM
Screen Shot 2021-06-03 at 5 23 12 PM

A few edge cases (and how they are handled):
- If an item exists in the data but is not in the sorted list, they are included after sortable items at the end
- If a field has delimiter, it can be included as well
- If a field has null values those are shown as `<null>` and handled appropriately

Additional tangential changes in this PR:
- changed "<null>" as a string to h.nullString constant since this is used in many places
- catch error if heatmap rendering fails; display  "visual cannot be displayed" on page
- revised CSS for sortable draggable instances to improve clarity on what's being moved
- made the table on a Exploratory heatmap a span12 instead of span9 below the heatmap
- (unrelated) wiped assessment cache when endpoint EHV term values are updated via API (and fixed related tests)


frontend/summary/summary/heatmap/common.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const getColumnValues = function(dataset, column, delimiter) {
+    // Get column values for a column in a dataset.
+    return _.chain(dataset)
+        .map(d => d[column])
+        .map(d => (delimiter && d ? d.split(delimiter) : [d]))
+        .flatten()
+        .uniq()
+        .sort()
+        .map(d => {
```
rabstejnek, line 12: Suggested change:
`        .map(d => {
            return {id: d || "", label: d || h.nullString};
        })
        .uniqBy("id")`

rabstejnek (changes requested):
Looks good! I did notice one bug though, where two null column choices show up because we have both an empty string column and a null column:
`/summary/visual/assessment/100500085/demo-heatmap-human-populations/update/`
image
```
{
    "id": null,
    "label": "<null>"
},
{
    "id": "",
    "label": "<null>"
}
```
This is already handled in the heatmap by [making all falsy columns set to empty string](https://github.com/shapiromatron/hawc/blob/30bb03842eb7b7f0c21b4ef42b2752b11f35269c/frontend/summary/summary/heatmap/HeatmapDatastore.js#L61), so I suggest doing the same here. I have a separate comment with a proposed fix.

frontend/summary/summary/heatmap/common.js
shapiromatron: Did it a little differently so it reads better, what do you think?

```javascript
export const getColumnValues = function(dataset, column, delimiter) {
    // Get column values for a column in a dataset.  Returns unique values, and removes
    // cases where multiple falsy-like values are represented.
    return _.chain(dataset)
        .map(d => d[column])
        .map(d => (delimiter && d ? d.split(delimiter) : [d]))
        .flatten()
        .uniq()
        .sort()
        .map(d => {
            return d ? {id: d, label: d} : {id: "", label: h.nullString};
        })
        .uniqBy("id")
        .value();
};

```

rabstejnek: Looks good!

rabstejnek (approved):

PR #451, opened 2021-06-03
Select a risk of bias approach instead of automatically creating

shapiromatron:
Instead of automatically creating rob metric/domain upon assessment creation, a user now has the ability to select which questions to use for an assessment. This allows a users in any HAWC deployment to use any set of questions.  This can increase flexibility as it decouples the deployment and corresponding HAWC_FLAVOR from the default questions being used.

**Now, if there are no risk of bias metrics, this is shown:**
Screen Shot 2021-06-04 at 2 59 22 PM

**A user can select a predefined approach:**
Screen Shot 2021-06-04 at 2 59 37 PM

**This page is also available when on the "update" screen:**
Screen Shot 2021-06-04 at 3 00 44 PM

Additional related changes:

- Updated admin view for Logs to be more query-efficient and show user-information
- remove `get_flavored_text` since it is no longer used
- update risk of bias assessment help text when cloning across assessments
- simplify the UI when cloning across assessments to just point to the other page
- standardized how interactive content is shown in django-templates `HAWCUtils.onSelectChangeShowDetail`
    - used in summary visual selector, summary table selector, and rob load selector
- django form changes:
    - build generic `form_actions_apply_filters()` for filtering actions
    - build generic `form_actions_create_or_close()` for popups
    - add new `submit_text` argument to BaseFormHelper to simplify custom create names


hawc/apps/riskofbias/actions/rob_approach.py
```diff
@@ -0,0 +1,105 @@
...
+    settings.save()
+
+    # create domains and metrics
+    metrics: List[RiskOfBiasMetric] = []
+    for sort_order, domain_data in enumerate(data["domains"], start=1):
+        metrics_data = domain_data.pop("metrics")
+        domain = RiskOfBiasDomain.objects.create(assessment_id=assessment_id, **domain_data)
+        metrics.extend(
+            [
+                RiskOfBiasMetric(domain_id=domain.id, **metric_data)
```
rabstejnek, line 63: `sort_order` will need to be passed to the constructors here and elsewhere once the changes from `main` are merged into this branch.

rabstejnek (approved):
Looks good! We should probably merge our `sort_order` code into `next` if thats where this PR is headed, and then address my comment below.

rabstejnek (changes requested):
See above

hawc/apps/riskofbias/actions/rob_approach.py
shapiromatron: now added since we merged the other PR in...

rabstejnek (approved):
Looks good!

PR #452, opened 2021-06-04
Refactors

shapiromatron:


shapiromatron:
Closing; saved a patch locally in case we need it.

PR #457, opened 2021-06-11
Decoupled RoB API w/ donut and study rob detail viz integration

rabstejnek:
Added new RoB api endpoints to decouple scores and nested metrics/domains. Also passes score metadata (text, colors, symbols, etc) so that we have a single source of truth on the backend.

These new APIs have been integrated into the donut visual (`Donut`) and the study rob visuals/components (`AggregateGraph`, `RiskOfBiasDisplay`)


shapiromatron (approved):
I made a few changes, but if you approve of my changes then please feel free to merge:

- created a `robDonutStartup()` method with a react component wrapper around the Donut plot; handled lifecycle stuff outside of the d3 plot 
- removed the singleton pattern for this store, since it's a shared component and I didn't want it to load on every page where riskofbias is loaded
- moved the url methods to constants so they can be updated in one place when they change
- use _.keyBy instead of _.reduce for readability in the store
- removed `getMultiScoreDisplaySettings` from store and used constants, updating the serializer. the cost for these dict lookups is low, and that duplicate code could be problematic (the settings will still be helpful on the aggregate visuals where a legend is required or options are required)

@rabstejnek if you're ok with these changes, please merge!

rabstejnek:
Looks good!

PR #462, opened 2021-06-21
update alt-text for img and svg

shapiromatron:
- Ensure that all `img` in hawc have an `alt` attribute
- Add aria descriptors for all `svg` in hawc; handled slightly differently.

For SVG [mozilla guide](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Role_Img):

```
SVG and role="img"

If you are using embedded SVG images in your page, it is a good idea to set role="img" on the outer 
<svg> element and give it a label. This will cause screen readers to just consider it as a single entity 
and describe it using the label, rather than trying to read out all the child nodes:

<svg role="img" aria-label="Description of your SVG image">
  <!-- contents of the SVG image -->
</svg>
```

Available roles:
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles


hawc/apps/summary/templates/summary/visual_selector.html
```diff
@@ -131,7 +131,7 @@ <h2>Embedded external website</h2>
     <div id="visual6" class="col-md-12">
       <h2>Exploratory heatmap</h2>
       <div class="row">
-        <img class="col-md-10 col-md-offset-1" src="{% static '/img/summary/exploratory-heatmap.png' %}">
+        <img alt="Image of a exploratory data heatmap" class="col-md-10 col-md-offset-1" src="{% static '/img/summary/exploratory-heatmap.png' %}">
```
rabstejnek, line 134: *an

rabstejnek (changes requested):
Just one stickler grammar comment, looks good!

rabstejnek (approved):

PR #464, opened 2021-06-24
Restore riskofbias cleanup functionality with custom scores

shapiromatron:
Update riskofbias cleanup form to work with custom metric specific scores.

Good test cases:

- /assessment/100500031/clean-study-metrics/
- /assessment/100500071/clean-study-metrics/


hawc/apps/riskofbias/serializers.py
```diff
@@ -39,11 +39,10 @@ class Meta:
 
 class SimpleRiskOfBiasMetricSerializer(serializers.ModelSerializer):
     response_values = serializers.ListField(source="get_response_values")
-    default_response = serializers.IntegerField(source="get_default_response")
```
shapiromatron, line 42: @rabstejnek I removed this since it's no longe needed on the form side; let me know if you still need it and I'm missing something

hawc/apps/riskofbias/serializers.py
```diff
@@ -173,13 +172,9 @@ class Meta:
...
 
     def create(self, validated_data):
-        override = models.RiskOfBiasScore.objects.create(
-            riskofbias=validated_data["riskofbias"],
-            metric=validated_data["metric"],
-            is_default=False,
-            label="",
-        )
-        return override
+        score = validated_data["metric"].build_score(validated_data["riskofbias"], True)
```
shapiromatron, line 175: @rabstejnek I used our `build_score` method here too just so scores are all made form the same place; let me know if that looks ok too

frontend/riskofbias/robStudyForm/ScoreForm.js
```diff
@@ -35,10 +36,8 @@ class ScoreInput extends Component {
     }
 }
 ScoreInput.propTypes = {
```
shapiromatron, line 38: @rabstejnek removed scoreId and defaultValue here too, let me know if ok...

hawc/apps/riskofbias/serializers.py
rabstejnek: Looks good, but since this is the override creator we should make `is_default` false.

frontend/riskofbias/robStudyForm/ScoreForm.js
rabstejnek: The purpose of the `defaultValue` here is that when a metric is changed, any wrong score on the score form is automatically defaulted. Without it, we get the original score which will not show the correct score on the select input (since its an invalid option) and will throw an error on form submit if not handled.

rabstejnek (changes requested):
Looks good but I had a few comments. Also there's a problem with the cleanup that's similar with what we're facing with the score form.
When a metric is changed and we attempt to cleanup old scores, we get something like this:
image
The old score isn't an option, so it isn't shown correctly here in the dropdown select (though you can see it in the little badge viz). When we bulk update here it doesn't set "Yes", but instead the old score of "Definitely low risk of bias".

shapiromatron (commented):
> Looks good but I had a few comments. Also there's a problem with the cleanup that's similar with what we're facing with the score form.
> When a metric is changed and we attempt to cleanup old scores, we get something like this:
> image
> The old score isn't an option, so it isn't shown correctly here in the dropdown select (though you can see it in the little badge viz). When we bulk update here it doesn't set "Yes", but instead the old score of "Definitely low risk of bias".

Thanks! I think I've fixed the issue, requesting a second review please!

rabstejnek (approved):
Looks good!

PR #468, opened 2021-06-29
Use same components for risk of bias metric/domain read and update

shapiromatron:
Use the same react-based components for viewing and updating domains/metrics. Previously the read-based components were template based, and they were out of sync with the update forms. This PR uses the same components for both situations.


rabstejnek (approved):
Looks good!

PR #469, opened 2021-06-29
Refactor RoB backend

rabstejnek:
- Replaced old API with new decoupled API
- Condensed and renamed RoB serializers
- Reenabled several RoB tests, fixed tests


hawc/apps/summary/serializers.py
```diff
@@ -53,6 +54,8 @@ class Meta:
 
 
 class VisualSerializer(CollectionVisualSerializer):
+    rob_settings = AssessmentRiskOfBiasSerializer(source="assessment")
```
shapiromatron, line 57: This is the root cause of all the additions. should we make this a conditional to only include this if we're dealing with  `if self.visual_type in [self.ROB_HEATMAP, self.ROB_BARCHART]`, else null?

frontend/study/Study.js
```diff
@@ -14,14 +14,15 @@ import {renderStudyDisplay} from "riskofbias/robTable/components/StudyDisplay";
 import {SCORE_SHADES, SCORE_TEXT} from "riskofbias/constants";
 
 class Study {
-    constructor(data) {
```
shapiromatron, line 17: some of the stuff in Study.js will need to be reverted w/ my recent changes; i noticed in a few other places as well...

hawc/apps/riskofbias/serializers.py
```diff
@@ -3,28 +3,21 @@
 from django.db import transaction
```
shapiromatron, line 3: this is great!

hawc/apps/study/views.py
```diff
@@ -1,3 +1,5 @@
+import json
```
shapiromatron, line 1: we can revert this too now that we baked into the serializer

shapiromatron (changes requested):
one big and a few little to get us back into the red (and more importantly, to prevent us from querying rob stuff on visuals that dont need it)

tests/data/api/api-visual-barchart.json
```diff
@@ -34,6 +34,137 @@
   "url": "/summary/visual/assessment/2/barchart/",
   "url_update": "/summary/visual/assessment/2/barchart/update/",
   "url_delete": "/summary/visual/assessment/2/barchart/delete/",
+  "rob_settings": {
```
shapiromatron, line 37: `rob_settings` is shown at the root visual level and it looks like it may be repeated at the the study level for each study - should we use a different study serializer that doesn't include `rob_settings`?

tests/frontend/integration/test_integrations.py
```diff
@@ -42,8 +42,7 @@ def test_bioassay(self):
         tests.bioassay(self.driver, self.live_server_url)
 
     def test_rob(self):
-        # TODO - fix ROB June 2021
-        # tests.rob(self.driver, self.live_server_url)
+        tests.rob(self.driver, self.live_server_url)
         pass
```
shapiromatron, line 46: remove pass here; no longer needed

shapiromatron (changes requested):
One larger change where we may need a different serializer and one one-liner...

tests/data/api/api-visual-barchart.json
shapiromatron: Well... as @rabstejnek indicated this was a little more involved than I expected. 

Keeping as is even though it's sending excess data b/c it allows us to used the cached Study objects, so the additional db hits are small.

We should try to remove these in the future. I added a task to describe what would be required. The Study javascript object currently unpacks risk of bias as well. We should refactor the Study object so that the study just returns study info and there are separate api requests for rob data. Then, the visual could just get the simple study, and the final rob for a list of studies.

shapiromatron (approved):

PR #470, opened 2021-06-29
Refactor RoB stores

rabstejnek:
Refactored several RoB stores to work off of the same base class.


shapiromatron (approved):
this is nice!

PR #472, opened 2021-07-01
Sort test jsons by key before write

rabstejnek:
Sets `sort_keys=True` on all jsons written to file.

This will prevent false diffs when different computers generate the json with arbitrary order.


shapiromatron (approved):
works like a charm

PR #474, opened 2021-07-01
riskofbias  toggle metric description visibilty

shapiromatron:
Metric descriptions were often very long and made data entry and presentation confusing. They are now hidden by default and can be expanded as needed.

**Show/hide toggle:**
Screen Shot 2021-07-01 at 4 49 24 PM
Screen Shot 2021-07-01 at 4 49 35 PM

The toggle is used on:
- study rob pages: `/rob/study/:id/`
- all rob study pages: `/rob/study/:id/all/`
- study rob forms: `/rob/:id/update/`
- heatmap modals and popups:

In addition, a few more changes:
- Added generic `maybeScrollIntoView` helper function with optional animation
- Added spacers on form to better shown boundaries between metrics
- Added an "Edit final review" button on study rob pages if you're a project mgr or higher
- Remove `metric.hide_description` from forms and UI; this work on javascript enables this


rabstejnek (approved):
Looks good! Should we go ahead and remove `hide_description` from `RiskOfBiasMetric` since we have it baked into the frontend now? I don't see it used anywhere else.

shapiromatron:
> Looks good! Should we go ahead and remove `hide_description` from `RiskOfBiasMetric` since we have it baked into the frontend now? I don't see it used anywhere else.

I thought we’d leave it in temporarily, just in case users want to roll back to the old way, we’d still have the data available, but I’ll add this to the list to revisit in 6 months since it’s likely we’ll do that…

PR #475, opened 2021-07-06
WIP: Packaging using Flit

rabstejnek:


shapiromatron (changes requested):
After discussion, we decided to not use flit and update vanilla setuptools instead; the advantages were not worth using a nonstandard packaging tool.

rabstejnek:
Closing in favor of #477

PR #476, opened 2021-07-07
Worker healthcheck

shapiromatron:
Add new healthcheck url to ping to check status of workers. The healthcheck is determined by running a diagnostic test every 5 minutes via a celery cron task, and then putting the timestamp of successful execution to a redis array with a maximum length. If the last successfully executed healthcheck job was greater than 15 minutes old, the healthy response is False and a 503 http error code is used; otherwise report a 200.

New urls (requires redis cache backend to view, even in dev mode):
- `/assessment/api/healthcheck/worker/`
- `/assessment/api/healthcheck/worker-plot/` [admin only]
- `/assessment/api/healthcheck/worker-stats/` [admin only]

Also added a simple plot to view successful executions.

Screen Shot 2021-07-07 at 10 47 45 AM

In addition:
- updated to [celery v5](https://docs.celeryproject.org/en/stable/history/whatsnew-5.0.html)
- added a redis instance to our github actions test suite


rabstejnek (approved):
Looks good!

PR #479, opened 2021-07-24
JS version updates

shapiromatron:


rabstejnek:
@shapiromatron I replaced all of the `d3.group` with our custom `groupNest` which uses the `group` arguments and structures it back into the old `nest` structure. I also fixed [selection.on](https://github.com/d3/d3-selection/blob/main/README.md#selection_on) in a few places to get the rob donut + aggregation and data pivot working, but we'll need to make more widespread changes to update all of our `on` calls [[1](https://stackoverflow.com/a/63693424)].

PR #482, opened 2021-07-30
Eco

ricerb:
We've now put the application behind a feature-flag, as it's still a work in progress. To update from prior code:

```bash
# (checkout a git commit prior to this version)
manage.py migrate eco zero
# (checkout this PR or later)
export HAWC_INCLUDE_ECO=True
mange.py migrate
```

## Ideas for proposed edits

- rename "Metadata" to "Design" or protocol or conditions or something more specific
- Change relations structure. Right now it'd be difficult to add a single effect with different causes, and there's no relation between a specific metadata object and both cause and effect.  Also, make the Result a little more generic, as even without individual quantitative results, it may still be valuable.

hawc-eco


hawc/apps/eco/models.py
```diff
@@ -535,68 +146,49 @@
...
-        help_text="Select one or more Level III Ecoregions, if known",
-    )
+    name = models.CharField(verbose_name="Ecoregion", max_length=100, blank=True,)
 
     def __str__(self):
-        return self.ecoregion
+        return self.name
 
     class Meta:
         verbose_name = "Ecoregion"
```
shapiromatron, line 194: Sorry, I can't quite reach the `class Reference` model, but we should remove this model, and use a `study.Study` model in all places where it's referenced.  Also, is it always true that there's only one Metadata for a Study, or is that a ForeignKey relation

hawc/apps/eco/models.py
```diff
@@ -15,386 +15,6 @@
     ("Model", "Model"),
     ("Not applicable", "Not applicable"),
 )
-country_choices = (
```
shapiromatron, line 18: can we use the existing `epi.Country` model instead of making this one again?

shapiromatron (changes requested):
Nice, this is looking much better. Looping in @rabstejnek who can help with that binding to the study.Study model

hawc/apps/eco/models.py
```diff
@@ -15,386 +15,6 @@
...
-    ),
-    ("79. Madrean Archipelago", "79. Madrean Archipelago"),
-    ("80. Northern Basin and Range", "80. Northern Basin and Range"),
-    ("81. Sonoran Basin and Range", "81. Sonoran Basin and Range"),
-    ("82. Acadian Plains and Hills", "82. Acadian Plains and Hills"),
-    ("83. Eastern Great Lakes Lowlands", "83. Eastern Great Lakes Lowlands"),
-    ("84. Atlantic Coastal Pine Barrens", "84. Atlantic Coastal Pine Barrens"),
-    ("85. Southern California/Northern Baja Coast", "85. Southern California/Northern Baja Coast"),
-)
 habitat_choices = (
```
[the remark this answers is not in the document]
ricerb: Done.

Quick related question: how should I be handling hierarchical lists of choices? I currently have a couple instances of fields set up to filter the subsequent field - is this okay? Quantitative.measure_type_filter is one example of this. Or should these be set up as one field in django with filtering on the front end only?

hawc/apps/eco/models.py
ricerb: Done.

hawc/apps/eco/models.py
```diff
@@ -97,48 +84,27 @@ class StudySetting(models.IntegerChoices):
...
+        Vocab,
         verbose_name="Habitat",
-        choices=HabitatType.choices,
+        limit_choices_to=Q(category="Habitat")
+        & Q(parent__isnull=True),  # not sure if isnull is best approach here
+        on_delete=models.CASCADE,
         blank=True,
         help_text="Select the habitat to which the evidence applies",
+        related_name="+",
     )
```
ricerb, line 97: @shapiromatron am I headed in the right direction here? I created a Vocab model with category, value, and parent fields and am filtering it to replace the IntegerChoices fields I had originally.

hawc/apps/eco/models.py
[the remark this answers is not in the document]
shapiromatron: Resolving this convo; we had a chat about different strategies, there's no easy answer.

hawc/apps/eco/models.py
shapiromatron: two options -

a.  we could have two categories, `habitat_type` and `habitat`. we can add additional validation to make sure that the a habitat's parent is the selected habitat type.
b. we could infer habitat_type from habitat without any data entry

ricerb:
Hey @shapiromatron I spent some time last week and this morning polishing the eco models. If this looks good to you, can we put this on the dev server again for Caroline to do a test data extraction in the admin view?

shapiromatron (approved):
@ricerb made a number of changes, here's a few things that were more about the coding style and design:

- convert to f-strings
- moved meta to top
- removed lots of the `blank=True`
- made more fields required to add a little more structure
- worked on the admin side to some simple changes to hopefully improve usability
- reformatted the fixtures to take up less vertical space; improve readability
- move the choice types to a `constants.py` file

If you're ok with the changes after testing, feel free to merge!

ricerb:
@shapiromatron this looks good to me!

I added NA and NR options to some of the fields where "blank=True" was removed in order to reduce ambiguity for fields with missing data. We'll need to ask the scientists to confirm that these are appropriate.

PR #486, opened 2021-08-18
Audit logging for changes

rabstejnek:
Refactored base views (`BaseDelete`,`BaseCreate`,`BaseUpdate`,`BaseCreateWithFormset`,`BaseUpdateWithFormset`) and added logging where it was missing.

- Base views now log more relevant information
  - All base views now log the associated object
- Refactored base views to be more easily subclassed
  - Updated current subclassed views to make use of new logging
- Edits to Log model
  - added optional `content` - JSON field for storing associated data
  - added optional `content_object` - generic relation for associated object
  - removed `last_updated`; model is read-only


shapiromatron (changes requested):
After a deeper dive into this PR and investigation on what's being captured with some models, I think the original task was scoped unrealistically.

I think we should keep the majority of this PR, so that we're tracking all of the individual changes for creates, updates, and deletes, with the content object, assessment, and user. However, we should remove the diffing capabilities. That seems to be handled, or at least the data for deep diffs would be available in [django-reversion](https://django-reversion.readthedocs.io/en/stable/index.html).   An example of the data generated:

```python
from hawc.apps.assessment.models import Assessment
from reversion.models import Version

a = Assessment.objects.get(id=100000061)
versions = Version.objects.get_for_object(a)
for ver in versions:
    print(ver.id, ver.field_dict)
>>>101177338 {'id': 100000061, 'name': 'Chloroform', 'year': 2020, 'version': 'clone from 2018 version', 'cas': '', 'assessment_objective': '', 'editable': True, 'public': False, 'hide_from_public_page': False, 'is_public_training_data': False, 'enable_literature_review': True, 'enable_project_management': True, 'enable_data_extraction': True, 'enable_risk_of_bias': True, 'enable_bmd': True, 'enable_summary_text': True, 'conflicts_of_interest': '', 'funding_source': '<p>EPA</p>', 'noel_name': 1, 'rob_name': 1, 'vocabulary': 1, 'modify_uncontrolled_vocabulary': True, 'created': datetime.datetime(2018, 10, 18, 15, 42, 1, 865000, tzinfo=<UTC>), 'last_updated': datetime.datetime(2021, 8, 31, 12, 40, 45, 608000, tzinfo=<UTC>), 'dtxsids': ['DTXSID0020523'], 'project_manager': [641, 638, 100500166, 565, 654, 642, 100000024, 100500255, 119, 100500200, 568, 81, 100000069, 100500122, 598, 100500029, 100500099, 636, 100500006, 100500101, 716, 100500000, 569, 674, 100500027, 100500109, 656, 595, 100000068, 680, 564, 682, 659, 676, 100000025, 678, 20, 100500088, 100500104], 'team_members': [638, 100000000, 100500106, 100500030, 100000058, 658, 637, 575, 100500111, 569, 656, 100500229, 100500258, 680, 659, 678, 100500104], 'reviewers': []}
>>> ...
```

Thus, this PR will log all changes that are made, but not the individual diffs. We can investigate displaying diffs for a subset of model types, perhaps assessments, or other key diffs.

To complete this PR -could we remove the diffing capabilities?

shapiromatron (approved):
Looks great! I made a few more changes.

- updated logging text pattern, eg., `Created study.study #101061556: "Yang L, Quan S, and Abraham NG 1999"`
- removed the `save_and_log` and just moved that code into the `form_valid`
- added a shared `create_object_log` function
- made the form_valid atomic instead of the whole post; this is because these locks can be costly so we only want to use them when the data are valid

rabstejnek:
@shapiromatron the changes look good!

PR #490, opened 2021-08-24
standardize unique for assessment error message

shapiromatron:
Users were previously confused with a global error message instead of one applied to a specific field. This PR does two things:

- In cases where a global model error was thrown (lit.Search title and slug, study.Study short_citation) change from a global form error to a field specific error
- standardize unique for assessment error message in hawc

Example for field specific errors:

Screen Shot 2021-08-24 at 9 44 39 AM
Screen Shot 2021-08-24 at 9 45 51 AM


rabstejnek (approved):
Looks good!

PR #493, opened 2021-09-01
Add async task expirations for recurring tasks

rabstejnek:
Added expirations to celery beat, so that scheduled tasks don't queue up multiple times if a service is down.

Inspired by [Caktus](https://www.caktusgroup.com/) blogpost: https://www.caktusgroup.com/blog/2021/08/11/using-celery-scheduling-tasks/

### DIFF
--- a/hawc/main/celery.py
+++ b/hawc/main/celery.py
@@ -24,39 +24,39 @@ app.conf.beat_schedule = {
     "worker-healthcheck": {
         "task": "hawc.apps.common.tasks.worker_healthcheck",
         "schedule": timedelta(minutes=5),
-        "options": {"expires": timedelta(minutes=5)},
+        "options": {"expires": timedelta(minutes=5).total_seconds()},
     },
     "lit-schedule_topic_model_reruns-10-min": {
         "task": "hawc.apps.lit.tasks.schedule_topic_model_reruns",
         "schedule": timedelta(minutes=10),
-        "options": {"expires": timedelta(minutes=10)},
+        "options": {"expires": timedelta(minutes=10).total_seconds()},
     },
     "lit-update_pubmed_content-1-day": {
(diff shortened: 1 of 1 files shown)


hawc/main/celery.py
```diff
@@ -24,32 +24,39 @@ def debug_task(self):
     "worker-healthcheck": {
         "task": "hawc.apps.common.tasks.worker_healthcheck",
         "schedule": timedelta(minutes=5),
+        "options": {"expires": timedelta(minutes=5)},
```
shapiromatron, line 27: does expires accept a timedelta?  I'd assume it does, just didn't see it in the docs?

shapiromatron (approved):
Looks great; I saw you originally used timedelta/2, but I prefer the full time as you changed it here; tasks should be idempotent so worst case they'll run twice, but that's better than having them all stack...

hawc/main/celery.py
rabstejnek: I would assume it does too, but I'll double check and comment it here.

rabstejnek: The [options are the same as in `apply_async`](https://docs.celeryproject.org/en/stable/userguide/periodic-tasks.html#available-fields), which has an [example of `expires` using `timedelta`](https://docs.celeryproject.org/en/stable/userguide/calling.html#expiration).

rabstejnek: I stand corrected, it looks like it may only accept a float/datetime [[1](https://github.com/celery/celery/blob/9e435228cb106588f408ae71b9d703ff81a80531/celery/app/task.py#L441),[2](https://github.com/celery/celery/blob/9e435228cb106588f408ae71b9d703ff81a80531/celery/app/amqp.py#L301)]. I will call `total_seconds()` on each of these `timedelta`s to convert it to seconds just to be safe.

shapiromatron (approved):

PR #495, opened 2021-09-07
Update rob assignment form

shapiromatron:
The prior read and update views for assigning reviewers for study evaluation were updated for the following reasons:

1. The layout was problematic when there were varying number of individual reviewers for a study
2. A user couldn't easily add multiple reviewers for a single study for calibration without altering the required reviews
3. The update form was incredibly slow and you could only bulk-submit large changes
4. Importantly, the behavior wasn't intuitive. In many cases users thought they were reassigning an existing review to a new reviewer to complete, but in fact, it was marking the exiting review as inactive and creating a new review. This means that users thought they were loosing data, but the data were still present and hidden by the user-interface.

The user-interface has been rewritten, with the following changes:

- Show three columns; one for study, one for individual reviews, and one for final reviews
- When updating assignments, allow user to create new reviews or update existing. They have the option to change the review author OR toggle a review as active/inactive.
- Footnotes were added to the table to describe the new concept of active/inactive
- Dramatically improved performance, since all rendering of code is completed in the frontend rather than using django forms and formsets on the backend.

One quibble about the current implementation is the duplication of serialization between the riskofbias views where data are hydrated into the form, and the responses after a successful POST/PATCH. This syncing may cause issues in the future, but we do have unit tests in place for the serializers, and integration tests for the frontend views.

**Read-only view:**
image

**Create/update view:**
image


rabstejnek (approved):
Looks good!

PR #497, opened 2021-09-08
assessment log change views

shapiromatron:
#486 implemented a backend change on all form-based views to log assessment, user, item, and change description. This PR builds a number of views to allow project managers and other team members to view logs associated with data in hawc.

## Assessment log list
Screen Shot 2021-09-28 at 10 47 09 PM

## Log object list
Screen Shot 2021-09-28 at 10 48 01 PM

## Log detail
Screen Shot 2021-09-28 at 10 49 09 PM

## Content type info
Screen Shot 2021-09-28 at 10 50 26 PM


hawc/apps/assessment/models.py
```diff
@@ -919,9 +919,30 @@ class Log(models.Model):
...
+            return str(self.content_object) + " Log"
+        if self.assessment is not None:
+            return str(self.assessment) + " Log"
+        return "Custom Log"
+
+    def get_absolute_url(self):
+        return reverse("assessment:log_detail", args=(self.id,))
+
     def get_api_url(self):
+        # TODO delete api stuff?
```
rabstejnek, line 933: @shapiromatron This method isn't called anywhere, so I can see it safely being deleted. Do you think we should go a step further and remove the log API altogether?

shapiromatron: Let's add a jira card to revisit in Q1 2022, sound ok?

rabstejnek:
Here are some screenshots of the changes made to log views.
## Assessment logs
### Without any filters
image
### After object filter is applied
image
## Log detail
image
## Content object logs
image

hawc/apps/assessment/models.py
rabstejnek: Done!

rabstejnek:
@shapiromatron ready for review!

shapiromatron:
This is coming along great. I played around a fair amount today, and had some comments. Let me know if you agree with them, and if so, would be willing to work through them:

1. `assessment_log_list`:
- [ ] I added `on`, `before`, and `after`, but can't think of an easy way to filter on this
- [ ] Could we render a {{form}} in the django template with these?
    - perhaps we can add a django form at the top? it should show the user and date fields, do not show content_type or content_object
- [ ] keep the filter icon on the Item (change to black), but remove from User after form

2. `log_detail` page
- [ ] need to use get_template_name in case assessment is None
- [ ] for breadcrumbs, if we have an assessment, make detail page just a read-only crumb below the assessment_log_list list page
- [ ] add the `content_object.get_absolute_url()` as a detail view

3. `log_object_list` page:
- [ ] for breadcrumbs, if we have an assessment, make detail page just a read-only crumb below the assessment_log_list list page
- [ ] add the `content_object.get_absolute_url()` as a detail view
- [ ] if item is deleted; chagne the title of the page to log.get_generic_object_name

rabstejnek:
@shapiromatron your comments make sense; I'll work on implementing them!

rabstejnek:
> This is coming along great @rabstejnek . I played around a fair amount today, and had some comments. Let me know if you agree with them, and if so, would be willing to work through them:
> 
> 1. `assessment_log_list`:
> 
> * [x]  I added `on`, `before`, and `after`, but can't think of an easy way to filter on this in the current UI, therefore...
> * [x]  Could we render a {{form}} in the django template with these?
>   
>   * perhaps we can add a django form at the top? it should show the user and date fields, do not show content_type or content_object
> * [x]  keep the filter icon on the Item (change to black), but remove from User after form
> 
> 1. `log_detail` page
> 
> * [x]  need to use get_template_name in case assessment is None
> * [x]  for breadcrumbs, if we have an assessment, make detail page just a read-only crumb below the assessment_log_list list page
> * [x]  add the `content_object.get_absolute_url()` as a detail view
> 
> 1. `log_object_list` page:
> 
> * [x]  for breadcrumbs, if we have an assessment, make detail page just a read-only crumb below the assessment_log_list list page
> * [x]  add the `content_object.get_absolute_url()` as a detail view
> * [x]  if item is deleted; chagne the title of the page to log.get_generic_object_name

Requested changes have been made! Just a few points to make:
- I kept the filter icon on the assessment log page the same color; after making it black, I personally thought it looked better the original way.
- on log detail and log object list, I don't use `get_template_name`; I instead use the `extend` tag and `yesno` filter to decide whether to use `assessment-rooted` or `crumbless`
- on log detail and log object list, example breadcrumbs are `Home / PFAS Logs / Log` and `Home / PFAS Logs / Logs`, respectively
@shapiromatron let me know if these changes appropriately address all the comments you had, and if there are any other changes to make!

rabstejnek:
@shapiromatron How do you want to handle logs without an associated assessment? My current implementation shows a non-`assessment-rooted` view of such a log if the user is staff:
image
Do you think we should just show 404 instead, or do you like the availability of a view for staff?

shapiromatron:
> @shapiromatron How do you want to handle logs without an associated assessment? My current implementation shows a non-`assessment-rooted` view of such a log if the user is staff:
> Do you think we should just show 404 instead, or do you like the availability of a view for staff?

That looks good to me!

shapiromatron:
@rabstejnek ok, I change the crumbs a little bit, and changed the form too since it was confusing that some things could be edited on the form, but not all.  Take a look at let me know what you think! Also added some permissions and template tests. Hopefully we're getting ready to merge!

rabstejnek (approved):
Changes look good!

PR #500, opened 2021-09-09
Fix Actions dropdown button on endpoint detail page

rabstejnek:
The Actions dropdown button now only renders when there are items to display.

Logic is confusion: many possible states:
1. if not in detail mode, no action button
1. if user cannot edit and no bmd model, no action button
1. if user cannot edit and bmd model, action button w/ session 
1. if user can edit, show action with endpoint editing, and ....
    1. if no bmd and bmd is possible, add create bmd buton
    1. if bmd user can edit and bmd, add views session button
    1. if bmd not enabled, no bmd model section under actions

### DIFF
--- a/hawc/apps/animal/templates/animal/endpoint_detail.html
+++ b/hawc/apps/animal/templates/animal/endpoint_detail.html
@@ -2,7 +2,6 @@
 
 {% block content %}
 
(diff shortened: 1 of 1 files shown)


shapiromatron (approved):
PR #500! 

🎆

PR #502, opened 2021-09-10
Summary table updates

rabstejnek:
Made several updates to summary tables.

- Moved the "Save/Update" and "Cancel" buttons out of the "Overall" tab on the summary table forms, allowing them to be used regardless of which form tab is active.
- Added http://public.tableau.com to `valid_netloc_endings`.
- Updated JS method `urlify`:
  - It now takes `maxLength` and `suffixLength` parameters.
  - If the slugified version of the passed in string is > `maxLength`, then it is truncated and a random string is appended to the slug.  This random suffix is to make it more likely that the slug will be unique in the case that the unique portion of the slug was truncated off.

Example urlify:
```javascript
window.app.HAWCUtils.urlify('this is a short string')
'this-is-a-short-string'

window.app.HAWCUtils.urlify('this is a long string this is a long string this is a long string')
'this-is-a-long-string-this-is-a-long-string-t-ac3f'
'this-is-a-long-string-this-is-a-long-string-t-6206'
'this-is-a-long-string-this-is-a-long-string-t-9edd'
```

_Other changes:_
- Made `FormActions` for reusable "Submit"/"Cancel" buttons and implemented them where appropriate.
- Fixed rendering of invalid exploratory heatmaps, and added Actions so that users can fix them.
  - To reproduce test case, create a new exploratory heatmap without selecting a dataset.


frontend/summary/summaryTable/editing/Root.js
```diff
@@ -10,27 +10,38 @@ import DjangoForm from "./DjangoForm";
...
+                        <DjangoForm />
+                    </TabPanel>
+                    <TabPanel>
+                        <Form store={tableStore} />
+                    </TabPanel>
+                    <TabPanel>
+                        <Table store={tableStore} forceReadOnly={true} />
+                    </TabPanel>
+                </Tabs>
+                <div className="form-actions">
```
shapiromatron, line 36: can we make this div a reusable react component?  Maybe we can call it a `FormAction`?

It looks like there are currently 3 that already exist `BaseFormVisualReact, App, and DjangoForm`

The default for saveBtnText can be "Save"

hawc/apps/assessment/templates/assessment/effecttag_form.html
```diff
@@ -16,7 +16,7 @@
   <script type="text/javascript">
     $(document).ready(function(){
       $('#id_name').on('keyup', function(){
-        $('#id_slug').val(window.app.HAWCUtils.urlify($(this).val(), 50));
+        $('#id_slug').val(window.app.HAWCUtils.urlify($(this).val(),128));
```
shapiromatron, line 19: should the default length be 50? https://docs.djangoproject.com/en/3.2/ref/models/fields/#slugfield

frontend/shared/utils/HAWCUtils.js
```diff
@@ -203,8 +203,22 @@ class HAWCUtils {
         return false;
     }
 
-    static urlify(str) {
-        return slugify(str, {remove: /[^\w\s-_]/g});
+    static randomString(length) {
```
shapiromatron, line 206: very similar code already exists here w/ a different implementation https://github.com/shapiromatron/hawc/blob/e54a8bee877e2cd63d03109106c3c108be3bf197/frontend/shared/utils/helpers.js#L246-L250

frontend/shared/utils/HAWCUtils.js
```diff
@@ -203,8 +203,22 @@ class HAWCUtils {
...
+        const chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+        return Array(length)
+            .fill()
+            .map(() => chars.charAt(Math.floor(Math.random() * chars.length)))
+            .join("");
+    }
+
+    static urlify(str, maxLength = 50, suffixLength = 4) {
```
shapiromatron, line 214: cool, i forget about how default params are allowed now...

shapiromatron (changes requested):
Looking good but a few more changes

hawc/apps/assessment/templates/assessment/effecttag_form.html
shapiromatron: doh; i think you did do that, and i'm looking at a different case. my bad.

frontend/shared/utils/HAWCUtils.js
rabstejnek: A quick benchmark shows the newer implementation being a bit faster (I imagine since its not relying on a regex):
image
Since it'll also be helpful if this function can return a string variable in length, I'm going to replace this old method with the new one, and import it in to `HAWCUtils` for use.

frontend/summary/summaryTable/editing/Root.js
rabstejnek: I found another case, `frontend/mgmt/TaskTable/containers/Root.js`, where I was able to replace the form actions. I also replaced all the other instances, except for `BaseFormVisualReact`; this belongs to the aggregate endpoints visual form, which I was not able to test due to existing errors.

rabstejnek:
@shapiromatron I sneaked a few other small fixes in this PR that I came across during testing. I've documented them in my initial comment.

rabstejnek:
@shapiromatron All of the requested changes are done, except for adding the `FormActions` component to `BaseFormVisualReact`; this belongs to the endpoint aggregation form, which I was not able to test because there are existing errors on submit. After these errors are addressed, it should be as easy as adding `<FormActions isForm={true} submitText="Save" cancelURL={this.config.cancel_url} />`. This is the one instance where the form actions actually take place within a form, which is why the `isForm` prop is present, and why we don't have to pass it a `handleSubmit`.

Two other things I've noticed:
- [This](https://github.com/shapiromatron/hawc/blob/summary-table-updates/frontend/summary/summaryForms/BaseVisualFormReact.js#L125) takes a `ref`, but from what I can see one is never created. It may be unnecessary.
- [This](https://github.com/shapiromatron/hawc/blob/summary-table-updates/frontend/summary/summaryForms/BaseVisualForm.js) seems to be a non-React version of this form, and from what I can tell its not used anywhere. We may be able to remove this file entirely.

shapiromatron:
> @shapiromatron All of the requested changes are done, except for adding the `FormActions` component to `BaseFormVisualReact`; this belongs to the endpoint aggregation form, which I was not able to test because there are existing errors on submit. After these errors are addressed, it should be as easy as adding `<FormActions isForm={true} submitText="Save" cancelURL={this.config.cancel_url} />`. This is the one instance where the form actions actually take place within a form, which is why the `isForm` prop is present, and why we don't have to pass it a `handleSubmit`.
> 
> Two other things I've noticed:
> 
> * [This](https://github.com/shapiromatron/hawc/blob/summary-table-updates/frontend/summary/summaryForms/BaseVisualFormReact.js#L125) takes a `ref`, but from what I can see one is never created. It may be unnecessary.
> * [This](https://github.com/shapiromatron/hawc/blob/summary-table-updates/frontend/summary/summaryForms/BaseVisualForm.js) seems to be a non-React version of this form, and from what I can tell its not used anywhere. We may be able to remove this file entirely.

Working on this now. To your second point; it looks like `summaryForms/BaseVisualForm.js` is used in multiple places
Screen Shot 2021-09-20 at 6 43 33 PM
?

frontend/shared/utils/helpers.js
```diff
@@ -6,6 +6,7 @@ import _ from "lodash";
 import moment from "moment";
 
 const stopwords = new Set("the is at which of on".split(" ")),
+    hexChars = "abcdef0123456789",
```
shapiromatron, line 9: switched to hex characters instead of all for readability

frontend/shared/utils/HAWCUtils.js
```diff
@@ -203,8 +204,14 @@ class HAWCUtils {
...
 
-    static urlify(str) {
-        return slugify(str, {remove: /[^\w\s-_]/g});
+    static urlify(str, maxLength = 50, suffixLength = 4) {
+        // transforms a given string into a url friendly slug
+        // if the slug is greater than maxLength, then it is truncated and appended
+        // with a random string (to ensure uniqueness if uniqueness is truncated)
+        var slug = slugify(str, {remove: /[^\w\s-_]/g});
+        return slug.length > maxLength
+            ? slug.slice(0, maxLength - suffixLength - 1) + "-" + h.randomString(suffixLength)
```
shapiromatron, line 213: added a hyphen between the slug and the entropy

frontend/summary/summary/ExploreHeatmap.js
```diff
@@ -178,7 +177,7 @@ class ExploreHeatmap extends BaseVisual {
             caption = new SmartTagContainer(captionDiv),
             $plotDiv = $("<div>"),
             callback = resp => {
-                if (resp.dataset) {
+                if (resp.dataset || resp.error) {
```
shapiromatron, line 180: i changed how this was implemented; and just  carried w/ the standard path even if an error was logged, just because it looks like they were generally the same so we didnt have as much duplicate code

shapiromatron (approved):
Just like #506; made some minor changes. If you're happy with them, please feel free to merge!

rabstejnek:
> > @shapiromatron All of the requested changes are done, except for adding the `FormActions` component to `BaseFormVisualReact`; this belongs to the endpoint aggregation form, which I was not able to test because there are existing errors on submit. After these errors are addressed, it should be as easy as adding `<FormActions isForm={true} submitText="Save" cancelURL={this.config.cancel_url} />`. This is the one instance where the form actions actually take place within a form, which is why the `isForm` prop is present, and why we don't have to pass it a `handleSubmit`.
> > Two other things I've noticed:
> > 
> > * [This](https://github.com/shapiromatron/hawc/blob/summary-table-updates/frontend/summary/summaryForms/BaseVisualFormReact.js#L125) takes a `ref`, but from what I can see one is never created. It may be unnecessary.
> > * [This](https://github.com/shapiromatron/hawc/blob/summary-table-updates/frontend/summary/summaryForms/BaseVisualForm.js) seems to be a non-React version of this form, and from what I can tell its not used anywhere. We may be able to remove this file entirely.
> 
> Working on this now. To your second point; it looks like `summaryForms/BaseVisualForm.js` is used in multiple places?
> Screen Shot 2021-09-20 at 6 43 33 PM
> 
> Update - completed. I also fixed the `EndpointAggregationFormReact.js` and you're now able to update and create.
> 
> I think I found the file you meant and deleted it - `EndpointAggregationForm.js`. I also go the update form working again. Wow, this stuff is terribly complex. We should investigate the bets way to remove complexity here; either use the exploratory heatmap pattern, or switch over to API-based. I wonder if going API-based is better in the long run since we know better how to do it now that we've built `SummaryTableViewset` in the last year.

Sorry, I had meant `EndpointAggregationForm.js`!

rabstejnek:
Looks good!

PR #503, opened 2021-09-10
make accepting terms of  service optional and based on deployment

shapiromatron:
Previously, when creating an account, you were required to accept the terms of service. Now, this is configurable based on the deployment. The default is the current behavior, where accepting the terms of service is required.

If the env variable `HAWC_ACCEPT_LICENSE_REQUIRED` is false, then then django setting `ACCEPT_LICENSE_REQUIRED` is also false, and accepting a license upon account creation or login is no longer required.

To test, update your `local.py` settings and change ACCEPT_LICENSE_REQUIRED to True or False.


rabstejnek (approved):
Looks good!

PR #505, opened 2021-09-14
refactor Endpoint JS summary tables and modernize codebase

shapiromatron:
This PR fixes a number of assorted bugs around frontend animal.Endpoint javascript code. Instead of a simple patch to fix the bugs, we broadened the fix to a number of refactorings which should modernize and hopefully simplify the codebase

Bugs fixed:
- gracefully handle cases where the # doses and # endpoint groups don't align. This should never occur, but does, based on prior bugs in hawc that have been fixed #424.

Refactorings:
- abstracted all doseUnits modification outside of Endpoint into a new ActiveDose class with new API
- rewrote AnimalGroupEndpoint table as a react module, instead of a mixture of legacy approaches
- generalized the `eyeEndpointModal` class and behavior using css instead of custom jQuery
- moved helper methods to render some views of an Endpoint out of the Endpoint model

Additional changes:
- with AnimalGroup Endpoint table, added a "preview" button to show an endpoint modal popup
- updated admin for animal.DosingRegime and animal.Endpoint to better mirror hawc data entry forms
- (deployment) - shorten git SGA from 12 to 8 characters; 16^8 = 4.2 billion
- (deployment) - don't restart the app-db `sync` container


frontend/animal/AnimalGroupTable/AnimalGroupTable.js
```diff
@@ -0,0 +1,181 @@
...
+            </tfoot>
+        );
+    }
+    render() {
+        const {endpointsDr} = this.props.store;
+        if (endpointsDr.length === 0) {
+            return <p>No endpoints are available with dose-response data.</p>;
+        }
+        return (
+            <table className="table table-condensed">
```
rabstejnek, line 169: Should be `table-sm`

frontend/animal/AnimalGroupTable/AnimalGroupTable.js
```diff
@@ -0,0 +1,181 @@
...
+        let dr_control, html, fn, response;
+        return (
+            <tr key={`e-${id}`} className="previewModalParent">
+                <td>
+                    <i
+                        className="float-right fa fa-fixed fa-eye previewModalIcon"
+                        onClick={() => endpoint.displayAsModal({complete: true})}></i>
+                    <a href={url}>{name}</a>
+                </td>
+                <td>{organ}</td>
```
rabstejnek, line 32: Right now if there's no `organ` it just leaves the cell empty, which is stylistically different than `nRow` which fills in empty cells with `-`. Not a big deal, but thought I'd point it out anyway.
image

rabstejnek (changes requested):
Looks good! Just one bootstrap class to fix, and another comment to address if you think it's worth addressing.

frontend/animal/AnimalGroupTable/AnimalGroupTable.js
shapiromatron: ah, good find. i restored the "-" functionality.

rabstejnek (approved):

PR #508, opened 2021-09-24
add goto page number selector

shapiromatron:
Previously, our default paginator allowed a user to go left or right one page at a time. This may be problematic for the user as well as unnecessary server load if there are many pages and you want to go to the end, or the middle. 

Therefore, we add a goto page where a user can select any page and they are redirected there

Old:
image

New:
image

### DIFF
--- a/hawc/templates/includes/paginator.html
+++ b/hawc/templates/includes/paginator.html
@@ -26,5 +26,16 @@
         <a class="page-link" href="#">&gt;&gt;</a>
     </li>
     {% endif %}
+
+    <li class="ml-3 page-item">
+        <label class="page-link font-weight-normal text-muted" for="quickPageSelector">Go to page:&nbsp;</label>
+    </li>
+    <li class="page-item from-group">
+        <select id="quickPageSelector" class="form-control" onchange="location=this.value;">
+            {% for n in page_obj.paginator.page_range %}
+            <option {% ifequal page_obj.number n %}selected{% endifequal %} value="?{% url_replace request 'page' n %}">{{ n }}</option>
+            {% endfor %}
+        </select>
+    </li>
 </ul>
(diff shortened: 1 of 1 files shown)


hawc/templates/includes/paginator.html
```diff
@@ -26,5 +26,16 @@
         <a class="page-link" href="#">&gt;&gt;</a>
     </li>
     {% endif %}
+
+    <li class="ml-3 page-item">
```
rabstejnek, line 30: If we add class `disabled` we can get rid of the hover styles over the label, so that users don't mistake the label as an interactive element.

hawc/templates/includes/paginator.html
```diff
@@ -26,5 +26,16 @@
...
+        <label class="page-link font-weight-normal text-muted" for="quickPageSelector">Go to page:&nbsp;</label>
+    </li>
+    <li class="page-item from-group">
```
rabstejnek, line 33: Misspelled `form-group`

rabstejnek (changes requested):
Looks good! Just one misspelling to fix, and one suggestion.

hawc/templates/includes/paginator.html
shapiromatron: fixed!

hawc/templates/includes/paginator.html
shapiromatron: Yes, this was intentional. I wanted people to know that that it is interactive; if they click the goto label, it focuses the page select. This is consistent with how other labels in forms work. Ok to leave as is?

rabstejnek: Sure, makes sense!

rabstejnek (approved):

PR #510, opened 2021-09-29
literature tagging user-interface improvements

shapiromatron:
We've had a number of requests to improve the literature module user-interface to make it easier to tag. This PR is designed to reduce the amount of scrolling up and down, as well as making reference order deterministic for all references.

1. All reference list views (including tagging and reading) are sorted. They are sorted by year descending, then by author name descending: eg., `2019 Abc, 2019 Xyz, 2018 Abc, 2018 Xyz` ...
image

2. Reference tag list container heights are customizable:
image

3. Tagtree is scrollable and resizable on all views
Screen Shot 2021-09-29 at 9 53 01 PM

Overall example after updates:
Screen Shot 2021-09-29 at 10 03 27 PM


frontend/lit/components/TagTree.js
```diff
@@ -15,28 +15,34 @@ class TagNode extends Component {
...
+                        {hasChildren ? (
+                            <button
+                                className="btn btn-sm pull-right p-0 px-2"
+                                onClick={toggleExpander}>
+                                <i className={`fa ${expanderIcon}`}></i>
+                            </button>
+                        ) : null}
+                    </div>
+                    <div style={{flex: 1}}>
+                        <span onClick={() => handleOnClick(tag)}>
```
rabstejnek, line 40: Move `onClick` to the parent div; as of now you have to actually click the text to add a tag, while before you could click the block element.

rabstejnek (changes requested):
Looks good! Just one requested change.

rabstejnek (approved):

PR #512, opened 2021-09-29
enable editing DOI on reference form

shapiromatron:
Update reference form `/lit/reference/:id/update/` so the DOI can be edited.  Add a permissive regex expression to check for doi-like values that should resolve if they are valid:

Screen Shot 2021-09-29 at 10 59 39 PM

The doi is entered in a form like `10.1000/xyz123`, as recommended from https://www.doi.org/

---

Also refactored the add hero/pubmed ids to generalize, since the same code was copied 3x times


rabstejnek (approved):
Looks good!

PR #513, opened 2021-09-30
add tag tree fetch/clone/update to API & client

DookTibs:
@shapiromatron one thing to call out that is nagging me; the format returned by all of these is the native Treebeard format (so name/slug are wrapped in a "data" dictionary), whereas the format accepted by update_tag_tree just has name/slug at the top level, no "data" nesting those values (the API transforms it into the format Treebeard wants). I did it this way b/c it felt simpler for the client, but it is admittedly a little inconsistent and potentially confusing. If you'd prefer I just required API consumers to wrap it themselves, I can redo it that way easy enough.

-----

Adds three new API endpoints:
* /lit/api/assessment/<assessment_id>/get_tag_tree - fetches the tag tree for an assessment as a nested structure
* /lit/api/assessment/<target_assessment_id>/clone_tag_tree_from_assessment?assessment_id=<source_assessment_id> - clones tag
  tree from one assessment to another
* /lit/api/assessment/<assessment_id>/update_tag_tree - takes POSTed JSON data and replaces the existing tag tree

Also updates the Python HAWC client with methods matching the three new endpoints.

Addresses https://trello.com/c/g3rUWQMG/168-api-to-bulk-replace-tags

### DIFF
--- a/hawc/apps/common/models.py
+++ b/hawc/apps/common/models.py
@@ -340,6 +340,119 @@ class AssessmentRootMixin:
         return depth
 
     @classmethod
+    def validate_and_transform_tag_tree(
+        cls, tags_at_level: List[object], issues: List[str], breadcrumbs: List[int] = []
+    ):
+        """
+        Goes through the tags_at_level and for each tag:
+            * verifies that it has a name
+            * if a slug is present, checks that it's valid. If it's missing, generates one.
+            * deletes any keys other than name/slug/children
+
+        Issues found are appended in readable fashion to the issues list; an empty
+        issues list indicates that no problems were found. Note that things like a missing slug
+        or extraneous keys are NOT reported as issues; they are just corrected in the underlying
+        tags data structure.
+
+        This method modifies the supplied tag_tree by relocating the name/data attributes inside a "data" node.
+
+        Args:
+            tags_at_level (List[object]): any level of the tag tree
+            issues (List[str]): storage for reporting any issues encountered while validating the tree
+            breadcrumbs (List[int]): array of indices indicating where in the tree structure the code is at
+        """
+
+        def prettify_indices(list_of_indices: List[int]):
+            """
+            Given a list containing just numeric indices, returns a dot-delimited version with the
+            indices wrapped in square brackets.
+
+            For example,
+            [8, 6, 7] -> "[8].[6].[7]"
+
+            Args:
+                list_of_indices (List[int]): list of indices
+
+            Returns:
+                str: formatted string
+            """
+            return ".".join([f"[{x}]" for x in list_of_indices])
+
+        valid_keys = ["name", "slug", "children"]
+        idx = 0
+        for tag in tags_at_level:
+            tag_name = None
+            slug = None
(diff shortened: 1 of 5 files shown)


hawc/apps/common/models.py
```diff
@@ -340,6 +340,119 @@ def get_maximum_depth(cls, assessment_id):
...
+                del tag["name"]
+            if "slug" in tag:
+                del tag["slug"]
+            tag["data"] = {"name": tag_name, "slug": slug}
+
+            idx += 1
+
+    @classmethod
+    @transaction.atomic
+    def replace_tag_tree_in_assessment(cls, assessment, tag_tree: List[object]):
```
shapiromatron, line 432: use `Dict` or `Any` instead of object; use `assessment_id: int` as well; rename to `replace_tree`

hawc/apps/common/models.py
```diff
@@ -340,6 +340,119 @@ def get_maximum_depth(cls, assessment_id):
...
+        complete_tree = [{"data": {"name": root_name, "slug": root_name}, "children": tag_tree}]
+
+        root_node.delete()
+        cls.load_bulk(complete_tree, parent=None, keep_ids=False)
+        cls.clear_cache(assessment.id)
+        return cls.get_all_tags(assessment.id, json_encode=False)
+
+    @classmethod
+    @transaction.atomic
     def copy_tags(cls, copy_to_assessment, copy_from_assessment) -> Dict[int, int]:
```
shapiromatron, line 456: could you rewrite `copy_tags` to use this same `replace_tag_tree_in_assessment`?

shapiromatron (changes requested):
This works, but it's a lot of lower level custom code that's not leveraging some of the existing tools we have in hawc; rest framework serializers, jsonschema, etc. Plus, we can restructure this to make it more restful since we really have a get/post pattern for this PR.

See the jupyter notebook in trello as an example proof of concept for how we can use serializers and jsonschema

shapiromatron:
> @shapiromatron one thing to call out that is nagging me; the format returned by all of these is the native Treebeard format (so name/slug are wrapped in a "data" dictionary), whereas the format accepted by update_tag_tree just has name/slug at the top level, no "data" nesting those values (the API transforms it into the format Treebeard wants). 

Yeah, that makes sense. I'm fine with either way of requiring a single root node that you ignore, or a list of children. But I agree, we should be consistent. It should be easy to do with the serializer approach

DookTibs:
@shapiromatron updated addressing comments - now a single tagtree API endpoint; client still has a clone method that just hits it twice with source/target assessment id's. Reworked to use jsonschema for validation. Input/output formats are now consistent with name/slug wrapped in data. I ended up not actually even needing the TreeSerializer with this approach so did not add that but otherwise I think all your comments are addressed. Let me know if this needs more changes though.

shapiromatron:
@DookTibs this is looking great; i'm going to take on this branch and make a few mods, will loop you in after I'm done.  Thanks! Ok to move to the next task!

PR #519, opened 2021-10-08
Bulk tag references - frontend

rabstejnek:
Add a user interface for users to to upload and bulk-apply tags to references in HAWC. Must be team-member or higher. This is a continuation of #514.

The UI consists of three steps:

1. Upload Excel spreadsheet
2. Map rows in spreadsheet to references in HAWC via HAWC reference ids, PubMed ids, and HERO ids
3. After mapping, assign a row and select a unique value, and which tag you'd like to associate with this tag. Select to either append or remove tags to these references. It's expected that after setting up steps 1 and 2 above, step 3 will be repeated multiple times, assigning different values to different tags.

## Ideal case - all matches

This case describes the situation where every single row in the spreadsheet matches a reference in HAWC.

Screen Shot 2021-10-12 at 11 01 51 PM
Screen Shot 2021-10-12 at 11 11 25 PM
Screen Shot 2021-10-12 at 11 11 39 PM

## Alternative case - some matches

This case describes the situation where some, but not all rows in the spreadsheet match a reference in HAWC.

Screen Shot 2021-10-12 at 11 05 35 PM
Screen Shot 2021-10-12 at 11 05 52 PM

### DIFF
--- a/tests/frontend/integration/tests/literature.py
+++ b/tests/frontend/integration/tests/literature.py
@@ -38,4 +38,8 @@ def literature(driver, root_url):
     h.wait_until(h.Text("1 references found.").exists)
     assert len(driver.find_elements_by_css_selector(".referenceDetail")) == 1
 
+    # /lit/assessment/:id/tag/bulk/
+    h.go_to(root_url + "/lit/assessment/1/tag/bulk/")
+    h.wait_until(h.Text("Select an Excel file (.xlsx) to load and process.").exists)
+
     shared.logout()


shapiromatron:
@rabstejnek ok, turning this back over to you. How does it look?

shapiromatron:
@DookTibs requesting review to see if this would work with common workflows. Existing exports from external tools may need to be modified slightly, but hopefully the burden is relatively minor

rabstejnek:
> @rabstejnek ok, turning this back over to you. How does it look?

Tested it out, looks good!

DookTibs:
@rabstejnek @shapiromatron looks really useful - I grabbed this on a local instance and Kevin and I ran through some test scenarios and it worked like a charm. Would it be hard to add csv support in addition to xlsx?

(just to be clear and correct me if I'm wrong; I did not look at the code other than a quick peek for curiosity's sake. I assumed I was tagged for review in more of a 'would this be useful for ICF' kind of sense.)

shapiromatron:
@DookTibs would you mind approving :)

PR #527, opened 2021-11-02
Heatmap fixes

rabstejnek:
1. Fixed bug where sorted, filtered rows/columns would fail to render.
(to reproduce, sort by acute/chronic and species on `/summary/visual/assessment/100000061/Immune-Effects-Animal/`)

2. Visuals using VisualToolbar now initially render by whichever is smallest, native size or fitted size. Also improved on implementation.

|Before|After|
|-|-|
|image|image|

3. Autosizing cells now takes into account tick rotation and word wrap.

|Before|After|
|-|-|
|image|image|

4. Custom rotate on y axis ticks has been fixed.

|Before|After|
|-|-|
|image|image|

5. Refactored code to combine similar `build_bottom_axis` and `build_top_axis` to `build_x_axis`.


shapiromatron (approved):
Amazing job! Looks great.

Really well done w/ the layout to find max text dimensions and rotation logic. It's must clearer than the previous version.

PR #529, opened 2021-11-04
use django form for rasterizer download

shapiromatron:
Minor updates to improve data validation to handle errors during validation instead of processing.

- Used a django form for the `DownloadPlot` view to handle type checks 
- Added validation check to ensure base64-encoded svg can be properly decoded
- (unrelated) prevent error if unauthenticated user attempts to logout


rabstejnek (approved):
Looks good!

PR #531, opened 2021-11-09
Change animal endpoint field "diagnostic" to "as_reported"

rabstejnek:


rabstejnek:
@shapiromatron Do you think the form or detail page should be updated in any way now that `as_reported` is required?

shapiromatron:
Closing; we're planning a future sprint for lots of animal data rewrites; this work will go there.

PR #533, opened 2021-11-12
DOI creation and cleanup

casey1173:
Dramatically expand creation of DOI in all HAWC literature imports. Automatically validates and creates  digital object Identifiers (DOIs) for References imported from PubMed, HERO, and RIS.

With some rough benchmarks, increases the number of DOIs in HAWC by around 50%, and references with a DOI by around 60-70%.

In addition, we added some new management commands to clean existing DOI:

```bash
# clean and validate existing DOIs to make sure they are in the correct format
manage.py doi --validate_existing

# parse existing reference metadata and all new DOI references where possible for all assessments
manage.py doi --create_from_existing

# same as above, but on a single assesment
manage.py doi --create_from_existing --assessment 1

# perform a full-text regex search on all metadata; higher hit-rate, lower DOI quality
manage.py doi --create_from_existing --full_text
```

Bonus changes

- updated django security version
- fixed a view bugs with some views missed in #507


hawc/apps/common/management/commands/clean_doi.py
```diff
@@ -0,0 +1,154 @@
...
+                    clean_count = clean_count + 1
+    print(f"\tCleaned and updated {clean_count} existing doi IDs")
+
+    # pull references with incorrect DOI id and replace that ID with the correct duplicate doi ID
+    print(f"Merging {len(duplicate_doi_ids)} duplicate doi ID(s)")
+    for dupe, old_doi in duplicate_doi_ids.items():
+        print(f"\t Merging doi:'{old_doi.unique_id}' into doi:'{dupe}'")
+        for reference in Reference.objects.filter(identifiers__id=old_doi.id):
+            reference.identifiers.remove(old_doi)
+            reference.identifiers.add(Identifiers.objects.filter(unique_id=dupe)[0])
```
rabstejnek, line 96: Maybe look into also saving the Identifiers from `doi_id_saved` into a list too, so that you don't have to make this query every time. Or make `doi_id_saved` into a dict with key `unique_id` and value `Identifier`?

hawc/apps/common/management/commands/clean_doi.py
```diff
@@ -0,0 +1,154 @@
...
+        for ids in ref.identifiers.all():
+            doi = constants.DOI_EXTRACT.search(ids.content)
+            if doi:
+                doi = doi.group(0)
+                if doi.endswith(","):
+                    doi = doi[: len(doi) - 1]
+                if doi.endswith('"'):
+                    doi = doi[: len(doi) - 1]
+                if doi.endswith("."):
+                    doi = doi[: len(doi) - 1]
```
rabstejnek, line 150: You can do `doi[:-1]` here and above.

hawc/apps/lit/managers.py
```diff
@@ -100,9 +100,11 @@ def get_from_ris(self, search_id, references):
             ids.append(ident)
 
             # create DOI identifier
-            if ref["doi"] is not None:
+            if constants.DOI_EXTRACT.search(str(ref["doi"])):
```
rabstejnek, line 103: Make a variable out of this regex search so that you don't have to perform the same search below.

rabstejnek: Do this for the other manager methods too.

scripts/DOI Cleanup.ipynb
```diff
@@ -0,0 +1,267 @@
+{
```
rabstejnek, line 1: You can remove this notebook if it's all been implemented in the command.

rabstejnek (changes requested):
Looks good! Just left a few comments that you're free to address. Let me know if you have any questions about what I wrote!

hawc/apps/common/management/commands/clean_doi.py
```diff
@@ -0,0 +1,154 @@
...
+        ):  # html conversion made a difference but theres an existing duplicate
+            duplicate_doi_ids[new_doi] = doi_id
+        elif not constants.DOI_EXACT.fullmatch(
+            new_doi
+        ):  # doi needs to be further cleaned/validated
+            new_doi = constants.DOI_EXTRACT.search(new_doi)
+            if new_doi:
+                new_doi = new_doi.group(0)
+            if new_doi and new_doi.endswith("."):  # remove period at end of doi if it exists
+                new_doi = new_doi[: len(new_doi) - 1]
```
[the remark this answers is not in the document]
casey1173: yes quite a few of them ended with a period, I think from users copy/pasting it from other citations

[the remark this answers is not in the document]
casey1173: in fact it might be worth adding this case to the manager imports since the regex doesn't catch it

hawc/apps/common/management/commands/clean_doi.py
```diff
@@ -0,0 +1,158 @@
...
+            if new_doi and new_doi.endswith("."):  # remove period at end of doi if it exists
+                new_doi = new_doi[:-1]
+            if new_doi is None or not constants.DOI_EXACT.fullmatch(new_doi):
+                # extraction failed; doi is tarnished
+                tarnished_doi_ids.append(doi_id)
+            else:
+                if new_doi in doi_id_saved.keys():
+                    other_dupes = duplicate_doi_ids.get(new_doi, [])
+                    other_dupes.append(doi_id)
+                    duplicate_doi_ids[new_doi] = other_dupes
```
shapiromatron, line 81: use [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict); `duplicate_doi_ids=defaultdict(list)` and then you just `duplicate_doi_ids[key].append(value)`

hawc/apps/common/management/commands/clean_doi.py
```diff
@@ -0,0 +1,158 @@
...
+            new_doi = constants.DOI_EXTRACT.search(new_doi)
+            if new_doi:
+                new_doi = new_doi.group(0)
+            if new_doi and new_doi.endswith("."):  # remove period at end of doi if it exists
+                new_doi = new_doi[:-1]
+            if new_doi is None or not constants.DOI_EXACT.fullmatch(new_doi):
+                # extraction failed; doi is tarnished
+                tarnished_doi_ids.append(doi_id)
+            else:
+                if new_doi in doi_id_saved.keys():
```
shapiromatron, line 78: dont use keys, just say `if new_doi in doi_id_saved`

hawc/apps/common/management/commands/clean_doi.py
```diff
@@ -0,0 +1,158 @@
...
+    doi_id_saved = {}
+    doi_id_updated = []
+    for doi_id in doi_identifiers:
+        if constants.DOI_EXACT.fullmatch(doi_id.unique_id):
+            doi_id_saved[doi_id.unique_id] = doi_id
+            # doi_id_saved.append(doi_id.unique_id)
+
+    duplicate_doi_ids = {}  # stores unclean DOIs that have a clean duplicate for updating later
+    tarnished_doi_ids = []  # tarnished DOI IDs cannot be cleaned/validated (ex: '10')
+    for doi_id in doi_identifiers:
```
shapiromatron, line 52: rename to `doi_ident`, or just `ident`? when i see `doi_id`, i think it's just a str or an id, not an object

hawc/apps/common/management/commands/clean_doi.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # pull references with incorrect DOI id and replace that ID with the correct duplicate doi ID
+    print(f"Merging {len(duplicate_doi_ids)} duplicate doi ID(s)")
+    for dupe, old_dois in duplicate_doi_ids.items():
+        import pdb
+
+        pdb.set_trace()
+        for old_doi in old_dois:
+            print(f"\t Merging doi:'{old_doi.unique_id}' into doi:'{dupe}'")
+            for reference in Reference.objects.filter(identifiers__id=old_doi.id):
```
shapiromatron, line 98: use the reverse relation; `old_doi.references.all()`

shapiromatron (commented):
Looking good! I think all this code works; most of my comments here are around making things [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). To set a marker in the sand, right now we have around 200 LOC.   My requested changes will likely add more code b/c we'll be able to test our DOI extractor methods, but the non-test code should be much less than 200 now.

rabstejnek (changes requested):
Looks good! My only requested change is the one concerning `defaultdict`, the other is a comment we may want to bring @shapiromatron in on.

hawc/services/utils/doi.py
```diff
@@ -0,0 +1,41 @@
...
+from hawc.apps.lit import constants
+
+
+def get_doi_if_valid(text: str):
+    doi = constants.DOI_EXTRACT.search(text)
+    if doi:
+        doi = doi.group(0)
+        doi = html.unescape(doi)
+        doi = urllib.parse.unquote(doi)
+        while doi.endswith((".", ",", '"')):
```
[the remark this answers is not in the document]
shapiromatron: I agree @rabstejnek , but couldn't figure out how. When I added w/ some toy examples the regex wasn't greedy enough; ie., it would stop at the first period, which is often valid.

hawc/apps/common/management/commands/clean_doi.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Args:
+        doi_identifiers (QuerySet[Identifiers]): List of doi IDs to be checked
+    """
+    self.stdout.write(f"Validating {doi_identifiers.count()} doi IDs...")
+
+    doi_id_saved = {}
+    doi_idents_updated = []
+    # duplicate_doi_ids: stores unclean DOIs that have a clean duplicate for updating later
+    duplicate_doi_ids = DefaultDict(list)
```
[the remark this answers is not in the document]
shapiromatron: https://docs.python.org/3/library/collections.html#collections.defaultdict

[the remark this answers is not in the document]
shapiromatron: @casey1173; can you resolve the linting issue after this change? after that I will review

hawc/services/utils/doi.py
```diff
@@ -0,0 +1,47 @@
...
+
+    Returns:
+        Optional[str]: A DOI string if one can be found
+    """
+    text = html.unescape(text)
+    text = urllib.parse.unquote(text)
+    doi = constants.DOI_EXTRACT.search(text)
+    if doi:
+        doi = doi.group(0)
+        if doi.endswith((".", ",", '"')):
```
casey1173, line 24: should be while

shapiromatron (approved):

rabstejnek (approved):

PR #534, opened 2021-11-15
split study detail and rob detail apis

shapiromatron:
The base study api endpoint was overloaded with all individual riskofbias reviews as well as the final review, which was generally the only review of interest for most data displays and visualizations.

This PR splits the data into two API requests:
- `/study/api/study/:id/`  - existing endpoint, but only shows the final evaluation
- `/study/api/study/:id/all-rob/` - new endpoint, shows all individual evaluations


rabstejnek (approved):
Looks good!

PR #535, opened 2021-11-16
Api fixes

rabstejnek:
Updates to API after testing openapi schema.


hawc/apps/assessment/serializers.py
```diff
@@ -64,7 +64,7 @@ class AssessmentRootedSerializer(serializers.ModelSerializer):
     NO_PARENT = -1
 
     def get_parent(self, assessment_id, validated_data, canSelectRoot):
-        parent_id = validated_data.pop("parent")
+        parent_id = validated_data.pop("parent", self.NO_PARENT)
```
rabstejnek, line 67: There was an error if no `parent` was given; if you don't think this is a suitable default, we can try to make it required instead.

hawc/apps/lit/serializers.py
```diff
@@ -224,7 +224,7 @@ def to_representation(self, instance):
         assessment_id = self.context["assessment"].id
         root = models.ReferenceFilterTag.get_assessment_root(assessment_id)
         tree = root.dump_bulk(root, keep_ids=True)
-        return {"tree": tree[0]["children"]}
+        return {"tree": tree[0].get("children")}
```
rabstejnek, line 227: If just the root tag w/o children then `children` wouldn't be set; this circumvents the KeyError that was thrown.

hawc/apps/lit/serializers.py
```diff
@@ -372,6 +372,9 @@ def update(self, instance, validated_data):
         # updates the reference tags
         if "tags" in validated_data:
             instance.tags.set(validated_data.pop("tags"))
+        # updates the searches
```
rabstejnek, line 375: `searches` is required for PUT requests, so this allows them to be set.

hawc/apps/mgmt/serializers.py
```diff
@@ -3,13 +3,17 @@
...
 from ..myuser.serializers import HAWCUserSerializer
+from ..study.models import Study
 from ..study.serializers import StudyAssessmentSerializer
 from . import models
 
 
 class TaskSerializer(serializers.ModelSerializer):
     owner = HAWCUserSerializer(read_only=True)
     study = StudyAssessmentSerializer(read_only=True)
+    study_id = serializers.PrimaryKeyRelatedField(
```
rabstejnek, line 14: Study is needed for writes; this allows it to be passed in as `study_id`

hawc/apps/riskofbias/serializers.py
```diff
@@ -104,7 +104,7 @@ class Meta:
 
 
 class RiskOfBiasScoreSerializer(serializers.ModelSerializer):
-    overridden_objects = RiskOfBiasScoreOverrideObjectSerializer(many=True)
+    overridden_objects = RiskOfBiasScoreOverrideObjectSerializer(many=True, read_only=True)
```
rabstejnek, line 107: The override serializer is not really setup for writes, so we just ignore it on writes.

hawc/apps/assessment/serializers.py
shapiromatron: I changed to make required; we'll always need to have a parent

hawc/apps/lit/serializers.py
shapiromatron: changed to return an empty list instead of null `tree[0].get("children", [])`

shapiromatron (approved):
Approved; if your ok with my minor change, feel free to merge.

PR #537, opened 2021-11-17
Made search/import descriptions use rich text

rabstejnek:
Updated the search and import forms to use rich text for the description input.

image

This rich text is now displayed wherever description was displayed.

image

image


rabstejnek:
@shapiromatron Since this rich text can lead to much more spaced out descriptions, should we remove it from the lit overview table? Here's an extreme example:
image

shapiromatron (approved):

shapiromatron:
> @shapiromatron Since this rich text can lead to much more spaced out descriptions, should we remove it from the lit overview table? Here's an extreme example: image

Looks like we could use [truncatewords_html](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#truncatewords-html) though I'm not sure how it would behave with both html and non-html.  

I'm not that worried about it though. 

```python
from hawc.apps.lit.models import Search
from django.db.models.functions import Length
import pandas as pd

d=Search.objects.filter(search_type='s').annotate(size=Length('description')).values_list('size', flat=True)
ser = pd.Series(d)
ser.hist()
ser.describe()
```

```
In [17]: ser.describe()
Out[17]:
count    1634.000000
mean       22.050184
std        42.583745
min         0.000000
25%         0.000000
50%         4.000000


75%        32.000000
max       578.000000
dtype: float64
```

Figure_1

PR #539, opened 2021-11-18
layout updates for internal communication

shapiromatron:
- Only show internal communications when data exists
- Add an asterisk to present fields which are not shown unless logged in
- Allow for wysiwyg data entry via [quill](https://quilljs.com/)

**Team views:**
Screen Shot 2021-11-18 at 10 51 56 AM
Screen Shot 2021-11-18 at 10 53 51 AM


hawc/apps/study/templates/study/study_detail.html
```diff
@@ -56,18 +56,25 @@ <h2 class="d-inline-block">{{object}}</h2>
 
   <div id="study_details"></div>
 
-  {% if obj_perms.edit %}
+  {% if obj_perms.edit and internal_communication %}
```
casey1173, line 59: once a communication is added, Internal Communications is always listed on the detail page--even if it is updated to say nothing. Not really an issue but it could be changed by adding `and internal_communication != "<p><br></p>"` to the if statement above, and for the corresponding assessment_detail template if statement

casey1173 (changes requested):
Looks good! Everything seems functional. I have one change I'd recommend, see my first comment

shapiromatron:
@casey1173 ok, addressed your (good) comments - how does it look now?

casey1173 (approved):

PR #541, opened 2021-11-19
redirect to url w/ external authentication login

shapiromatron:
Just like built-in django login; pass through the correct path for for redirecting after successful authentication with an external system. For example, the external url submits and returns a `next` query argument, eg., `/user/login/wam/?next=/assessment/123456/`


rabstejnek (approved):
Looks good!

PR #543, opened 2021-11-19
Move model choices

rabstejnek:
Moved all model choices from `model` module into `constants` module for each app.
Also updated the choices to use [Django's enumeration types](https://docs.djangoproject.com/en/3.2/ref/models/fields/#enumeration-types).

**Note**:
- Boolean choices were not converted to use Django's enumeration types, since `bool` cannot be subclassed.
- `SCORE_CHOICES` was also not converted, since naming the entries becomes troublesome with the duplicate labels.
  - Perhaps we can separate them into different enums, and then concatenate them for the field choices, ie
    ```
    class HighLowBiasScore(IntegerChoices):
      NA = 10, "Not applicable"
      NR = 12, "Not reported"
      ...

    class GoodDeficientScore(IntegerChoices):
      NA = 20, "Not applicable"
      NR = 22, "Not reported"
      ...

    score = models.PositiveSmallIntegerField(choices=HighLowBiasScore.choices + GoodDeficientScore.choices)
    ```
    This can be included in this PR or a separate one if it seems like a good idea.


rabstejnek:
@shapiromatron Let me know what you think about this pattern and naming conventions, and then I can apply it to the other apps!

hawc/apps/animal/actions/model_metadata.py
```diff
@@ -10,6 +10,10 @@ def tuple_to_dict(tuple):
     return {e[0]: e[1] for e in tuple}
 
 
+def enum_to_dict(Enum):
```
shapiromatron, line 13: move to common app; change argument signature `enum_to_dict(enum: Enum) -> Dict`

hawc/apps/animal/actions/model_metadata.py
```diff
@@ -25,36 +29,36 @@ def study_metadata(self):
...
             strains=list(Strain.objects.all().values("id", "species_id", "name").order_by("id")),
-            lifestage=tuple_to_dict(models.AnimalGroup.LIFESTAGE_CHOICES),
+            lifestage=enum_to_dict(constants.Lifestage),
         )
 
     def dosing_regime_metadata(self):
         return dict(
-            route_of_exposure=models.DosingRegime.ROUTE_EXPOSURE_CHOICES_DICT,
+            route_of_exposure=enum_to_dict(constants.RouteExposure),
             positive_control=tuple_to_dict(models.DosingRegime.POSITIVE_CONTROL_CHOICES),
```
shapiromatron, line 46: get rid of this `tuple_to_dict` as well?

hawc/apps/animal/constants.py
```diff
@@ -0,0 +1,144 @@
...
+    NA = "", "N/A (not generational-study)"
+    P0 = "P0", "Parent-generation (P0)"
+    F1 = "F1", "First-generation (F1)"
+    F2 = "F2", "Second-generation (F2)"
+    F3 = "F3", "Third-generation (F3)"
+    F4 = "F4", "Fourth-generation (F4)"
+    OT = "Ot", "Other"
+
+
+# these choices for lifesage expose/assessed were added ~Sept 2018. B/c there
```
shapiromatron, line 43: put this comment in a [triple-quoted string w/ numpy format](https://github.com/numpy/numpy/blob/v1.21.0/numpy/lib/npyio.py#L1202-L1447)

hawc/apps/animal/forms.py
```diff
@@ -144,23 +144,23 @@ def __init__(self, *args, **kwargs):
...
+        lifestage_dict = dict(constants.Lifestage.choices)
 
         if self.instance.lifestage_exposed in lifestage_dict:
-            le_choices = models.AnimalGroup.LIFESTAGE_CHOICES
+            le_choices = constants.Lifestage.choices
         else:
             le_choices = (
                 (self.instance.lifestage_exposed, self.instance.lifestage_exposed),
-            ) + models.AnimalGroup.LIFESTAGE_CHOICES
+            ) + constants.Lifestage.choices
```
shapiromatron, line 155: does this actually work? it looks like it should, but wanted to make sure...

hawc/apps/animal/forms.py
```diff
@@ -758,9 +758,9 @@ class EndpointFilterForm(forms.Form):
     )
 
     sex = forms.MultipleChoiceField(
-        choices=models.AnimalGroup.SEX_CHOICES,
+        choices=constants.Sex.choices,
         widget=forms.CheckboxSelectMultiple,
-        initial=[c[0] for c in models.AnimalGroup.SEX_CHOICES],
+        initial=[c[0] for c in constants.Sex.choices],
```
shapiromatron, line 763: use ` ExperimentType.__members__.keys()` instead?

hawc/apps/animal/models.py
```diff
@@ -476,37 +429,11 @@ class DosingRegime(models.Model):
...
-    POSITIVE_CONTROL_CHOICES = ((True, "Yes"), (False, "No"), (None, "Unknown"))
-
-    NEGATIVE_CONTROL_CHOICES = (
-        ("NR", "Not-reported"),
-        ("UN", "Untreated"),
-        ("VT", "Vehicle-treated"),
-        ("B", "Untreated + Vehicle-treated"),
-        ("N", "No"),
+    # TODO tried Class PositiveControl(int,models.Choices), but can't subclass bool?
+    POSITIVE_CONTROL_CHOICES = (
```
shapiromatron, line 433: bummer; let's move to constants anyways and keep this comment that we cant use the choices for booleans?

shapiromatron (approved):
I think this will clean things up quite a bit.  Hopefully, you've tackled the most complicated one already, though lit may be a little tricky too.

Ok to proceed!

hawc/apps/animal/forms.py
rabstejnek: Yep!
image

hawc/apps/animal/forms.py
rabstejnek: Even better, it looks like I can just do `.values`!
image

shapiromatron:
Wow, this is a beast. Looking great, but it's going to take me a little while to review :)

PR #501, opened 2021-09-09
bugs (upper/lower CI, topic modeling scheduling, bmd page rendering) and frontend testing

shapiromatron:
**Assorted bugfixes:**

- [x] Fixed the upper and lower confidence interval calculations for dichotomous data on frontend and backend
   - ... added tests on backend and the frontend to ensure values are expected
- [x] Fixed literature topic modeling celery task to only schedule assessments with enough references
- [x] Fixed bug where page rendering failed if a recommended model can not be found (eg., two independent sessions)
     - refactor bmd.Recommendation tab (fixing bug was hard b/c code was hard to review)
- [x] fix `/in-vitro/assessment/:id/endpoints/` rendering bug

**Admin improvements**

- reviewed all hawc admin views and added appropriate select_related/prefetch_related for speed
    - assessment.Assessment and myuser.UserProfile revised and updated
    - assessment.TimeSpentEditing and ContentType are now read only
    - only show Endpoint Methodology/Endpoint Notes if text is present

**Bonus: 🚀🚀🚀**

- added the [mocha](https://mochajs.org/) testing framework for frontend tests AND run in github actions CI
- django minor version bump

The frontend test functionality as implemented is limited; right now, we can't test React components or things with external library such as jQuery. However, this new testing functionality can and is being used on pure javascript code and is therefore helpful for numerical unit-tests. Thus, using the testing may require some restructuring of existing code, but that may not be  bad thing...

PR #544, opened 2021-11-21
bmd - remove js `eval`; fix dose-response line plotting

shapiromatron:
Previously, we used the JS `eval` method to evaluate the y values when plotting a dose-response curve. This was not ideal because eval is not a recommended, so it was a todo to one day remove. This timeline to remove was advanced by a regression in #501.  The scope context rule are complicated, but the net result is when we refactored the test so it could be testable, it was no longer in scope for plotting. 

This PR (1) fixes this regression and (2) removes the `eval` approach for the dose-response plot.


rabstejnek (approved):
Looks good!

PR #545, opened 2021-11-23
add swagger API user interface 

shapiromatron:
Add a browsable swagger user interface `{{admin}}/api/swagger/` (staff only):

Screen Shot 2021-11-22 at 10 52 30 PM

To fix styling:

- removed custom code styling in css
- switch to [kbd](https://getbootstrap.com/docs/4.5/content/code/#user-input) for a few commands previously using code 
    -  conditional formatting options on data pivots
    - RIS instructions (`/lit/ris-export-instructions/`)
Screen Shot 2021-12-01 at 8 18 29 PM

### DIFF
--- a/frontend/summary/dataPivot/ConditionalFormat.js
+++ b/frontend/summary/dataPivot/ConditionalFormat.js
@@ -293,7 +293,7 @@ class _DataPivot_settings_conditional {
                         .add_select(parent.settings.type, hash.get(v), true)
                         .data("key", v);
                     self.discrete_styles.push(style);
-                    add_input_row(discrete, `Style for <code>${v}</code>:`, style);
+                    add_input_row(discrete, `Style for <kbd>${v}</kbd>:`, style);
                 });
 
                 if (vals.unique.length === 0) {
--- a/hawc/apps/assessment/views.py
+++ b/hawc/apps/assessment/views.py
@@ -283,6 +283,11 @@ class Error401(TemplateView):
     template_name = "401.html"
 
 
+@method_decorator(staff_member_required, name="dispatch")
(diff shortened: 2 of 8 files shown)


rabstejnek (changes requested):
Its functional, but there are some styling problems that make part of the swagger ui hard to read:
image
This is from the `hawc.css` stylesheet. Maybe it would be better to extend "admin/base_site.html" instead of "base.html", kind of like the admin assessment size view does? (https://github.com/shapiromatron/hawc/blob/main/hawc/templates/admin/assessment-size.html)
This should make it so that `hawc.css` isn't imported, and may also look nicer thematically since this is also an admin only view.

shapiromatron:
> Its functional, but there are some styling problems that make part of the swagger ui hard to read: image This is from the `hawc.css` stylesheet. Maybe it would be better to extend "admin/base_site.html" instead of "base.html", kind of like the admin assessment size view does? (https://github.com/shapiromatron/hawc/blob/main/hawc/templates/admin/assessment-size.html) This should make it so that `hawc.css` isn't imported, and may also look nicer thematically since this is also an admin only view.

I ilke the admin css idea; I'll give this a go...

shapiromatron:
@rabstejnek make required changes. I ended up getting rid of the custom css code, and switched over to `kbd` for the few places we needed it. 
Screen Shot 2021-12-01 at 8 24 52 PM

Using base admin failed when you have a dark-mode set on your computer, which I do, at least on my mac. [dark theme](https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#admin-theming)

rabstejnek (approved):

PR #546, opened 2021-11-23
audit change logs

shapiromatron:
This PR adds a level of redundancy to where audit changes are reported. Previously (see #496); changes to content with resulted in the creation of a new db change record. 

This PR does two things:
1. Adds an application log to `hawc.audit.change` that mirrors the change record in the database
2. Captures all django application logs and serves them through gunicorn in the container environment; these are further piped into the docker logs. (`django > gunicorn > docker`)

### DIFF
--- a/hawc/main/settings/staging.py
+++ b/hawc/main/settings/staging.py
@@ -43,9 +43,9 @@ elif email_backend == "CONSOLE":
 else:
     raise ValueError(f"Unknown email backend: {email_backend}")
 
-LOGGING["loggers"]["django"]["handlers"] = ["file"]
-LOGGING["loggers"]["hawc"]["handlers"] = ["file"]
-LOGGING["loggers"]["hawc.request"]["handlers"] = ["hawc-request"]
+LOGGING["loggers"]["django"]["handlers"] = ["console", "file"]
+LOGGING["loggers"]["hawc"]["handlers"] = ["console", "file"]
+LOGGING["loggers"]["hawc.request"]["handlers"] = ["console", "hawc-request"]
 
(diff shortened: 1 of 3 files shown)


hawc/main/settings/staging.py
```diff
@@ -43,9 +43,9 @@
 else:
     raise ValueError(f"Unknown email backend: {email_backend}")
 
-LOGGING["loggers"]["django"]["handlers"] = ["file"]
-LOGGING["loggers"]["hawc"]["handlers"] = ["file"]
-LOGGING["loggers"]["hawc.request"]["handlers"] = ["hawc-request"]
+LOGGING["loggers"]["django"]["handlers"] = ["console", "file"]
+LOGGING["loggers"]["hawc"]["handlers"] = ["console", "file"]
+LOGGING["loggers"]["hawc.request"]["handlers"] = ["console", "hawc-request"]
```
rabstejnek, line 49: Do you not have to add a handler for your new `hawc.audit.change` logger?

rabstejnek (approved):
Looks good! I left one comment that may need to be addressed.

hawc/main/settings/staging.py
shapiromatron: Nope; it's handled by the "hawc" logger as a parent.

shapiromatron: Didn't find a great link in the python docs... see https://khashtamov.com/en/ultimate-guide-to-python-logging/

> Loggers can inherit logging settings from their ancestors. The inheritance is defined by the . in logger's name. If you have the logger called my_package.logger1 it will inherit the settings defined for my_package.

rabstejnek: Ahh makes sense!

PR #548, opened 2021-11-26
add new email provider - sendgrid

anthonyWillisInfotel:
This permits the use of the sendgrid smtp api for emails within HAWC.
I have tested this using the sendgrid settings provided to me by IARC tech support.


shapiromatron (approved):
LGTM!

PR #551, opened 2021-11-29
test fixture improvements; api tweaks

shapiromatron:
- create iris dataset files with `load_test_db` command
- fix media preview admin api w/ json recursion on filesystem paths
- fix lit topic model api when not enough references
- add 30-yr auth token for test fixture
- add name to dose-units search field
- for animal endpoint API creation, relax mutual exclusion in api; not enforced in form view, so this makes the API and form consistent


shapiromatron:
@rabstejnek can you prioritize this review please ;) I need to deploy a few of these changes on prime

rabstejnek (approved):
Looks good! Ran the OpenAPI schema notebook and the problematic get requests are now working like a charm!

PR #552, opened 2021-11-30
Client updates2

shapiromatron:


shapiromatron:
Closed; merging key changes into https://github.com/shapiromatron/hawc/tree/2022-client

PR #553, opened 2021-12-01
Study outcome table (backend)

rabstejnek:
# Study outcome table (backend)

This first iteration of our study outcome table is rolled up to the study level, ie, a list of rows with corresponding study ids are provided to build this table.

Since the underlying data can be deleted, it is handled in the following way: any studies that are not found during data extraction are replaced by a row in the table noting that it was not found, instead of throwing an error. Likewise, any risk of bias score not found is defaulted to "NM".

Other row types (ie animal group, endpoint, etc) and column types (ie NOEL, LOEL, etc) will be implemented in future PRs.

## Snippets

```python 
StudyOutcomeTable.parse_obj({
  "assessment_id": 100000037,
  "subheaders": [{"label": "This is a subheader", "start": 2, "length": 2}],
  "columns": [
    {"label": "Study name", "attribute": "study_name", "key": "1", "width": 1},
    {"label": "Animal description", "attribute": "species_strain_sex", "key": "2", "width": 2},
    {"label": "Doses", "attribute": "doses", "key": "3", "width": 2},
    {"label": "Rob score", "attribute": "rob_score", "metric": 100001110, "key": "4", "width": 4},
  ],
  "rows": [
    {"id": 100000970, "type": "study", "customized": []},
    {"id": -1, "type": "study", "customized": []},
    {"id": 100000892, "type": "study", "customized": [{"key":"3","html":"<p>This is a custom dose</p>"},{"key":"4","score_id":-1}]},
  ],
})
```
image


hawc/apps/summary/models.py
```diff
@@ -247,6 +252,10 @@ def to_docx(self, base_url: str = ""):
         return ReportExport(docx=docx, filename=self.slug)
 
     def clean(self):
+        # assessment id should match content
```
rabstejnek, line 255: This is my idea on how to make sure `content` doesn't include any objects that are outside of the scope of this assessment; if `assessment_id` is in content, it MUST match the `SummaryTable`'s assessment_id. All other validation (aka study/metric validation for study outcome tables) are done in the table's pydantic model using the given `assessment_id`.

hawc/apps/summary/models.py
```diff
@@ -239,6 +242,8 @@ def build_default(cls, assessment_id: int, table_type: int) -> "SummaryTable":
         instance = cls(assessment_id=assessment_id, table_type=table_type)
         schema = instance.get_content_schema_class()
         instance.content = schema.get_default_props()
+        if "assessment_id" in instance.content:
```
rabstejnek, line 245: This allows a default to be given when it is assessment specific. Note the default for the study outcome table is pretty much empty since it relies a lot on other ids (study/metric), but at least it still follows the pattern.

hawc/tools/tables/sot.py
```diff
@@ -0,0 +1,140 @@
...
+        return BaseCellGroup.construct(cells=cells)
+
+    def _rows_group(self):
+        cells = []
+
+        study_ids = {row.id for row in self.cell_rows if row.type == "study"}
+        final_scores = pd.DataFrame.from_records(
+            FinalRiskOfBiasScore.objects.filter(study_id__in=study_ids).values()
+        )
+        bioassay_data = Endpoint.heatmap_study_df(
```
rabstejnek, line 96: I am using `heatmap_study_df` for all of the non-RoB data. However, it may be a little incomplete when it comes to columns we want to show, and things like sex/species/strains are aggregated by piping them together, which may not be what we want for this table. Should we:
1) split strings that are piped together and present it in more of a readable format for the table
2) add other fields to `heatmap_study_df` that we need for this table (ie it provides dose units, but not actual doses)
3) OR just make a new API endpoint altogether specifically suited to this table (maybe even revisit the animal bioassay materialized view?)

hawc/apps/summary/models.py
shapiromatron: I don't think a user should be able to provide this in the content; we should update any method calls to pass the assessment_id and the content on the backend so that it's immutable and available where needed. Is this a big lift?

hawc/apps/summary/models.py
shapiromatron: Same as above, we should remove this from the content and just rely on `instance.assessment_id`

hawc/tools/tables/sot.py
shapiromatron: I was thinking we could use the `heatmap_doses_df` actually so we could get down to loel/noel data in the future.  We'd need to have rollup functions by data type, so for now, you could do a rollup for species by study for example as a starting point and

hawc/tools/tables/parser.py
```diff
@@ -29,6 +29,13 @@ def ol_wrapper(texts):
     return f"<ol>{''.join(list_items)}</ol>"
 
 
+def color_background(cell, color):
```
shapiromatron, line 32: add annotations. is color a hex color str?

hawc/tools/tables/sot.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+class AttributeChoices(Enum):
+    RobScore = "rob_score"
+    StudyName = "study_name"
+    SpeciesStrainSex = "species_strain_sex"
+    Doses = "doses"
+    FreeHtml = "free_html"
+
+    def get_rob_score(self, selection: pd.DataFrame) -> BaseCell:
```
shapiromatron, line 23: it would be great to have a unit-test for each of these methods. I think we may have to replicate this on the frontend too

hawc/tools/tables/sot.py
```diff
@@ -0,0 +1,174 @@
...
+    type: str
+    customized: List
+
+
+class JudgmentCell(BaseCell):
+    judgment: int
+
+    def to_docx(self, parser: QuillParser, block):
+        text = SCORE_SYMBOLS[self.judgment]
+        color = SCORE_SHADES[self.judgment].strip("#")
```
shapiromatron, line 79: does `lstrip` work? might be better if possible

hawc/tools/tables/sot.py
```diff
@@ -0,0 +1,174 @@
...
+    key: str
+
+
+class Row(BaseModel):
+    id: int
+    type: str
+    customized: List
+
+
+class JudgmentCell(BaseCell):
```
shapiromatron, line 74: rename to `RobJudgment`? Since the EPT also has judgment too, it might be good from to identify which one.

hawc/tools/tables/sot.py
```diff
@@ -0,0 +1,174 @@
...
+        return getattr(self, f"get_{self.value}")(selection)
+
+
+class Subheader(BaseModel):
+    label: str
+    start: conint(ge=0)
+    length: conint(ge=1)
+
+
+class Column(BaseModel):
```
shapiromatron, line 61: we probably need to add a width here, with a default value of 1 for all to be equal widths?

shapiromatron (changes requested):
Looking great Danny. I agree with all of your design choices so far. 

I have three larger comments:
1. add a `width` parameter or something similar to the columns so they can be resized
2. add some unit-tests for the individual cell item attributes
3. add basically a larger test to demonstrate successful demonstration of the `StudyOutcomeTable`

hawc/tools/tables/sot.py
```diff
@@ -0,0 +1,174 @@
...
+            return bioassay_data.loc[bioassay_data["study id"] == row.id]
+
+    def _filter_rows(self, bioassay_data: pd.DataFrame) -> List[Row]:
+        return [row for row in self.cell_rows if row.id in bioassay_data["study id"].values]
+
+    def _rows_group(self):
+        cells = []
+
+        study_ids = {row.id for row in self.cell_rows if row.type == "study"}
+        final_scores = pd.DataFrame.from_records(
```
shapiromatron, line 135: you might be able to set the index for the study_id and metric_id here, that may make the lookups faster, if those keys are always reuqired which i think they are

hawc/tools/tables/sot.py
shapiromatron: to close the loop on our chat, feel free to use whatever existing mechanisms we already have to solve this. it just seemed like a good home on the current data structures

shapiromatron:
Based on our chat, added a few more things:

- Use the grey "NM" not measured symbol for an SQJudgment not found
- if a row identifier isn't found, still dont crash hard, but print in the first column `{type} {id}` not found eg., `study 1231231` not found

Danny brought up a good limitation here - unless data are extracted, the studies won't be shown. We may have to fix this with future versions of the table, and should indicate that this is a limitation.

hawc/tools/tables/sot.py
rabstejnek: Overrides are also included in the dataframe, so `study_id` and `metric_id` won't make unique indices.

shapiromatron: unlike databases, i dont htink indexes in data frames have to be unique. https://stackoverflow.com/questions/16626058/what-is-the-performance-impact-of-non-unique-indexes-in-pandas

rabstejnek: Good to know! I'll give it a shot. I was deterred by the ominous wording under `verify_integrity`: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.set_index.html

rabstejnek: I added the index in my latest commit fd34bba7c47451f256b317d826feb6ffe53c18ce. There was a bit of a tradeoff, which was kind of alluded to in your SO link; if the index is unique, then `loc` will return a series, and if its not unique, then it will return a dataframe. I also had to take into account `KeyError`s when indexing into the dataframe.

shapiromatron: Do you think this tradeoff is worth it? If so, we can keep, otherwise I'm fine with reverting.

hawc/tools/tables/base.py
```diff
@@ -41,10 +41,18 @@ def parse_args(cls, *args):
 class BaseCellGroup(BaseModel):
     cells: List[BaseCell] = []
 
+    class Config:
+        underscore_attrs_are_private = True
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self._setup()
```
shapiromatron, line 49: sorry, just saw this, is this still needed? I didn't see it being used anywhere

shapiromatron (approved):

rabstejnek:
@shapiromatron I've added a new `data_source` field for study outcome tables and an API endpoint to retrieve queried data necessary to build tables on the frontend.

rabstejnek:
@shapiromatron this is ready for re-review!

rabstejnek:
Rolled up into #581

PR #547, opened 2021-11-24
add searching/filtering to study, assessment, and visual lists

casey1173:
### updated study list filtering:
- filter by Name, data type, identifiers, and published/unpublished (team members only)
- added button links for studies with identifiers
- removed javascript filtering
image

### updated public assessments page: 
- removed 'Version' column
- added Name filtering and ordering
- added brief metadata descriptions of references, assessments, and visualizations in HAWC
image

### updated visual list filtering:
- filter by title or caption text, visual type, and published/unpublished (team members only)
- removed javascript filtering
image

PR #554, opened 2021-12-06
enable copying of visuals and tables

casey1173:
Data pivots, and many extractions can be copied or "cloned".

Adds ability to copy preexisting visualizations
image

Adds ability to copy preexisting summary tables
image

We made a simplifying assumption only allow copying within the same assessment. This can be changed if needed, but it would require modifying visual settings to remove assessment-specific settings.

Additional changes:

- make datetimes sortable in #547
- (regression) fixed missing constants from #543


shapiromatron (approved):
Looks good! A few minor comments on data structures that I changed, but this looked good. In retrospect this was a lot more challenging than I realized; nice job...

PR #555, opened 2021-12-08
EPT fix

rabstejnek:
The errors encountered in some EPT tables stem from changing `short_description` from a text field to a rich text field (#459).

To fix the javascript errors, I had to wrap the text passed to `hasInnerText` with a <p\> tag so that jQuery was always working with a correct selector expression.

However, there was still problems with how the table was rendering:

image

To fix this I added the function `addOuterTag`, which ensures that a given tag encases a given string. This allows both html and non-html strings to work with `injectText` and `injectPopup`.

Example table:
http://127.0.0.1:8000/summary/assessment/100000061/tables/EPT-Cardiometabolic-Effects-Inhalation-Chloroform/


shapiromatron (approved):
looks good; i added some unit tests to figure out `addOuterTag` works. Looks like it fails if the tag has an attribute, but that shouldn't be an issue Quill outputs.  Had to do a little import magic b/c of how jQuery gets injected into webpack.

PR #558, opened 2021-12-14
add DOI to bulk tagging

casey1173:
Bulk reference tagging previous worked with HERO and PubMed IDs; expand to also allow DOIs.

image


hawc/apps/lit/managers.py
```diff
@@ -484,6 +484,13 @@ def identifiers_dataframe(self, qs: QuerySet) -> pd.DataFrame:
         for hawc_id, pubmed_id in pubmeds:
             data[hawc_id]["pubmed_id"] = int(pubmed_id)
 
+        # capture DOI ids
```
shapiromatron, line 487: need to add doi to line 466:

```python
captured = {None, constants.HERO, constants.PUBMED}
```

rabstejnek (approved):

shapiromatron (approved):

PR #561, opened 2021-12-21
search for reference and study on identifier admin

casey1173:
In the admin, add the ability to search for a given PubMed or HERO or DOI, and view all references and studies across all assessments.

image


shapiromatron:
@casey1173 Regarding your site stuff, you can just edit that on your local version of hawc. Just change here http://127.0.0.1:8000/admin/sites/site/. No need bake any of these changes into the source code.

See https://docs.djangoproject.com/en/4.0/ref/contrib/sites/

shapiromatron (approved):
nice job! confusing code b/c of the reverse m2m relationships, but i played around and that looks like the only way to do it!

PR #563, opened 2021-12-21
convert test files into a package

casey1173:
Convert the individual file structure of tests into a python package. This lifts the unique filename restriction and allows for the possibility of building reusable test-helper methods.


tests/data/cassettes/apps.epi.test_api/TestExposureApi.test_valid_requests.yaml
```diff
@@ -0,0 +1,73 @@
+interactions:
```
shapiromatron, line 1: @casey1173 you shoudn't need to create these, i think you should just be able to rename the existing.

For example, this one is `tests/data/cassettes/test_epi_api/TestExposureApi.test_valid_requests.yaml`

casey1173: thanks! I was wondering what was going on with the 11,000+ LOC added, I didn't even touch those files

shapiromatron (approved):
Approved; I had to merge into main to get the unit tests to run on github; it might be good to make sure tests run before requesting review.  After the tested ran, I had to rename the an additional folder of vcr cassettes, but then passed with now issues.

PR #564, opened 2021-12-23
htmx helper utility + htmx for assessment attachments

casey1173:
Created a `hawc.apps.common.htmx.HtmxViewSet` base class to encapsulate common htmx operations.  Designed with concept similar to  the django-rest-framework [ViewSet](https://www.django-rest-framework.org/api-guide/viewsets/). This class view which abstracts away much of the boilerplate code, designed for use with [htmx](https://htmx.org/), including:

- built in permissions checking
- automated logging and auditing of write operations

Implemented this new pattern on the assessment level attachment field. Attachments can now be added inline instead of going through a series of django pages for a smoother user experience.

image

Also made a few minor changes to the assessment list page:

- an indicator is shown if the attachment is publicly available (or not)
- descriptive text is now required for each attachment


hawc/apps/assessment/views.py
```diff
@@ -448,31 +449,105 @@ class AttachmentCreate(BaseCreate):
...
+        context["canEdit"] = context["obj_perms"]["edit"]
+        # context["newAttach"] = True
+        return render(self.request, "assessment/components/attachment_row.html", context)
 
 
 class AttachmentRead(BaseDetail):
     model = models.Attachment
 
     def get(self, request, *args, **kwargs):
+        if request.GET.get("test", -1) != -1:
```
shapiromatron, line 477: it might be easier to check if the header is htmx and if so return the template, i think htmx requests use a header which would be nice to have an is_htmx utility function

hawc/apps/assessment/views.py
```diff
@@ -448,31 +449,105 @@ class AttachmentCreate(BaseCreate):
...
             return PermissionDenied
 
 
+class AttachmentList(BaseList):
+    model = models.Attachment
+    parent_model = models.Assessment
+    parent_template_name = "parent"
+    template_name = "assessment/_attachment_list.html"
+    object_list = None
+    form_class = forms.NewAttachmentForm
```
shapiromatron, line 496: i don't think you need a form class here or in the context

hawc/apps/assessment/views.py
```diff
@@ -448,31 +449,105 @@ class AttachmentCreate(BaseCreate):
...
+
+
 class AttachmentUpdate(BaseUpdate):
     success_message = "Assessment updated."
+    template_name = "assessment/components/attachment_edit_row.html"
     model = models.Attachment
-    form_class = forms.AttachmentForm
+    form_class = forms.NewAttachmentForm
+
+    @transaction.atomic
```
shapiromatron, line 522: no need for this to be atomic since it's a single write op; you need atomic is there are multiple write ops that would want to run all or none

hawc/apps/assessment/views.py
```diff
@@ -448,31 +449,105 @@ class AttachmentCreate(BaseCreate):
...
+    def form_valid(self, form):
+        super().form_valid(form)
+        return render(
```
shapiromatron, line 525: instead, can you set the get success url to the attachment detail?

hawc/templates/base.html
```diff
@@ -45,8 +45,14 @@
     {% block extrastyle %}{% endblock %}
 
   </head>
-  <body class="html">
 
+  <body class="html">
+    <script src="https://unpkg.com/htmx.org@1.6.1"></script>
```
shapiromatron, line 50: can you add this to our static files to instead of using a cdn. see the naming and version pattern for other js

shapiromatron (commented):
had some fun looking around a bit, some of these may not be possible but ideas to investigate

hawc/apps/assessment/templates/assessment/components/attachment_row.html
```diff
@@ -0,0 +1,21 @@
+            <td><a href="{{object.attachment.url}}">{{object.title}}</a></td>
+            <td>{{object.description|safe}}</td>
+            {% if canEdit %}
+            <td>
+                <button
+                    class="btn btn-sm btn-info"
+                    hx-get="{{object.get_edit_url}}"
+                    hx-trigger="click"
+                    hx-target="#attach-row-{{object.pk}}"
+                    hx-swap="innerHTML">
```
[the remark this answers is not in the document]
casey1173: I tried this and the default target was the button

hawc/apps/assessment/views.py
casey1173: the form is part of the attachment list

hawc/apps/assessment/templates/assessment/components/attachment_edit_row.html
```diff
@@ -0,0 +1,43 @@
...
+                hx-get="{% url 'assessment:attachment_detail' object.pk %}">
+                Cancel</button>
+            {% endif %}
+            </div>
+        {% crispy form %}
+        </form>
+    </div>
+</td>
+{% block extrajs %}
+<script type="text/javascript">
```
shapiromatron, line 32: this is smart; i was trying to figure this out...

shapiromatron:
@casey1173  merged #568 into this PR; let me know how it goes with your work integrating. After you've got things in a good place w/ the attachments, I think we can remove all the changes we made in the ani app, that was just useful for designing the proof of concept....

shapiromatron (approved):
🎸

PR #565, opened 2022-01-03
epi version flag on assessment

munnsmunns:
Add a new field on assessment.Assessment named epi_version with choices 1 or 2. 2 is the default value, though it is set to 1 on existing data.

The value is updated on the /assessment/:id/enabled-modules/update/ page


hawc/apps/assessment/migrations/0030_assessment_epi_version.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.2.10 on 2022-01-03 20:11
```
shapiromatron, line 1: let's just create a single migration for this.  The default should be 2, but you may need to add a [RunPython](https://docs.djangoproject.com/en/4.0/ref/migration-operations/#django.db.migrations.operations.RunPython) to set the existing values to 1.

hawc/apps/assessment/migrations/0031_alter_assessment_epi_version.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.10 on 2022-01-03 20:11
```
shapiromatron, line 1: remove this file; manually edit 0030 above

hawc/apps/assessment/models.py
```diff
@@ -257,6 +257,10 @@ def get_rob_name_default():
     )
     created = models.DateTimeField(auto_now_add=True)
     last_updated = models.DateTimeField(auto_now=True)
+    epi_version = models.PositiveSmallIntegerField(
```
shapiromatron, line 260: add an `IntegerChoices` in the `assessment.constants` instead and revise the help-text

tests/hawc/apps/assessment/test_view_permissions.py
```diff
@@ -20,6 +20,7 @@
     "team_members_1": ("1", "2"),
     "reviewers_0": ("1",),
     "reviewers_1": ("1",),
+    "epi_version": ("1",),
```
shapiromatron, line 23: can you add to the `tests/data/fixtures.db.yaml` the epi_version: 1 for all the assessments there as well?

hawc/apps/assessment/forms.py
```diff
@@ -139,6 +139,7 @@ class Meta:
             "enable_risk_of_bias",
             "enable_bmd",
             "enable_summary_text",
+            "epi_version",
```
shapiromatron, line 142: in addition to including here, you probably need to exclude from the class AssessmentForm above

shapiromatron (changes requested):
Great first start! Added a few comments for changes, let me know if you'd like to discuss any futher...

hawc/apps/assessment/models.py
shapiromatron: and also, please keep the created and last_updated fields at the end and move this field above them

shapiromatron (changes requested):

hawc/apps/assessment/models.py
```diff
@@ -255,6 +255,11 @@ def get_rob_name_default():
         this option until you like to "freeze" your assessment, and then this can be unchecked, if
         needed.""",
     )
+    epi_version = models.PositiveSmallIntegerField(
+        choices=constants.EpiVersion.choices,
+        default=2,
+        help_text="Set based on Epi Version used for this assessment.",
```
shapiromatron, line 261: maybe change this text "Data extraction schema used for epidemiology studies"

hawc/apps/assessment/migrations/0030_assessment_epi_version.py
```diff
@@ -0,0 +1,26 @@
...
+
+from django.db import migrations, models
+
+
+def set_epi_version(apps, schema_editor):
+    # set epi_version to 1 on existing assessments
+    Assessment = apps.get_model("assessment", "Assessment")
+    db_alias = schema_editor.connection.alias
```
shapiromatron, line 9: remove the db_alias command? I think based on other migrations in hawc, you can just do

```python
apps.get_model("assessment", "Assessment").objects.update(epi_version=1)
```

shapiromatron: One more update, let's try to only update those where we have epi data extracted. 

```python
from django.db models import Count
Assessment.objects.annotate(n_sp=Count("references__study__study_populations")).filter(n_sp__gt=0).count()
```

There should be around 40?

shapiromatron (approved):
nice first time contribution 🚀 !!!

PR #566, opened 2022-01-05
Build models for epi v2

munnsmunns:
* Create models in epiv2/models.py
* Add constants.py with all enums and constants for epiv2
* build admin pages for epiv2
* integrate reversion
* create a new extraction in the db.yaml fixture (note: on my local machine, the tests would not pass until I used the command ```dropdb hawc-test```


hawc/apps/epiv2/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
shapiromatron, line 1: ok to delete this file

hawc/apps/epiv2/views.py
```diff
@@ -0,0 +1,3 @@
+from django.shortcuts import render
```
shapiromatron, line 1: and this one too (for now)

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,143 @@
...
+    last_updated = models.DateTimeField(auto_now=True)
+
+
+class Outcome(models.Model):
+    study_population = models.ForeignKey(
+        StudyPopulation, on_delete=models.CASCADE, related_name="outcomes"
+    )
+    health_outcome = models.CharField()
+    health_outcome_system = models.CharField()
+    measurement_timing = (
```
shapiromatron, line 105: i agree that we don't need another model, though i think we should put this on the `DataExtraction` table instead, since you may measure the same outcome over multiple time periods

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+    outcome = models.ForeignKey(Outcome, on_delete=models.SET_NULL)
+    exposure = models.ForeignKey(Exposure, on_delete=models.SET_NULL)
+    n = models.PositiveIntegerField()
+    effect_estimate = models.CharField()
+    effect_estimate_type = models.CharField()
+    effect_description = models.CharField()
+    exposure_rank = models.PositiveSmallIntegerField()
+    # TODO: what are these fields
+    # CI LCL
```
shapiromatron, line 131: confidence interval; lower confidence level; upper confidence level, standard deviation, standard error. they should all be optional float fields

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,143 @@
...
+    # CI UCL
+    # SD or SE
+    pvalue = models.FloatField()
+    # TODO: this is derived from the p-value (p-value>0.05). Necessary to have as a field?
+    significant = models.BooleanField()
```
shapiromatron, line 136: good question; this is is significance, as reported by study authors, 0.05 is a general rule but not always the case

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,143 @@
...
+        StudyPopulation, on_delete=models.CASCADE, related_name="adjustment_factors"
+    )
+    description = models.CharField()
+
+
+class DataExtraction(models.Model):
+    study_population = models.ForeignKey(
+        StudyPopulation, on_delete=models.CASCADE, related_name="data_extractions"
+    )
+    outcome = models.ForeignKey(Outcome, on_delete=models.SET_NULL)
```
shapiromatron, line 123: should be cascade for outcome

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,143 @@
...
+    exposure = models.ForeignKey(Exposure, on_delete=models.SET_NULL)
```
shapiromatron, line 124: i would suspect this would be the `ExposureLevel` not the exposure?

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,143 @@
...
+    measurement_type = models.ManyToManyField(MeasurementType)
+    measurement_timing = models.CharField()
+    exposure_route = models.CharField()
+    analytic_method = models.CharField()
+    comments = models.CharField()
+    created = models.DateTimeField(auto_now_add=True)
+    last_updated = models.DateTimeField(auto_now=True)
+
+
+class ExposureLevel(models.Model):
```
shapiromatron, line 78: lots of these float fields should allow null

shapiromatron (commented):
Looking great so far, just did some reviews of your WIP

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+class AgeProfile(models.Model):
+    objects = managers.AgeProfileManger()
+
+    name = models.CharField(unique=True, max_length=128,)
+
+    def __str__(self):
+        return self.name
+
+
+class Country(models.Model):
```
shapiromatron, line 18: remove this; use the epi.models.Country instead

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+    )
+    participant_n = models.PositiveIntegerField(
+        verbose_name="Overall study population N",
+        help_text="Enter the total number of participants enrolled in the study (after exclusions).\n"
+        + "Note: Sample size for specific result can be extracted in qualitative data extraction",
+    )
+    created = models.DateTimeField(auto_now_add=True)
+    last_updated = models.DateTimeField(auto_now=True)
+
+    def __str__(self):
```
shapiromatron, line 93: use an f string for this, added in python3.6: 

```python

return f"{self.study}/{self.summary}"
```

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+        blank=True,
+        null=True,
+        verbose_name="Study name (if applicable",
+        help_text="Typically available for cohorts. Abbreviations provided in the paper are fine",
+    )
+    countries = models.ManyToManyField(Country, blank=True,)
+    region = models.CharField(
+        max_length=128, blank=True, null=True, verbose_name="Other geographic information"
+    )
+    # TODO: is there a better way of doing this with datetime objects to make it searchable
```
shapiromatron, line 81: no, but we could have some custom regex validators?  Not sure we can define it that tight

https://docs.djangoproject.com/en/4.0/ref/validators/#regexvalidator

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+    measurement_type = models.ManyToManyField(
+        MeasurementType, verbose_name="Exposure measurement type", blank=True,
+    )
+    biomonitoring_matrix = models.CharField(max_length=128,)
+    measurement_timing = models.CharField(
+        max_length=128,
+        blank=True,
+        null=True,
+        verbose_name="Timing of exposure measurement",
+        help_text='If timing is based on something other than age, specify the timing (e.g., start of employment at Factory A). If cross-sectional, enter "cross-sectional"',
```
shapiromatron, line 147: i kind of like just having a really long string instead of the breaks?

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+    )
+    exposure_route = models.CharField(
+        max_length=2, choices=constants.ExposureRoute.choices, blank=True, null=True,
+    )
+    analytic_method = models.CharField(max_length=128, blank=True, null=True,)
```
shapiromatron, line 152: remove some of the trailing commas if the lines not broken

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+        AdjustmentFactor, on_delete=models.SET_NULL, blank=True, null=True,
+    )
+    confidence = models.CharField(
+        max_length=128, verbose_name="Study confidence", blank=True, null=True,
+    )
+    data_location = models.CharField(
+        max_length=128, help_text="e.g., table number", blank=True, null=True,
+    )
+    statistical_method = models.CharField(max_length=128, blank=True, null=True,)
+    comments = models.CharField(max_length=128, blank=True, null=True,)
```
shapiromatron, line 304: make this a TextField; no length required; do the same for others where the text length may be hard to pin-down

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+class DataExtraction(models.Model):
+    objects = managers.DataExtractionManager()
+
+    study_population = models.ForeignKey(
+        StudyPopulation, on_delete=models.CASCADE, related_name="data_extractions"
+    )
+    sub_population = models.CharField(max_length=64,)
+    outcome = models.ForeignKey(Outcome, on_delete=models.CASCADE, blank=True, null=True,)
```
shapiromatron, line 260: for outcome and exposure_level

```python
outcome = models.ForeignKey(Outcome, on_delete=models.CASCADE)
exposure_level = models.ForeignKey( ExposureLevel, on_delete=models.CASCADE)
```

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+    exposure_level = models.ForeignKey(
+        ExposureLevel, on_delete=models.SET_NULL, blank=True, null=True
+    )
+    measurement_timing = models.CharField(max_length=128, blank=True, null=True,)
+    n = models.PositiveIntegerField(blank=True, null=True,)
+    effect_estimate_type = models.CharField(
```
shapiromatron, line 266: For CharField and TextFIeld, never set null=True; blank=True is fine, b/c otherwise you'll have empty strings and nulls in the database and that makes typing really complicated

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+    )
+    effect_description = models.CharField(
+        max_length=128,
+        blank=True,
+        null=True,
+        verbose_name="Effect estimate description",
+        help_text="Description of the effect estimate with units, including comparison group if applicable",
+    )
+    effect_estimate = models.FloatField(blank=True, null=True,)
+    exposure_rank = models.PositiveSmallIntegerField(
```
shapiromatron, line 277: i think we should remove blank/null here and just set default=0

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+        blank=True,
+        null=True,
+        help_text="Rank this comparison group by exposure (lowest exposure group = 1)",
+    )
+    ci_lcl = models.FloatField(verbose_name="Confidence Interval LCL", blank=True, null=True,)
```
shapiromatron, line 282: you might not need blank for a nullalble FLoatFIeld ? I can't remember

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+    ci_ucl = models.FloatField(verbose_name="Confidence Interval UCL", blank=True, null=True,)
+    sd_or_se = models.FloatField(
+        verbose_name="Standard Deviation or Standard Error", blank=True, null=True,
+    )
+    pvalue = models.CharField(verbose_name="p-value", max_length=128, blank=True, null=True,)
+    significant = models.BooleanField(
```
shapiromatron, line 288: I see why you did this b/c its the current pattern in hawc but its not working well in other places; maybe better to just use a PositiveSmallIntegerField or CharField w/ choices?

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+        return str(self.study) + "/" + self.summary
+
+
+class Chemical(models.Model):
+    objects = managers.ChemicalManager()
+
+    study_population = models.ForeignKey(
+        StudyPopulation, on_delete=models.CASCADE, related_name="chemicals"
+    )
+    name = models.CharField(max_length=64)
```
shapiromatron, line 103: add an optional relation to assessment.models.DSSTox,

shapiromatron: length may need to be longer?

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+    objects = managers.ExposureManager()
+
+    name = models.CharField(
+        max_length=64,
+        help_text="A unique name for this exposure that will help you identify it later.",
+    )
+    study_population = models.ForeignKey(
+        StudyPopulation, on_delete=models.CASCADE, related_name="exposures"
+    )
+    measurement_type = models.ManyToManyField(
```
shapiromatron, line 138: will chat w/ team about this one... this seems tricky

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+        AgeProfile,
+        blank=True,
+        help_text='Select all that apply. Note: do not select "Pregnant women" if pregnant women '
+        + "are only included as part of a general population sample",
+        verbose_name="Population age category",
+    )
+    age_description = models.CharField(
+        max_length=64, blank=True, null=True, verbose_name="Population age details",
+    )
+    sex = models.CharField(default="U", max_length=1, choices=constants.Sex.choices,)
```
shapiromatron, line 61: use the constant instead of a string, eg.,  `constant.Sex.UKNOWN`

tests/data/fixtures/db.yaml
```diff
@@ -1,3 +1,7 @@
+- model: sites.site
```
shapiromatron, line 1: let's make this clean and only commit the changes that are being made by adding this app

hawc/main/settings/base.py
```diff
@@ -118,6 +118,7 @@
     "hawc.apps.summary",
     "hawc.apps.mgmt",
     "hawc.apps.materialized",
+    "hawc.apps.epiv2",
```
shapiromatron, line 121: Selectively include w/ an environment variable; we'll remove when this gets merged into main

```python
if os.getenv("HAWC_INCLUDE_EPIV2", "False") == "True":
    INSTALLED_APPS = INSTALLED_APPS + ("hawc.apps.epiv2",)
```

In the meantime, I just updated my `venv/bin/activate` or the relevant file on windows and just add this variable, so whenever i activate the enviro it's already set

hawc/apps/epiv2/admin.py
```diff
@@ -0,0 +1,71 @@
...
+    extra = 0
+
+
+class OutcomeInline(admin.TabularInline):
+    model = models.Outcome
+    extra = 0
+
+
+class StudyPopulationAdmin(admin.ModelAdmin):
+    list_display = (
```
shapiromatron, line 42: add ability to search too maybe by name and study__short_citation for now if we can

Also add some list_filters for some of the metadata on this

hawc/apps/epiv2/models.py
```diff
@@ -0,0 +1,322 @@
...
+class MeasurementType(models.Model):
+    objects = managers.MeasurementTypeManager()
+
+    description = models.CharField(max_length=128,)
+
+    def __str__(self):
+        return self.description
+
+
+class StudyPopulation(models.Model):
```
shapiromatron, line 41: For each model , we need to add a few methods to get hooked up into our views. Please add these:

```python
   def get_assessment(self):
        return self.study.get_assessment()
   def get_study(self):
        return self.study
```

And then, let's add some unit-tests to test these out to make sure they work. That way we can test the fixture that you added as well.

If you want, we can pair on getting the tests started once you're ready...

shapiromatron: Also, for StudyPopulation, we'll need a `BREADCRUMB_PARENT = "study"` b/c we'll have a pure django view for that

shapiromatron (changes requested):
Looking great so far. This is going to be a mighty addition when its done.

hawc/apps/epiv2/admin.py
munnsmunns: Can you explain the list_filters comment a bit?

hawc/main/settings/base.py
munnsmunns: Where do you stick the environment variable in the activate file?
```bash
# This file must be used with "source bin/activate" *from bash*
# you cannot run it directly

deactivate () {
    # reset old environment variables
    if [ -n "${_OLD_VIRTUAL_PATH:-}" ] ; then
        PATH="${_OLD_VIRTUAL_PATH:-}"
        export PATH
        unset _OLD_VIRTUAL_PATH
    fi
    if [ -n "${_OLD_VIRTUAL_PYTHONHOME:-}" ] ; then
        PYTHONHOME="${_OLD_VIRTUAL_PYTHONHOME:-}"
        export PYTHONHOME
        unset _OLD_VIRTUAL_PYTHONHOME
    fi

    # This should detect bash and zsh, which have a hash command that must
    # be called to get it to forget past commands.  Without forgetting
    # past commands the $PATH changes we made may not be respected
    if [ -n "${BASH:-}" -o -n "${ZSH_VERSION:-}" ] ; then
        hash -r 2> /dev/null
    fi

    if [ -n "${_OLD_VIRTUAL_PS1:-}" ] ; then
        PS1="${_OLD_VIRTUAL_PS1:-}"
        export PS1
        unset _OLD_VIRTUAL_PS1
    fi

    unset VIRTUAL_ENV
    if [ ! "${1:-}" = "nondestructive" ] ; then
    # Self destruct!
        unset -f deactivate
    fi
}

# unset irrelevant variables
deactivate nondestructive

VIRTUAL_ENV="/home/mmunns/dev/hawc/venv"
export VIRTUAL_ENV

_OLD_VIRTUAL_PATH="$PATH"
PATH="$VIRTUAL_ENV/bin:$PATH"
export PATH

# unset PYTHONHOME if set
# this will fail if PYTHONHOME is set to the empty string (which is bad anyway)
# could use `if (set -u; : $PYTHONHOME) ;` in bash
if [ -n "${PYTHONHOME:-}" ] ; then
    _OLD_VIRTUAL_PYTHONHOME="${PYTHONHOME:-}"
    unset PYTHONHOME
fi

if [ -z "${VIRTUAL_ENV_DISABLE_PROMPT:-}" ] ; then
    _OLD_VIRTUAL_PS1="${PS1:-}"
    PS1="(venv) ${PS1:-}"
    export PS1
fi

# This should detect bash and zsh, which have a hash command that must
# be called to get it to forget past commands.  Without forgetting
# past commands the $PATH changes we made may not be respected
if [ -n "${BASH:-}" -o -n "${ZSH_VERSION:-}" ] ; then
    hash -r 2> /dev/null
fi

```

shapiromatron: I just put it at the very end. Mine looks like:

```bash
tail -n 5 venv/bin/activate

export "PUBMED_API_KEY=secret-key"
export "HAWC_INCLUDE_ECO=False"
export "HAWC_INCLUDE_EPIV2=True"
```

hawc/main/settings/base.py
```diff
@@ -123,9 +123,6 @@
 if os.getenv("HAWC_INCLUDE_ECO", "False") == "True":
     INSTALLED_APPS = INSTALLED_APPS + ("hawc.apps.eco",)
 
-if os.getenv("HAWC_INCLUDE_EPIV2", "False") == "True":
```
munnsmunns: Unfortunately, because the yaml file now has epiv2 models in it, the test database can't load unless epiv2 is installed so the unit tests always fail

shapiromatron: duh, that makes sense, my bad. ok to remove; we have this on a separate branch so we're good.

Is this PR ready to review?

hawc/apps/epiv2/migrations/0001_initial.py
```diff
@@ -0,0 +1,546 @@
...
+                    ),
+                ),
+                ("created", models.DateTimeField(auto_now_add=True)),
+                ("last_updated", models.DateTimeField(auto_now=True)),
+                (
+                    "design",
+                    models.ForeignKey(
+                        on_delete=django.db.models.deletion.CASCADE,
+                        related_name="outcomes",
+                        to="epiv2.studypopulationv2",
```
shapiromatron, line 68: @munnsmunns  here for example

shapiromatron: fixed; see latest code update

tests/data/fixtures/db.yaml
```diff
@@ -366,6 +366,54 @@
   fields:
     app_label: mgmt
     model: task
+- model: contenttypes.contenttype
```
shapiromatron, line 369: remove the materialized view

tests/data/fixtures/db.yaml
```diff
@@ -711,7 +759,8 @@
     year: 2020
     version: draft
     cas: 58-08-2
-    assessment_objective: '<p>Example the health effects of Chemical C. Drafted by scientists and contractors at US EPA.</p>'
```
shapiromatron, line 714: remove things between line 714 and 1090. It's not a big deal by themselves, but often times we'll get merge conflicts with other files when there are little things sprinkled in like this.

shapiromatron (approved):
Two super minor changes requested to the db.yaml, then ok to squash/merge!

PR #568, opened 2022-01-09
htmx viewset

shapiromatron:
Create a HAWC specific parent class view which abstracts away much of the boilerplate code, designed for use with [htmx](https://htmx.org/), including:

- built in permissions checking
- automated logging and auditing of write operations

Designed with concept similar to  the django-rest-framework [ViewSet](https://www.django-rest-framework.org/api-guide/viewsets/).

Example to test: `/ani/v2/experiment/study/:id/`


shapiromatron:
Closed; merging into #564

PR #572, opened 2022-01-13
Dsstox source update

casey1173:
The webservice we relied on to fetch and store chemical data in HAWC was unexpectedly shut down over the last month. This PR replaces the old webservice with a new one.

- Updated DSSTox data sourcing from actor webservices (permanately shut down) to due to actor web services shutdown, and replaced with the `comptox.epa.gov` API services.
- Removed the resolver which could match CASRN to DTXSID; this endpoint no longer exists with the new API
- Added a new `manage.py chemicals --refresh` command which requires the webservice and updates hawc data if the response from the webservice has changed. This task can be added to a cronjob to keep the data current and constantly refreshed.


hawc/services/epa/dsstox.py
```diff
@@ -34,32 +30,15 @@ def create_from_dtxsid(cls, dtxsid: str) -> "DssSubstance":
...
+            f"https://comptox.epa.gov/dashboard-api/ccdapp2/chemical-detail/search/by-dsstoxsid?id={dtxsid}"
+        )
+        response_dict = response.json()
+        if response_dict == [] or response_dict["dsstoxSubstanceId"] != dtxsid:
             raise ValueError(f"{dtxsid} not found in DSSTox lookup")
 
-        return obj
-
-    @classmethod
-    def create_from_identifier(cls, id_: str) -> "DssSubstance":
```
shapiromatron, line 45: since this endpoint no longer works b/c it uses actor; i just removed it. I did the same in the migrations; no need to keep old code around that will never work again.

hawc/apps/assessment/management/commands/chemicals.py
```diff
@@ -0,0 +1,43 @@
...
+
+def _dict_hash(dict: Dict[str, Any]) -> str:
+    """MD5 hash of a dictionary"""
+    dhash = hashlib.md5()
+    encoded = json.dumps(dict, sort_keys=True).encode()
+    dhash.update(encoded)
+    return dhash.hexdigest()
+
+
+class Command(BaseCommand):
```
shapiromatron, line 20: minor changes to this file; i made the main `handle` method basically a way to determine which actions to apply, this is consistent with what we did with the DOI PR.  Also added some text  so we know if it's stuck.

I'm getting a really strange error on my mac, but I didn't see it in other places including our container environment:

```
SSLError: HTTPSConnectionPool(host='comptox.epa.gov', port=443): Max retries exceeded with url:
 /dashboard-api/ccdapp2/chemical-detail/search/by-dsstoxsid?id=DTXSID1037303 (Caused by
 SSLError(SSLError(1, '[SSL: UNSAFE_LEGACY_RENEGOTIATION_DISABLED] unsafe legacy
 renegotiation disabled (_ssl.c:1129)')))
```

I found this link but didn't find much more on it? https://apple.stackexchange.com/questions/435577/

frontend/assessment/components/DssToxDetailTable.js
```diff
@@ -17,11 +17,13 @@ class DssToxDetailTable extends Component {
                         <tbody>
                             <tr>
                                 <td rowSpan={5} style={{borderTop: "0px"}}>
-                                    <img
-                                        alt="chemical structure image"
-                                        style={{minWidth: 150, minHeight: 150}}
-                                        src={data.svg_url}
-                                    />
+                                    <a href={data.img_url}>
```
shapiromatron, line 20: put this in a hyperlink so you can just click to see a bigger version

hawc/apps/assessment/serializers.py
```diff
@@ -11,7 +11,7 @@
 
 class DSSToxSerializer(serializers.ModelSerializer):
     dashboard_url = serializers.URLField(source="get_dashboard_url")
-    svg_url = serializers.URLField(source="get_svg_url")
+    img_url = serializers.URLField(source="get_img_url")
```
shapiromatron, line 14: since the image is now a PNG, not an SVG, I renamed

shapiromatron (approved):
Nice job! See my comments for changes which were made.

PR #574, opened 2022-01-13
Assessment model changes

shapiromatron:
- Additions and changes to the Assessment model
  - added 'authors' field
  - 'assessment_objective' is now a required field
  - added 'creator' field (automatically filled with the user who originally created the assessment)
  - changed Boolean 'public' field to Datetime 'public_on' (automatically filled with the datetime when the assessment was originally made public)
image

- Small updates to admin view for assessments
image


hawc/apps/assessment/migrations/0030_assessment_models.py
```diff
@@ -0,0 +1,96 @@
+import django.db.models.deletion
```
shapiromatron, line 1: can these 4 migrations be changed into one? It might have to be manual, migration, not a squash. It may not be possible, I know there's some magic here, but if that's the case, just have two migrations, one for the field creates and run python operations, and one for the old column removals. you could name them `_assessment_models_part_1` and part 2, etc.

hawc/apps/assessment/forms.py
```diff
@@ -40,7 +41,11 @@ class Meta:
         model = models.Assessment
 
     def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user", None)
```
shapiromatron, line 44: is it possible to remove the default None in the kwargs pop? If so, then it makes it easier to reason about since you know self.user could never be None (ie, only one code path)

hawc/apps/assessment/forms.py
```diff
@@ -40,7 +41,11 @@ class Meta:
...
         super().__init__(*args, **kwargs)
+        if self.instance.id is None:
+            self.instance.creator = self.user
+            self.fields["project_manager"].initial = [self.user]
```
shapiromatron, line 48: this is nice

hawc/apps/assessment/migrations/0031_assessment_released_on.py
```diff
@@ -0,0 +1,69 @@
...
+    operations = [
+        migrations.AddField(
+            model_name="assessment",
+            name="released_on",
+            field=models.DateTimeField(
+                help_text="The date this assessment was released to be viewed by the general public.",
+                null=True,
+            ),
+        ),
+        migrations.RunPython(set_released_on, migrations.RunPython.noop),
```
shapiromatron, line 68: might be good to add a reverse for this one. basically

```python
Assessment.objects.filter(released_on__notnull=True).update(public=True)
```

We dont generally need to run the reverse, but it can be nice to have if they're easy like this.

hawc/apps/assessment/templates/assessment/assessment_detail.html
```diff
@@ -104,6 +104,13 @@ <h1 class="d-inline-block">{{object}}</h1>
       </tr>
       {% endif %}
 
+      {% if object.assessment_objective %}
```
shapiromatron, line 107: Change to `{% if object.authors %}`

hawc/apps/assessment/forms.py
```diff
@@ -54,9 +59,11 @@ def __init__(self, *args, **kwargs):
         self.fields["reviewers"].widget = AutoCompleteSelectMultipleWidget(
             lookup_class=HAWCUserLookup
         )
+        self.fields["public_on"].widget = widgets.DateCheckboxInput()
```
shapiromatron, line 62: better to specify in the Meta, eg., 

```python
class MyForm(forms.ModelForm):

    class Meta:
        model = MyModel
        widgets = {
            'description': forms.Textarea,
        }
```

hawc/apps/assessment/widgets.py
```diff
@@ -0,0 +1,40 @@
+from django import forms
+from django.utils import timezone
+
+
+class DateCheckboxInput(forms.widgets.Input):
```
shapiromatron, line 5: why not override the checkbox input?

shapiromatron (commented):
Looking great so far,  I'm going to test out the checkbox widget real quick to see what I can make of it

hawc/apps/assessment/migrations/0031_assessment_released_on.py
munnsmunns: good call, I overlooked that one

hawc/apps/assessment/widgets.py
munnsmunns: I thought I was going to have to replace more code than I actually ended up doing. I'll go back and change that, I'm having some formatting issues right now so that may actually fix it

hawc/apps/assessment/forms.py
munnsmunns: Ha I'm not sure... I think you wrote this code!

munnsmunns: I'll take a look at it

munnsmunns: It looks like it breaks the test_edit_permissions if I remove it

shapiromatron: ok, silly me

hawc/apps/assessment/widgets.py
```diff
@@ -0,0 +1,9 @@
+from django.forms.widgets import CheckboxInput
```
shapiromatron, line 1: can you move this to the common app instead of assessment?

munnsmunns (approved):

munnsmunns:
@shapiromatron I can't officially request a review since you started the branch

shapiromatron:
Thanks @munnsmunns ! Looking forward to review!

PR #575, opened 2022-01-15
data pivot custom sorting

shapiromatron:
Add a custom data pivot sort type for data pivots. Primarily designed for cases where alphanumeric or numeric sort isn't sufficient, for example by experimental duration `{acute, subchronic, chronic}` is not alphabetical.

Hopefully this can prevent the need for the row-level overrides since that can be hard to manage:
Screen Shot 2022-01-16 at 11 14 17 AM

Results in:
Screen Shot 2022-01-16 at 11 39 40 AM

Reuses work done in #449 for exploratory heatmaps.

PS - one technical bonus of this effort is we can progressively convert the jQuery to React a single component at a time, instead of having to go all in...


frontend/summary/dataPivot/store.js
```diff
@@ -0,0 +1,84 @@
...
+}
+
+class Store {
+    constructor(dp) {
+        this.dp = dp;
+        this.sortStore = new SortStore(this);
+    }
+    overrideRefresh = null;
+    sync() {
+        // sync state in store to global object - TODO - remove?
```
rabstejnek, line 79: Is this TODO going to remain in to be handled by future React work?

frontend/summary/dataPivot/store.js
```diff
@@ -0,0 +1,84 @@
...
+            .value();
+    }
+}
+
+class Store {
+    constructor(dp) {
+        this.dp = dp;
+        this.sortStore = new SortStore(this);
+    }
+    overrideRefresh = null;
```
rabstejnek, line 77: Remove?

rabstejnek (approved):
Looks good! My only concern is that the row overrides section is no longer forced to update (since `handleTableChange` was omitted from this new implementation). I know we're looking to slowly React-ify this though, and I see an `overrideRefresh` property in the store, so I'm assuming this is coming back in future work!

frontend/summary/dataPivot/DPFOrderTab.js
```diff
@@ -371,7 +335,7 @@ let buildFilterTable = function(tab, dp, handleTableChange) {
 
         buildFilterTable(tab, dp, handleTableChange.bind(this));
         buildFilterBooleanDiv(tab, dp, handleTableChange.bind(this));
-        buildSortingTable(tab, dp, handleTableChange.bind(this));
```
rabstejnek, line 374: Since `handleTableChange` was removed the row customization part of the form is no longer forced to refresh when order changes.

frontend/summary/dataPivot/store.js
shapiromatron: Good catch. Let me add back the override refresh mechanism to  he store. I had it working and stripped it out.

shapiromatron: Added this back; it now behaves the same as before; good catch!

frontend/summary/dataPivot/DPFOrderTab.js
shapiromatron: thanks; this is now handled just as it was before

PR #576, opened 2022-01-18
add notification error for 500 error on data cleanup

munnsmunns:
Add an error modal on the data cleanup page if an error occurs:

Screen Shot 2022-01-20 at 12 15 34 PM

To test the user interface,  I put a really long string into the study short citation on this page using the text fixture: http://127.0.0.1:8000/assessment/1/clean-extracted-data/#

### DIFF
--- a/hawc/apps/common/api/viewsets.py
+++ b/hawc/apps/common/api/viewsets.py
@@ -37,6 +37,7 @@ class CleanupFieldsBaseViewSet(
     def get_queryset(self):
         return self.model.objects.all()
 
+    # should we catch the error here?
     @action(detail=False, methods=["get"])
     def fields(self, request, format=None):
         """
--- a/hawc/apps/common/api/viewsets.py
+++ b/hawc/apps/common/api/viewsets.py
@@ -37,7 +37,6 @@ class CleanupFieldsBaseViewSet(
     def get_queryset(self):
         return self.model.objects.all()
 
-    # should we catch the error here?
     @action(detail=False, methods=["get"])
     def fields(self, request, format=None):
         """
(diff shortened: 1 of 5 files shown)


hawc/apps/common/api/viewsets.py
```diff
@@ -37,6 +37,7 @@ class CleanupFieldsBaseViewSet(
     def get_queryset(self):
         return self.model.objects.all()
 
+    # should we catch the error here?
     @action(detail=False, methods=["get"])
```
munnsmunns, line 41: Is this the right spot to catch the error so it could be used for all 500 errors with the api?

shapiromatron: I wonder if we just catch this exception here might do the trick? https://github.com/shapiromatron/hawc/blob/ec785ff7d0e6918ba0e6cdf9dd66ca56e8eaf195/hawc/apps/common/api/mixins.py#L55-L59

We already have a response of 400 that may be appropriate; the issue is probably that we're not catching a `django.db.utils.DataError` so it's not being handled

shapiromatron: if that works, we can write a test for it?

munnsmunns:
Should I add a test for this?

shapiromatron:
> Should I add a test for this?

yes please! just assert that if you give it text too long, we get a 400 error

shapiromatron:
@munnsmunns I made a few changes and think this is ready - can you confirm it works on your end? If so, ok to merge

shapiromatron (approved):

shapiromatron:
@munnsmunns bump :)

PR #577, opened 2022-01-20
make literature reference author sorting case insensitive

anthonyWillisInfotel:
Make literature reference author sorts case-insensitive, primarily so surnames like "de Veers" or "von Trapp" are sorted alongside D and V, respectively.

As an example: 
Screen Shot 2022-01-20 at 8 58 19 AM

Feedback from IARC team, they prefer that the list of references should be sorted so that names like "de Veers", "von Trapp" should be sorted next to surnames starting with D and V respectively. (Not along with V and T).


frontend/lit/constants.js
```diff
@@ -6,13 +6,13 @@ export const ReferenceStorageKey = "sortReferencesBy",
...
             ["asc", "asc"],
         ],
         AUTH_DESC: [
             "AUTH_DESC",
             "author ↓",
-            [d => d.data.authors_short, d => d.data.year],
+            [d => d.data.authors_short.toUpperCase(), d => d.data.year],
             ["desc", "asc"],
         ],
         YEAR_ASC: [
```
shapiromatron, line 18: looks good; could you please add for the YEAR_ASC and YEAR_DESC case too, so that the secondary sorts are case insensitive as well?

shapiromatron (changes requested):
One minor change requested for the year secondary sorts to be consistent as well

frontend/lit/constants.js
anthonyWillisInfotel: Ok no probs

shapiromatron (approved):

PR #582, opened 2022-01-25
start published list

shapiromatron:
Add a new view to bulk-publish content for an assessment: `/assessment/:id/published/`

image


casey1173:
@shapiromatron ready for you to review

hawc/apps/assessment/views.py
```diff
@@ -855,3 +857,92 @@ def get_context_data(self, **kwargs: Any) -> Dict[str, Any]:
...
+            "visual": apps.get_model("summary", "Visual"),
+            "datapivot": apps.get_model("summary", "DataPivot"),
+            "dataset": apps.get_model("assessment", "Dataset"),
+            "summarytable": apps.get_model("summary", "SummaryTable"),
+            "attachment": apps.get_model("assessment", "Attachment"),
+        }
+        Model = lookup.get(type)
+        if not Model:
+            raise Http404()
+        return get_object_or_404(Model.objects.filter(id=object_id))
```
casey1173, line 897: changed this to accommodate for attachments--does it matter if we do not check for the correct assessment relation?

shapiromatron: can you elaborate? I'm not following with the relation comment

casey1173: yeah, this line used to be `return get_object_or_404(Model.objects.filter(assessment=item.assessment), id=object_id)` but I took out the assessment part since attachments are not directly related to assessments

casey1173: I feel like it should function the same with only the object id but I wanted to double check

shapiromatron: without that check, couldn't someone change content across assessments? That is, they could pass an assessment id where they have write access, and an object id that they don't have access to

casey1173 (approved):

hawc/apps/assessment/views.py
shapiromatron: @casey1173 I dont think this was changed right? Dont we need some sort of permissions check to ensure that the thing you are requesting to change is part of the assessment?

shapiromatron (commented):
Nice job @casey1173 ! 

I added some tests, following the htmx pattern from your epi-v2 work, but made minor changes otherwise...

(approved!)

PR #584, opened 2022-01-28
RoB Donut overlap fix

rabstejnek:
|Before|After|
|---|---|
|image|image|

Fixed a bug on the Donut visual that caused the toolbar to overlap other elements, which arose from #560 [[ref](https://github.com/shapiromatron/hawc/pull/560/files#diff-8f425a07a8c4069f0c55e38791c6110ee2d4271973bbe15ef61f682b0ba46d31L508)].
This PR makes several other tweaks to the hide toolbar / cog functionality, namely to move the cog further into the visual's padding to reduce the risk of it being covered up and to use `display: none;` to collapse the space of the toolbar when it is hidden instead of just changing it's opacity.
Other `D3Plot` visuals (RoB heatmaps, RoB barcharts, tagtrees, data pivots, endpoint dose response plots) were checked to make sure these global changes didn't produce any other bugs.


shapiromatron (approved):
Looks great; I shifted the title back to it's original location but made no changes to the cog work, well done!

PR #585, opened 2022-01-28
Change SelectedModel migration to use Session

rabstejnek:
Fix a few regressions from #521.


rabstejnek:
@shapiromatron When debugging the error at http://127.0.0.1:8000/summary/api/data_pivot/100500898/data/, it appeared that the dose units ID didn't match between `SelectedModel`s and `Session`s where they were expected to. I suspected this must have arose during migration, and there was a discrepancy in the migration file where we were assigning `sess` to a `SelectedModel` instead of a `Session`.
This proposed solution actually fixes the error when you undo and redo the migration, but I'm passing it off to you since I haven't done a deep dive into the logic behind the migration and you would be more familiar with it!

shapiromatron (approved):
This definitely seems correct to me

shapiromatron:
Oh, just found another breaking change with BMD related work, could you take a look to identify what's going on?

http://127.0.0.1:8000/summary/data-pivot/assessment/100000037/pfbs-candidate-pods-subchronic-and-chronic-rfd/

I think it's BMD related, so it's probably ok to add this PR.

rabstejnek:
> Oh, just found another breaking change with BMD related work, could you take a look to identify what's going on?
> 
> http://127.0.0.1:8000/summary/data-pivot/assessment/100000037/pfbs-candidate-pods-subchronic-and-chronic-rfd/
> 
> I think it's BMD related, so it's probably ok to add this PR.

Yep, I made a fix that checks to make sure a `SelectedModel` actually has an associated `Model` before checking for BMD values during data export. It now renders correctly!

image

shapiromatron (approved):

PR #586, opened 2022-01-28
Build start of epi-v2 rest API

munnsmunns:
- Adds an API endpoint `epidemiology/api/design` and tests for the Design model as a starting point for epi-v2
  - This endpoint will get child models (Chemical, Exposure, Data Extraction, etc.) but can't create/update them

image


tests/hawc/apps/epiv2/test_api.py
```diff
@@ -0,0 +1,201 @@
...
+            "summary": "test epidemiology study design",
+            "study_name": "Genetic and Biomarkers study for Childhood Asthma",
+            "region": "northern Taiwan",
+            "years": "2009-2010",
+            "participant_n": 456,
+            "age_profile": ["Children and adolescents <18 yrs"],
+            "countries": ["TW"],
+        }
+
+        just_created_design_id = None
```
shapiromatron, line 52: this is a strange pattern? should this be a fixture or a setup method instead?

hawc/apps/epiv2/urls.py
```diff
@@ -1,9 +1,15 @@
...
-from . import views
+from . import api, views
+
+router = DefaultRouter()
+router.register(r"design", api.Design, basename="design")
 
 app_name = "epiv2"
 urlpatterns = [
+    path("api/", include((router.urls, "api"))),
+    # Design (similar to Study Population object from epi version 1)
```
shapiromatron, line 12: remove this comment

hawc/apps/epiv2/serializers.py
```diff
@@ -0,0 +1,157 @@
...
+    chemicals = ChemicalSerializer(many=True, read_only=True)
+    criteria = CriteriaSerializer(many=True, read_only=True)
+    exposure = ExposureSerializer(many=True, read_only=True)
+    exposure_levels = ExposureLevelSerializer(many=True, read_only=True)
+    outcomes = OutcomeSerializer(many=True, read_only=True)
+    adjustment_factors = AdjustmentFactorSerializer(many=True, read_only=True)
+    data_extractions = DataExtractionSerializer(many=True, read_only=True)
+    url = serializers.CharField(source="get_absolute_url", read_only=True)
+
+    def create(self, validated_data):
```
shapiromatron, line 93: probably best to clean this file up. I'd create a new branch "epi-api-downstream", and i'd keep things in that branch. For this branch, lets' delete them.

That one no content is lost when we're ready to address them, but for now, this is a clean commit that doesn't have non-functional commented out code

hawc/apps/epiv2/serializers.py
```diff
@@ -0,0 +1,157 @@
...
+        return super().to_internal_value(data)
+
+
+class MeasurementTypeSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.MeasurementType
+        fields = ["description"]
+
+
+class ChemicalSerializer(serializers.ModelSerializer):
```
shapiromatron, line 32: see note below; probably best to delete this in the current PR being submitted (but keep in a separate branch for future work)

shapiromatron (changes requested):
Two minor tweaks. 

1. clean this PR so that only functional code gets added. I'd create a separate branch that we can use in the future to work on the downstream
2. update the tests to have some sort of helper function instead of a `nonlocal`; generally tests should be atomic

tests/hawc/apps/epiv2/test_api.py
shapiromatron: Just curious, what strategy did you end up going with and why? This'll be useful for documentation...

munnsmunns: I decided to postpone making changes to the tests, we can make a card to remove this pattern on both epi version 1 and 2

shapiromatron: That makes sense!

hawc/apps/epiv2/serializers.py
```diff
@@ -0,0 +1,122 @@
...
+    chemicals = ChemicalSerializer(many=True, read_only=True)
+    criteria = CriteriaSerializer(many=True, read_only=True)
+    exposure = ExposureSerializer(many=True, read_only=True)
+    exposure_levels = ExposureLevelSerializer(many=True, read_only=True)
+    outcomes = OutcomeSerializer(many=True, read_only=True)
+    adjustment_factors = AdjustmentFactorSerializer(many=True, read_only=True)
+    data_extractions = DataExtractionSerializer(many=True, read_only=True)
+    url = serializers.CharField(source="get_absolute_url", read_only=True)
+
+    def create(self, validated_data):
```
shapiromatron, line 93: add a `@transaction.atomic` decorator?

hawc/apps/epiv2/serializers.py
```diff
@@ -0,0 +1,122 @@
...
+        # remove nested models from validated_data
+        for model in nested_models:
+            data = validated_data.pop(model, None)
+            nested_validated_data.update({model: data})
+
+        # execute inherited create method
+        instance = super().create(validated_data)
+
+        # add nested models to the instance
+        data = nested_validated_data.get("countries")
```
shapiromatron, line 110: suggest changing this pattern to use a walrus and removing the trailing comma

```python
if countries := nested_validated_data.get("countries"):
    instance.countries.set(countries)
```

shapiromatron (approved):
two minor tweaks then good to merge!

PR #580, opened 2022-01-24
admin dashboard - new hawc_admin application

rabstejnek:
Rewrite the admin dashboard and associated views and API endpoints into a new django application. In addition to migrating code, we added a few new dashboard views. The current views include:

1. View assessment size for all assessments for key object counts
2. View top 20 assessments with growth in the last 6 months
3. Item creation timeline for a select assessment
4. Object growth over time
5. User insights dashboard

PR #588, opened 2022-02-03
Fix: Include missing endpoints in generated OpenAPI schema

rabstejnek:
Some endpoints were being left out of the OpenAPI schema, namely those using `AssessmentLevelPermissions` where it was expecting an `assessment_id` query param to be passed through. This PR includes a subclass of the default OpenAPI generator that ignores permission checks during endpoint determination, which allows for it to capture all endpoints.

### DIFF
--- a/hawc/main/openapi.py
+++ b/hawc/main/openapi.py
@@ -1,0 +1,64 @@
+import warnings
+from urllib.parse import urljoin
+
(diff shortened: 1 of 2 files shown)


shapiromatron (approved):
I had to move things a bit after merging #580; and then updated the comments and how the patch was documented a bit. No substantial changes. If you're happy with the edits, feel free to merge!

PR #590, opened 2022-02-03
allow duplicate IDs when importing HERO references

munnsmunns:
Previously, attempting to import a list of HERO ids would create a server error if any duplicates were contained in the list. Now, duplicates are automatically removed from the list when validating the string

### DIFF
--- a/hawc/services/epa/dsstox.py
+++ b/hawc/services/epa/dsstox.py
@@ -19,7 +19,7 @@ class DssSubstance(NamedTuple):
         """Fetch a DssTox instance from the actor webservices using a DTXSID.
 
         Args:
-            dtxsid (str): a DTXSID identifer
+            dtxsid (str): a DTXSID identifier
 
         Raises:
             ValueError: if object could not be created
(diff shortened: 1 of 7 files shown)


tests/hawc/apps/lit/test_api.py
```diff
@@ -320,6 +320,11 @@ def test_success(self, db_keys):
         resp = c.post(url, payload, format="json")
         assert resp.status_code == 201
 
+        # search string with duplicates
+        payload.update(search_string="5490558, 5490558", title="second demo")
```
shapiromatron, line 324: updated to simplify the payload update

hawc/services/epa/dsstox.py
```diff
@@ -19,7 +19,7 @@ def create_from_dtxsid(cls, dtxsid: str) -> "DssSubstance":
         """Fetch a DssTox instance from the actor webservices using a DTXSID.
 
         Args:
-            dtxsid (str): a DTXSID identifer
+            dtxsid (str): a DTXSID identifier
```
shapiromatron, line 22: a little scope creep from me :( -  fixed a typo throughout where we spelled "identifer" instead of "identifier"

shapiromatron (approved):
Nice; your code changes alerted me to an existing spelling error which I think I fixed in other places too - if things look good after your review of my changes, please merge!

PR #591, opened 2022-02-04
Fix: Lit API errors when tagtree is empty

rabstejnek:
Some API endpoints were raising errors when only the root tag was remaining on an assessment. This state could be reached by doing the following:

1. A user manually deletes all of the defined tags at `/lit/assessment/{id}/tags/update/`
2. A user `POST`s an empty tagtree `{"tree":[]}` to `/lit/api/assessment/100000039/tagtree/`

The most common fix in this PR concerns accessing the `children` key on a node dump; this has been changed to return a default empty list if the `children` key doesn't exist.


hawc/apps/summary/models.py
```diff
@@ -121,27 +119,6 @@ def get_absolute_url(self):
     def get_assessment(self):
         return self.assessment
 
-    @classmethod
-    def build_report(cls, report, assessment):
```
shapiromatron, line 125: @rabstejnek I just removed this whole method; we're not using it anywhere

PR #593, opened 2022-02-07
Remove nullable property from RoB study field

rabstejnek:
`RiskOfBias` should always have an associated study, so this PR removes `null=True` from the `study` field.

Note, the accompanying migration does not handle the case where a `RiskOfBias` with a null `study` exists; if this is the case, an `IntegrityError` will be thrown and the migration not applied. This forces a user to manually deal with the problematic `RiskOfBias`es, which seems preferable to just deleting them from within the migration.


hawc/apps/riskofbias/api.py
```diff
@@ -224,11 +224,10 @@ def create_v2(self, request):
 
     @action(detail=True, methods=("patch",))
     def update_v2(self, request, *args, **kwargs):
-        partial = kwargs.pop("partial", False)
```
rabstejnek, line 227: This is leftover from the default `update` method. Since this is a `PATCH` action and not a `PUT`, though, we can assume that `partial=True`.

shapiromatron (approved):

PR #594, opened 2022-02-07
allow duplicate rows when bulk data tagging

munnsmunns:
When uploading a CSV for bulk data tags, duplicate rows are now deleted instead of raising an error


shapiromatron (approved):
Looks great; did a little code shuffling and clarified instructions. More flexible + less code!

PR #595, opened 2022-02-08
500 animation

shapiromatron:
Follow on to #587 ; adding an animation to the 500 page.


shapiromatron (commented):
This is looking really good, and it's exciting because by adding this runtime we can add new observablehq notebooks into the hawc runtime!  However, I'm seeing that it's still loading a few external resources  - can you update the vendor those as well?

Might be best to put in a new folder- `/static/vendor/observablehq/`

(I just turn off my wifi to test to ensure that I can get everything working and only loading from the server)

Could you also put the license in that folder? https://github.com/observablehq/runtime/blob/main/LICENSE

hawc/static/vendor/observablehq/epicyclic-gearing.js
```diff
@@ -0,0 +1,128 @@
...
+          );
+        }
+        path.push("M0,", -r3, "A", r3, ",", r3, " 0 0,0 0,", r3, "A", r3, ",", r3, " 0 0,0 0,", -r3, "Z");
+        return path.join("");
+      }
+    )
+  });
+  main.variable(observer("d3")).define("d3", function () {
+    return (
+      d3
```
munnsmunns, line 124: had to change this line to use the d3.min.js file instead of importing it from a cdn

munnsmunns (approved):

PR #596, opened 2022-02-08
launch review bugs

shapiromatron:
Assorted grab-bag of minor updates to fix bugs identified in release releases:

- rename "pull-right" to "float-right" - muscle memory from bootstrap [3->4 migration](https://getbootstrap.com/docs/4.0/migration/#utilities)
- fix get_absolute_url case when animal.DosingRegime parent was deleted
- case-insensitive email matching when logging in via external auth
- set a maximum cache duration of 10 days instead of unlimited; prevent unbounded cache growth
- add search to species and strain on admin
- django minor version update
- fix assessment search page error where we only filter year if search string is numeric
- improve invalid hero id list to add a space between hero ids


rabstejnek (approved):
Looks good!

PR #598, opened 2022-02-10
Epi v2 child forms

munnsmunns:
- Adds the rest of the new epi models to the design detail/read page
- Adds htmx forms for updating new epi models on design update page

### Clicking the 'Update' option under Actions on the detail page redirects to the update page
image

### Clicking the 'Update' option under Actions on the update page swaps in the design update form
image

### The design update page includes all design child objects, each with create, edit, delete, and clone capabilities
image

### Htmx is used to swap the edit and create forms inline with the table, and to immediately clone and delete within the table
image


hawc/apps/epiv2/forms.py
```diff
@@ -12,6 +15,11 @@ class DesignForm(forms.ModelForm):
 
     UPDATE_HELP_TEXT = "Update an existing study-population."
 
+    # TODO: create a new lookup
```
shapiromatron, line 18: we dont need to make another one do we?

hawc/apps/epiv2/forms.py
```diff
@@ -31,22 +39,201 @@ def helper(self):
...
+        if design:
+            self.instance.design = design
+
+        self.fields["dsstox"].widget = selectable.AutoCompleteSelectWidget(
+            lookup_class=DssToxIdLookup
+        )
+
+    @property
+    def helper(self):
+        for fld in list(self.fields.keys()):
```
shapiromatron, line 104: I know this is a pattern in other forms that are in hawc, but i'm now thinking we should be specific where possible. can we change this to specific fields instead? Just in epiv2 for now..

```python
self.fields['field_name'].widget.attrs['rows'] = 3
```

hawc/apps/epiv2/forms.py
```diff
@@ -31,22 +39,201 @@ def helper(self):
...
+            widget = self.fields[fld].widget
+            if type(widget) == forms.Textarea:
+                widget.attrs["rows"] = 3
+        helper = BaseFormHelper(self)
+        helper.form_tag = False
+        helper.add_row("name", 4, "col-md-3")
+        helper.add_row("central_tendency", 6, "col-md-2")
+        helper.add_row("neg_exposure", 3, ["col-md-3", "col-md-4", "col-md-3"])
```
shapiromatron, line 161: 3+4+3 = 10, why not just make the `col-md-4` which makes sums to 12?

hawc/apps/epiv2/views.py
```diff
@@ -11,22 +17,226 @@ class DesignCreate(BaseCreate):
...
 
 class DesignUpdate(BaseUpdate):
     success_message = "Study-population updated."
     parent_model = Study
     parent_template_name = "study"
     model = models.Design
     form_class = forms.DesignForm
+    template_name = "epiv2/design_update.html"
+
+    def get_queryset(self):
```
shapiromatron, line 32: can this be moved to a manager or queryset method so we don't have it in two different views? 
https://ccbv.co.uk/projects/Django/4.0/django.views.generic.edit/UpdateView/

hawc/apps/epiv2/views.py
```diff
@@ -11,22 +17,226 @@ class DesignCreate(BaseCreate):
...
+                Prefetch(
+                    "data_extractions",
+                    queryset=models.DataExtraction.objects.select_related(
+                        "adjustment_factor", "outcome", "exposure_level"
+                    ),
+                ),
+            )
+        )
+
+    def get_context_data(self, **kwargs):
```
shapiromatron, line 97: I wonder if it's better to do it this way, or to update templates to just use the `{{object.exposures.all}}` etc, what do you think?

hawc/apps/epiv2/views.py
```diff
@@ -11,22 +17,226 @@ class DesignCreate(BaseCreate):
...
+        return render(request, self.detail_fragment, self.get_context_data())
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["model"] = self.model.__name__.lower()
+        return context
+
+
+# Criteria viewset
+class CriteriaViewset(DesignChildViewset):
```
shapiromatron, line 197: this is brilliant, nice job!

hawc/apps/epiv2/views.py
```diff
@@ -11,22 +17,226 @@ class DesignCreate(BaseCreate):
...
+            self.perform_update(request.item, form)
+            template = self.detail_fragment
+        context = self.get_context_data(form=form)
+        return render(request, template, context)
+
+
+class DesignChildViewset(HtmxViewSet):
+    actions = {"create", "read", "update", "delete", "clone"}
+    parent_model = models.Design
+    model = None  # required
```
shapiromatron, line 144: i like this pattern

shapiromatron (commented):
Looking great so far, a few comments but no big lifts I hope.

shapiromatron:
@casey1173 @munnsmunns ok, I think I cleared up the conflicts from the branch this was forked from. This also fixed the broken github actions too, so I think it's ready for more revisions from our meeting earlier today. Nice job ya'll, looking really good.

hawc/apps/epiv2/views.py
munnsmunns: I decided to update the templates, the views look a lot cleaner that way.

PR #599, opened 2022-02-15
Form to bulk convert references to studies

munnsmunns:
Update the reference page for creating studies from references. Now, multiple references can be converted to studies at the same time by selecting the checkbox next to each item in the list, and then clicking the "Create Studies" button at the bottom of the page. Created studies will have the study type(s) which are selected before clicking the submit button.
image


hawc/apps/lit/forms.py
```diff
@@ -552,31 +552,29 @@ def clean_excel_file(self):
 
 
 class BulkReferenceStudyExtractForm(forms.Form):
-    reference_ids = forms.ModelChoiceField(
-        widget=forms.CheckboxSelectMultiple, queryset=models.Reference.objects.all(),
+    reference_ids = forms.ModelMultipleChoiceField(
+        widget=forms.CheckboxSelectMultiple, queryset=models.Reference.objects.none()
```
munnsmunns, line 556: interesting, why does this have to be none?

hawc/apps/lit/forms.py
```diff
@@ -548,3 +549,50 @@ def clean_excel_file(self):
...
         return fn
+
+
+class BulkReferenceStudyExtractForm(forms.Form):
+    reference_ids = forms.ModelMultipleChoiceField(
+        widget=forms.CheckboxSelectMultiple, queryset=models.Reference.objects.none(), required=True
+    )
+    study_type = forms.TypedMultipleChoiceField(
+        widget=forms.CheckboxSelectMultiple,
+        choices=[(1, "Bioassay"), (2, "Epidemiology"), (3, "Epi Meta"), (4, "In Vitro")],
```
shapiromatron, line 560: instead of ints, can we just use the values that are on the model? For example, 1 could be "bioassay", 2 could be "epi", etc. That way you don't need that custom build-up in the bulk_create_studies.  

You can do 

```python
for st in study_type:
    attrs.update(st=True)
```

hawc/apps/lit/forms.py
```diff
@@ -548,3 +549,50 @@ def clean_excel_file(self):
                 '"Full text URL", case sensitive.'
             )
         return fn
+
+
+class BulkReferenceStudyExtractForm(forms.Form):
+    reference_ids = forms.ModelMultipleChoiceField(
+        widget=forms.CheckboxSelectMultiple, queryset=models.Reference.objects.none(), required=True
+    )
+    study_type = forms.TypedMultipleChoiceField(
```
shapiromatron, line 558: do we need to add `required=True` here too?

hawc/apps/lit/forms.py
```diff
@@ -548,3 +549,50 @@ def clean_excel_file(self):
...
+        choices=[(1, "Bioassay"), (2, "Epidemiology"), (3, "Epi Meta"), (4, "In Vitro")],
+        coerce=int,
+    )
+
+    def clean_reference_ids(self):
+        data = self.cleaned_data["reference_ids"]
+
+        for reference in data:
+            if reference.has_study:
+                raise forms.ValidationError("A Study has already been created from this Reference.")
```
shapiromatron, line 569: maybe add the `reference.id` to the error message

tests/hawc/apps/lit/test_forms.py
```diff
@@ -336,3 +343,56 @@ def test_update_pubmed(self, db_keys):
...
+                assessment=db_keys.assessment_working,
+                reference_qs=Reference.objects.filter(assessment=db_keys.assessment_working),
+            )
+            assert not form.is_valid()
+            assert form.errors == {
+                "reference_ids": ["A Study has already been created from this Reference."]
+            }
+
+            # reference not in queryset
+            form = BulkReferenceStudyExtractForm(
```
shapiromatron, line 390: nice, this is an important test

hawc/apps/lit/views.py
```diff
@@ -484,20 +484,37 @@ def get_success_url(self):
...
-class RefListExtract(BaseList):
-    parent_model = Assessment
-    model = models.Reference
-    crud = "Update"  # update-level permission required despite list-view
+class RefListExtract(TeamMemberOrHigherMixin, MessageMixin, FormView):
     template_name = "lit/reference_extract_list.html"
     breadcrumb_active_name = "Prepare for extraction"
+    model = Assessment
+    form_class = forms.BulkReferenceStudyExtractForm
+    success_message = "Items created!"
```
shapiromatron, line 492: change to "References converted to studies!"

hawc/apps/lit/views.py
```diff
@@ -484,20 +484,37 @@ def get_success_url(self):
...
+        )
+        return kwargs
 
     def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
-        context["breadcrumbs"].insert(2, lit_overview_breadcrumb(self.assessment))
-        return context
+        kwargs.update(
+            breadcrumbs=lit_overview_crumbs(self.request.user, self.assessment, "Reference upload"),
+            object_list=models.Reference.objects.get_references_ready_for_import(self.assessment),
```
shapiromatron, line 508: we get this queryset twice, once in the form_kwargs, and once in the context_data. Is there a way that we can get it once and use it in both places?  You might be able to change this line to 

```python
kwargs.update(
    breadcrumbs=lit_overview_crumbs(self.request.user, self.assessment, "Reference upload"),
    object_list=context['form'].fields['reference_ids'].queryset,
)
```

Not sure if form is in the context yet; it might be a `self.form` or in some other location?  If it's not, then we should add it somehow to this class, so we're not doing a db double-hit.

shapiromatron (changes requested):
Looking good. Mostly minor changes; I think if we change the study_type field to a text-type, we can be more efficient with the code and make it more readable too

hawc/apps/lit/views.py
munnsmunns: I got it! I had to move object_list after calling the super get_context_data

hawc/apps/lit/forms.py
munnsmunns: I was under the impression someone may wish to create a study without a study type, but if that's not true we should change it

tests/hawc/apps/lit/test_forms.py
```diff
@@ -390,11 +390,25 @@ def test_validation_failures(self, db_keys):
...
             assert not form.is_valid()
             assert form.errors == {
                 "reference_ids": ["Select a valid choice. 4 is not one of the available choices."]
             }
+
+            # invalid study type
+            form = BulkReferenceStudyExtractForm(
+                data={
+                    "reference_ids": [Reference.objects.get(pk=db_keys.reference_unlinked)],
+                    "study_type": ["crazy"],
```
shapiromatron, line 406: 😆

hawc/apps/lit/forms.py
shapiromatron: Ahh, that makes sense; they could just want to perform study evaluation. Let's leave as-is, good thinking.

shapiromatron (approved):
well done! I made a few stylistic changes, and also de-dented your test class (it was indented into another class). 

Also globally got rid of `table-condensed` in favor of `table-sm`; we must have missed this in our migration from bootstrap 3->4.

PR #605, opened 2022-02-18
removes arrows for singular domain & metric

casey1173:
Removes up/down arrows for reordering domains and metrics when there is only one of either

### Before
image

### After
image

### After
image


frontend/riskofbias/robMetrics/DomainTable.js
```diff
@@ -58,17 +57,21 @@ class DomainTable extends Component {
                             <th>Domain #{domainIndex + 1}</th>
                             <th>{detailedView ? <DetailDomain object={domain} /> : domain.name}</th>
                             <th>
-                                <button
-                                    className="btn btn-info"
-                                    onClick={() => moveDomain(domainIndex, false)}>
-                                    <i className="fa fa-arrow-up"></i>
-                                </button>
+                                {domainCount > 1 && (
```
shapiromatron, line 60: this works fine, but elsewhere we use a ternary pattern for this instead of a chained logical. 

```javascript
{domainCount > 1 ? (
<button
        className="btn btn-info"
        onClick={() => moveDomain(domainIndex, false)}>
        <i className="fa fa-arrow-up"></i>
    </button>
) : null}
```

Suggest changing for internal consistency.

shapiromatron (changes requested):
Looks great. Really minor edit based off a style guide. 

Generally it's nice to keep these the store via an `computed` when possible b/c you just have fewer props to pass around with the components, but since we're only using this component in one place (and in the case of metrics, you can't have a computed with an input argument), this makes sense.

Welcome to our javascript world :)

shapiromatron (approved):

PR #607, opened 2022-02-22
SET data change

rabstejnek:
This PR updates SET to use a dedicated serializer for its data instead of piggybacking off of other data calls in HAWC. This allowed it to be extended and integrated more easily into SET.

-----

Added requested columns to animal bioassay SET.

image

-----

The form for one of` these new columns, "Doses", is further broken down into a "Dose unit" selection.

|No specific dose unit selected|ppm selected|
|-----|-----|
|image|image|

-----

The type selection for rows has also been fleshed out, now allowing for options other than study.

|Type "Study"|Type "Experiment"|Type "Animal group"|
|-----|-----|-----|
|image|image|image|


hawc/apps/summary/table_serializers.py
```diff
@@ -0,0 +1,304 @@
...
+        return data
+
+    @property
+    def cache_key(self):
+        return f"data-source-{self.validated_data['data_source']}-published-only-{self.validated_data['published_only']}"
+
+    @property
+    def _custom_fields(self):
+        # fields that use custom aggregation
+        if self.validated_data["data_source"] == "ani":
```
shapiromatron, line 124: use a constant for this here and elsewhere? `DataSourceChoices.Animal`?

hawc/apps/summary/table_serializers.py
```diff
@@ -0,0 +1,304 @@
...
+        if obj.dosing_regime.duration_exposure_text:
+            text = f"{text} ({obj.dosing_regime.duration_exposure_text})"
+        return text
+
+    def get_dose_dict(self, obj):
+        dose_dict = OrderedDict()
+        dgs = obj.dosing_regime.doses.all()
+        for dg in dgs:
+            if dg.dose_units.name not in dose_dict:
+                dose_dict[dg.dose_units.name] = str(dg.dose)
```
shapiromatron, line 89: this is annoying, but can we do an `int_or_float()` type thing here; so that values like 0 are shown as 0.0 instead of 0

hawc/apps/summary/table_serializers.py
```diff
@@ -0,0 +1,304 @@
...
+        # add identifying columns
+        df.insert(0, "type", model)
+        df.insert(1, "id", df[column])
+        # drop irrelevant id fields
+        return df.drop(columns=self._id_fields[self._id_fields.index(column) + 1 :])
+
+    def _get_rob_df(self):
+        study_ids = Study.objects.filter(**self._study_filters).values_list("id", flat=True)
+        rob_values = FinalRiskOfBiasScore.objects.filter(study_id__in=study_ids).values()
+        df = pd.DataFrame(
```
shapiromatron, line 273: are there only a few column renames? might be better to do something like...

```python
  df = pd.DataFrame(rob_values).rename(columns={"original": "changed"})
```

shapiromatron (changes requested):
A few minor requests, mostly trying to put some of the texty fields in constants where possible.

hawc/apps/summary/table_serializers.py
shapiromatron: ahhh, makes sense. can you add a comment to explain why you did it this way?

shapiromatron: ahhh, makes sense. can you add a comment to explain why you did it this way?

tests/data/api/api-summary-table-set-data.json
```diff
@@ -1,47 +1,4 @@
 {
   "data": [],
```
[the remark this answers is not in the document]
rabstejnek: I believe the test that generates this was written before the addition of `published_only`; I changed the test to use `published_only=False` so that actual data comes back to this record.

hawc/apps/summary/table_serializers.py
rabstejnek: Done!

shapiromatron (approved):
Nice job! I just very quickly made some pretty sharp looking tables without a lot of work. This is definitely a top highlight of our next release...

PR #608, opened 2022-02-22
February hawc client release

shapiromatron:
Mint a new version of the hawc_client library: [pypi.python.org/pypi/hawc_client](https://pypi.org/project/hawc-client/)

Added an additional change to use an authentication token from your `/user/profile` page, if the hawc server is not using django based authentication.

```python
from hawc_client import HawcClient

client = HawcClient("https://hawc.website.com")

# from /user/profile, after authenticating
client.set_authentication_token("my-secret-token")  

# fetch data from the client which may require authentication
client.epi.metadata(12345)
```


munnsmunns (approved):
Looks great! I was able to use the new authentication function with the live site and had no issues

PR #609, opened 2022-02-24
improve accession number parsing in RIS

shapiromatron:
Improve accession number parsing. Values equal to a string of "null" converted to None. Also added tests.

Users reported cases where multiple references were imported and unintentionally merged together. This was determined that the RIS file had a null value `AN  - null`. Since accession numbers are treated as strings, the null value was determined to be an identical number, and therefore references were merged together.


casey1173 (approved):
Looks good to me.

PR #610, opened 2022-02-25
Icf 365 drf add nonstandard change logging

BerkowitzMaxICF:
Call create_object_log or bulk_create_object_log for non-standard methods in the REST API.


hawc/apps/common/views.py
```diff
@@ -101,6 +103,36 @@ def create_object_log(verb: str, obj, assessment_id: int, user_id: int):
     reversion.set_comment(comment)
 
 
+def bulk_create_object_log(verb: str, obj_list: Iterable[Any], user_id: int):
+    """
+    Create an object log for each item modified in list.
+
+    Calling this method should be wrapped in a transaction. Does not associate a reversion; bulk
```
shapiromatron, line 110: added a docstring and removed the reversion check; it didn't look like it was actually used

shapiromatron (approved):
Made a number of changes, including adding tests to ensure that these logs were properly captured.

I also added transactions around most of the perform creates/updates, etc, that's out of scope for this request but I noticed it when we were going through it. 

Feel free to merge if you're ok with my changes! Great first time contribution!

PR #611, opened 2022-03-03
Paginate assessment lists

munnsmunns:
Add pagination to the public assessment list and the user's assessment list. Each page displays 50 assessments.

### DIFF
--- a/hawc/apps/assessment/forms.py
+++ b/hawc/apps/assessment/forms.py
@@ -143,6 +143,11 @@ class AssessmentFilterForm(forms.Form):
         helper.add_row("search", 2, ["col-md-8", "col-md-4"])
         return helper
 
+    def clean_order_by(self):
+        order_by = self.cleaned_data["order_by"]
+        if order_by == "":
+            raise forms.ValidationError("'' is not valid.")
+
     def get_filters(self):
         query = Q()
         if name := self.cleaned_data.get("search"):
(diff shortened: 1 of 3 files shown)


hawc/apps/assessment/forms.py
```diff
@@ -143,6 +143,12 @@ def helper(self):
         helper.add_row("search", 2, ["col-md-8", "col-md-4"])
         return helper
 
+    def clean_order_by(self):
+        order_by = self.cleaned_data["order_by"]
+        if order_by == "":
+            return "-last_updated"
```
munnsmunns, line 149: This fixes a bug that when going to the second page, order_by becomes an empty string if the user doesn't click "Apply Filters" while on the first page. Perhaps there's another way to fix this but this seems to be the most simple

shapiromatron (approved):
i accidentally dreamed code last night after our chat thinking about this one.  Here's my main change:

```python
def get_order_by(self):
    value = self.cleaned_data.get("order_by") if self.is_valid() else None
    return value or self.DEFAULT_ORDER_BY
```

The key here is an empty string is falsey, so it evaluates the OR and therefore the default order.

Also cleaned up some unused code and simplified the list view. Feel free to merge if the changes work and you're happy with the revisions.

hawc/apps/assessment/forms.py
shapiromatron: Comments were interesting on this one, wow the initial vs default isn't what I expected. https://www.peterbe.com/plog/initial-values-bound-django-form-rendered

munnsmunns: Yeah it is weird that there is no default option

PR #612, opened 2022-03-05
Convert dose table to a list

casey1173:
#### Before
image

#### After
image


shapiromatron (approved):
Looks good; I rewrote the mapping code for readability, but made no changes to the functionality.

PR #614, opened 2022-03-07
remove extra branding setting

munnsmunns:
Remove the EXTRA_BRANDING setting. All styles are now based on the HAWC_FLAVOR setting.


shapiromatron (approved):
nice!

PR #615, opened 2022-03-08
fix set focus on input forms

casey1173:
Added focus to the first input element in forms with quill, which steals focus


frontend/shared/smartTags/SmartTagEditor.js
```diff
@@ -5,6 +5,10 @@ import SmartTagContainer from "./SmartTagContainer";
 class SmartTagEditor extends SmartTagContainer {
     constructor($el, options) {
         $el.quillify();
+        if (options.takeFocusEl) {
```
shapiromatron, line 8: @casey1173 I added a new option to set focus if quill steals it... let me know what you think

shapiromatron (approved):
Looks good. Please test out the strategy I used instead; I lot of the summary visual forms steal the focus after calling `new smartTags.SmartTagEditor`, so instead of your approach I added a new option to that constructor to change focus to a different element.

If this works and you're happy with strategy, ok to merge!

frontend/shared/smartTags/SmartTagEditor.js
casey1173: Good solution!! Glad you were able to find the source of the problem with these forms

casey1173:
Nice solution; all summary forms are working as intended

PR #616, opened 2022-03-08
Update EndpointByStudyLookup.get_query

munnsmunns:
Add validation to `EndpointByStudyLookup` search fields to prevent unintended access to related models.


hawc/apps/common/lookups.py
```diff
@@ -60,9 +60,24 @@ def get_underscore_field_val(self, obj: Any, underscore_path: str) -> Any:
         Returns:
             Any: the desired attribute of the object or child object.
         """
+
+        attr_whitelist = [
```
shapiromatron, line 64: this specific list is endpoint specific; in the base class I'd make maybe rename this `lookup_whitelist`, and then put this specific list in the EndpointByStudyLookup.  Also, change it to a set instead of a list?

hawc/apps/common/lookups.py
```diff
@@ -60,9 +60,24 @@ def get_underscore_field_val(self, obj: Any, underscore_path: str) -> Any:
...
+            "last_updated",
+            "data_type",
+            "response_units",
+            "observation_time",
+            "system",
+        ]
         obj_ = obj
         try:
             for attr in underscore_path.split("__"):
+                if attr not in attr_whitelist:
```
shapiromatron, line 79: RelatedLookup is used in lots of places, but we only allow a user provided query string in `EndpointByAssessmentLookup`. 

It might be best to create a new subclass of `RelatedLookup` called `UserProvidedRelatedLookup` that has this additional check?

Also, I'd probably pull this check out of the try/except and put it up above.  Something like:

```python
        attributes = underscore_path.split("__")
        if not all(attr in self.attr_whitelist for attr in attributes):
            raise AttributeError("Access to {underscore_path} is not allowed.")

        try:
            for attr in attributes:
                obj_ = getattr(obj_, attr)
        except AttributeError:
            raise AttributeError(f"Element {underscore_path} not found in {obj}")

```

shapiromatron (changes requested):
Looking good, just a little code shuffling I think to narrow the focus of this change.

hawc/apps/common/lookups.py
shapiromatron: Thinking about this more, might be best to add a `get_search_path` method that receives the user request, and does our whitelist validation check upstream without messing with the query

munnsmunns: I'm confused. How does adding `get_search_path` make sure it's validated before `get_query`? would that method be called inside `get_query`?

shapiromatron (approved):
Looks good, generalized with a generic base class!

PR #617, opened 2022-03-08
Adds filter to rob assignment page

casey1173:
Add study search filter to the rob assignment update page:

Screen Shot 2022-03-11 at 2 55 18 PM


shapiromatron (approved):
Looks great. Updated the frontend a little bit to show the published/unpublished since we're trying to better show this to our users.

PR #619, opened 2022-03-08
metric validation in riskofbias api creation

casey1173:
Adds a validation check to the rob metric serializer; now all metric scores should have a metric_id that is in the required metrics list for the study


shapiromatron (approved):
nice job!

PR #621, opened 2022-03-09
Client checks if auth tokens are valid

munnsmunns:
Adds a check in the `client.set_authentication_token` function which raises an exception if the auth token is invalid.


shapiromatron (approved):
Looks good; added a successful client test case

PR #622, opened 2022-03-11
Build new features flags structure in hawc

shapiromatron:
Propose adding a new feature flag for enabling/disabling features during development.

The flags can be added to a standard pydantic structure in `hawc.constants`: 

```python
from pydantic import BaseModel

class FeatureFlags(BaseModel):
    THIS_IS_AN_EXAMPLE: bool = True
    ENABLE_ECO: bool = False
    ENABLE_EPI_V2: bool = False
```

Flags are available everywhere here:

```python
from django.conf import settings

assert settings.Features.THIS_IS_AN_EXAMPLE is True
```

In development environments, for most flags you can toggle them by modifying your `hawc/main/settings/local.py` file and just turning the flag on/off:

```python
# override feature flags
HAWC_FEATURES.THIS_IS_AN_EXAMPLE = True
```

In server environments, you'll save a JSON file to a environment variable, for example:

```bash
export HAWC_FEATURE_FLAGS='{"THIS_IS_AN_EXAMPLE":true}'
```

### DIFF
--- a/hawc/apps/assessment/forms.py
+++ b/hawc/apps/assessment/forms.py
@@ -195,6 +195,8 @@ class AssessmentModulesForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        if not settings.HAWC_FEATURES.ENABLE_EPI_V2:
+            self.fields["epi_version"].disabled = True
         self.fields[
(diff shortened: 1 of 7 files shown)


casey1173 (approved):

munnsmunns (approved):
Looks good!

PR #623, opened 2022-03-14
Add popup to extend expiring sessions

munnsmunns:
When a user's session cookie is within 15 minutes of expiring, they will receive a popup requesting them to click OK in order to extend the session. If they click the popup, we'll extend the current user session for an additional duration of a session. If they do not click, and then click after the session has already expired, they'll return to the homepage.


frontend/shared/startup.js
```diff
@@ -29,6 +29,26 @@ const getCookie = function(name) {
     sessionid = getCookie("sessionid"),
     csrfSafeMethod = method => /^(GET|HEAD|OPTIONS|TRACE)$/.test(method);
 
+var intervalID;
```
shapiromatron, line 32: could you move this code into a separate file,  for example, see the as an example. That way all the variable closures will a single place

```javascript
import sidebarStartup from "./utils/sidebarStartup";
```

frontend/shared/utils/checkSession.js
```diff
@@ -0,0 +1,43 @@
...
+            if (expTime - Date.now() < SESSION_EXPIRE_WARNING) {
+                const resp = confirm(
+                    "Your current session will expire in 15 minutes. Please click OK to stay logged in."
+                );
+                if (resp) {
+                    // confirmed request; cancel this interval check and make request
+                    clearInterval(checkId); // prevent popup stacking on top of each other
+                    $.post("/update-session/", {refresh: true}).done(response => {
+                        // session update failed; go home
+                        if (response.new_expiry_time === null) {
```
shapiromatron, line 26: i added this check in for cases where a user may press "ok" after the session has already expired

frontend/shared/utils/checkSession.js
```diff
@@ -0,0 +1,43 @@
+const checkSession = function() {
```
shapiromatron, line 1: did a little reorganization of this function, hopefully it reads ok

frontend/shared/utils/checkSession.js
```diff
@@ -0,0 +1,43 @@
...
+                        }
+                        setExpireTime(response.new_expiry_time);
+                        checkId = setInterval(check, CHECK_INTERVAL_MS);
+                    });
+                }
+            }
+        };
+
+    // only setup check if we have a valid session expiration
+    if (getExpireTime()) {
```
shapiromatron, line 38: added a check here; we only want to set this interval if the page reports a valid session duration

hawc/apps/assessment/views.py
```diff
@@ -666,9 +666,26 @@ def isTruthy(self, request, field):
...
+            if request.user.is_authenticated:
+                old_time = request.session.get_expiry_date().isoformat()
+                request.session.set_expiry(None)  # use the global session expiry policy
+                new_time = request.session.get_expiry_date().isoformat()
+                response = {
+                    "message": f"Session extended from {old_time} to {new_time}.",
+                    "new_expiry_time": new_time,
+                }
+            else:
+                response = {
```
shapiromatron, line 684: we hit this case if a request isn't bound to a user, which happens when a session expires

shapiromatron (approved):
I made some changes to the code but nothing major; I was able to color within the edges you set up. Nice job. Please give it a test to make sure it still performs as you'd expect, but if it works, ok to merge!

munnsmunns:
> I made some changes to the code but nothing major; I was able to color within the edges you set up. Nice job. Please give it a test to make sure it still performs as you'd expect, but if it works, ok to merge!

Looks good! You covered a couple scenarios I didn't think about.

PR #624, opened 2022-03-14
Remove the bulk update task assignment page

munnsmunns:


shapiromatron (approved):
Looks good. I followed the trail down the other API changes on the frontend and backend and remove and few more things that were exclusive to that endpoint.

PR #630, opened 2022-04-24
client side SVG rasterization

shapiromatron:
Remove phantomjs in favor of client side conversion of images to SVG and PNG format. This also removes the PPTX and PDF formats, as a PPTX can easily be generated from one of the downloaded files, and a PDF wasn't frequently used.

Resulting conversions are much quicker, and we reduced complexity significantly for a comparable build.  There were no additional frontend dependencies required.


rabstejnek (approved):
Looks good! I did notice that the border on PNGs are incomplete though:
image
I'm not sure if this was a preexisting thing, but thought it worth pointing out. I don't think it's worth blocking the PR though so it's good to merge as far as I'm concerned!

shapiromatron:
> Looks good! I did notice that the border on PNGs are incomplete though: image I'm not sure if this was a preexisting thing, but thought it worth pointing out. I don't think it's worth blocking the PR though so it's good to merge as far as I'm concerned!

Shoot, we'll definitely get comments about that. I suspect the viewport of the canvas size needs to be slightly tweaked to account for the border.

PR #636, opened 2022-07-06
admin dashboard - changes per day

shapiromatron:
Add a new admin dashboard to display some high-level graphics on django revisions per day.  Returns data for the last year (365 days).

- A github style punchcard
- A line chart of changes per day of year (w/ summary data)
- A boxplot of changes per day of week (expected lots during the week, less on weekends)

image

image
image


rabstejnek (approved):
Looks good! The only thing is the overflow from the punchcard visual looks a little off, but that probably doesn't need to be addressed since this is just an admin view.

PR #638, opened 2022-07-07
gallimaufry

shapiromatron:
A hodgepodge of assorted, minor changes to the application:

- remove out git configuration for fonts; no longer needed with #630
- fix extra space for the literature tagtree when the tree is small
    - wasted space between the tags and the literature imports, `/lit/assessment/:id/`
- add raw_id to many animal data extraction admin fields

One larger change is a fix in the determination if a risk of bias evaluation is complete. We now strip whitespace after removing tags. This is important because it appears that if all text is deleted in the quill text editor, it will save ```<p> </p>``` (whitespace between the p). Previously, this would be considered complete, that is now lo longer the case.


munnsmunns (approved):
LGTM!

PR #640, opened 2022-07-07
Add command to start postgres on windows

munnsmunns:
Adds startdb command to make.bat, which starts the postgres database for working in a dev environment.

### DIFF
--- a/make.bat
+++ b/make.bat
@@ -112,6 +112,6 @@ goto :eof
 cloc --exclude-dir=migrations,node_modules,public,private,vendor,venv --exclude-ext=json,yaml,svg,toml,ini --vcs=git --counted loc-files.txt .
 goto :eof
 
-:startdb  (start postgres database)
+:startdb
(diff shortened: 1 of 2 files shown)


shapiromatron (approved):
Fine by me; might be good to ask @rabstejnek and @casey1173  what they think, if they'd use and if so, if we should have other features or descriptions.

Should we add a line the the docs? Maybe in this section?
https://github.com/shapiromatron/hawc/blob/main/docs/source/development.rst#windows

make.bat
```diff
@@ -109,3 +111,7 @@ goto :eof
 :loc
 cloc --exclude-dir=migrations,node_modules,public,private,vendor,venv --exclude-ext=json,yaml,svg,toml,ini --vcs=git --counted loc-files.txt .
 goto :eof
+
+:startdb  (start postgres database)
```
shapiromatron, line 115: oh, maybe remove `  (start postgres database)` text too - i dont think its on other goto statements

casey1173:
Love this!!! Would definitely use.

PR #641, opened 2022-07-07
update scrub_db admin names

shapiromatron:
Update the scrubbed database admin name and password to match the test fixture database. Less context switching between using "real" data and "test" data.

The admin email/username is `admin@hawcproject.org`, and the password is now `pw`.


rabstejnek (approved):
Looks good, makes sense!

PR #642, opened 2022-07-08
epi v2 finalization

munnsmunns:
Implements requested changes from the most recent Epi group feedback

- Change "Create" button to "Add Row"
- Add Hair to biomonitoring choices
- Add "Units" field to Data extraction
- Make Source, Age category, Exposure measurement types required
- Make Sex, Type of variance estimate (assuming variance is not filled out), Lower/upper interval type (assuming lower/upper interval is not filled out), Lower/upper bound type (assuming lower/upper bound is not filled out) not required
- double length of AdjustmentFactor description field
- Add a symbols dropdown to forms for easy copy/pasting of greek symbols, super/subscripts, etc.
- Re-order Outcome fields and rename to System, Effect, Effect detail, and Endpoint/outcome
- Central tendency displays median or mean if available, does not show range in the table view

Remove the feature flag, clean up migrations, and release!


hawc/apps/epiv2/constants.py
```diff
@@ -63,6 +63,7 @@ class VarianceType(models.IntegerChoices):
 
 
 class ConfidenceIntervalType(models.TextChoices):
+    NONE = "NA", NO_LABEL
```
munnsmunns, line 66: There are a few fields like this where the scientists asked us to make them not required, but they already have an option that is "---" or "Not reported". Maybe the "---" is confusing and they think it isn't a valid option? or we should just hide the asterisk on those fields?

hawc/apps/epiv2/constants.py
```diff
@@ -50,6 +50,7 @@ class EffectEstimateType(models.TextChoices):
     SCC = "SCC", "Spearman's Correlation Coefficient"
     PC = "PC", "Percent change"
     MD = "MD", "Mean difference"
```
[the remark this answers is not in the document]
munnsmunns: For all fields with choices or just effect estimate type?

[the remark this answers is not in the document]
shapiromatron: Just this one, b/c now if you type in "other" you'll specify a full value in the table

hawc/apps/epiv2/constants.py
munnsmunns: Perhaps N/A would be a better display value for these fields?

shapiromatron: agree!

shapiromatron (changes requested):
Since this will be our release, can we also tear out the "coming soon" stuff that we previously had:

1. Remove the ENABLE_EPI_V2 feature flag
2. Update the `Study detail` page to allow for epi v2 extractions

shapiromatron (approved):
Nice job @munnsmunns and @casey1173! Excited to release this!

PR #643, opened 2022-07-08
Use form validation instead of triggering integrity error

munnsmunns:
Create a generic helper method to check unique by assessment validation, when we don't include the assessment in a model form. This helper method is applied to numerous forms via a `clean_FOO` method.

We initially tried including the assessment field in the form as a disabled hidden field, but ran into some issues with the value being hydrated correctly with a foreign key relation on updates.

- assessment.DatasetForm
- lit.SearchForm
- riskofbias.RoBDomainForm
- summary.SummaryTableForm
- summary.VisualForm
- summary.DataPivotForm

In addition, expose `non_field_errors` in the react `DjangoForm` if errors are returned in validation.


hawc/apps/summary/forms.py
```diff
@@ -1040,7 +1040,8 @@ def clean_slug(self):
 class DataPivotUploadForm(DataPivotForm):
     class Meta:
         model = models.DataPivotUpload
-        exclude = ("assessment",)
+        fields = "__all__"
+        widgets = {"assessment": forms.HiddenInput}
```
shapiromatron, line 1044: hmm, can we make this read-only so you can't change the assessment?

shapiromatron (approved):
@munnsmunns ok, made some edits and I think this is a reasonable solution that works well - as a bonus - net negative code! Please give it a review to makes sure it works as expected for creates, updates, etc.

PR #646, opened 2022-07-11
Fix misc 500 errors

rabstejnek:
`POST /lit/assessment/{assessment_id}/search/copy/`

We were using javascript to prevent the default POST behavior while defining our own custom behavior; this has been moved over to the view for handling.

`POST /summary/api/summary-text/`

We were not validating the assessment id well enough; now we rely on `EditPermissionsCheckMixin` and the underlying `AssessmentLevelPermissions` to correctly validate / check auth.


shapiromatron (approved):
Looks good, nice job. I fixed an n+1 error w/ form rendering, and moved the url logic to the form since that logic seemed like it belonged to the form instead of the view.

PR #647, opened 2022-07-12
catch RIS import errors with form validation

shapiromatron:
RIS files are messy and if improperly formatted, can lead to a bad import into HAWC. In most cases, our import form properly validates the RIS file and returns errors to the user. However, in a few edge cases (where a year is an empty string), the RIS passes validation, but then throws an error when we attempt to import references. This work properly tries to more strictly validate the file and pass exceptions to users.

We also did a little rewriting of tests, and standardized where the cleaning is done. Leaning on conventions with pydantic, our importer throws ValueError, and if found are caught by the form and cast as a ValidationError.


casey1173 (approved):
Nice, cuts down on the code a bit :)

PR #648, opened 2022-07-12
fix blank text in study evaluations

shapiromatron:
The client and server tests for determining if a study evaluation is complete were not returning the same answer; this PR corrects this.  A study evaluation is now marked complete if the following are true:

- All non-NR scores are marked as complete
- All NR scores with non-blank text are marked as complete

The default score when a new evaluation is created is NR with blank text. These two rules ensure that each score has been "touched" in one way or another

### DIFF
--- a/hawc/apps/riskofbias/migrations/0026_fix_blank_text.py
+++ b/hawc/apps/riskofbias/migrations/0026_fix_blank_text.py
@@ -1,0 +1,18 @@
+from django.db import migrations
+
+from ..constants import NR_SCORES
+
+
+def forwards_func(apps, schema_editor):
+    RiskOfBiasScore = apps.get_model("riskofbias", "RiskOfBiasScore")
+    RiskOfBiasScore.objects.filter(score__in=NR_SCORES, notes="<p> </p>").update(notes="<p>-</p>")
+
+
+def reverse_func(apps, schema_editor):
+    RiskOfBiasScore = apps.get_model("riskofbias", "RiskOfBiasScore")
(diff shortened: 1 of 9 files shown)


hawc/apps/riskofbias/migrations/0026_fix_blank_text.py
```diff
@@ -0,0 +1,18 @@
+from django.db import migrations
+
+from ..constants import NR_SCORES
+
+
+def forwards_func(apps, schema_editor):
+    RiskOfBiasScore = apps.get_model("riskofbias", "RiskOfBiasScore")
+    RiskOfBiasScore.objects.filter(score__in=NR_SCORES, notes="<p> </p>").update(notes="<p>-</p>")
```
shapiromatron, line 8: prior to #638, on the server this was considered valid b/c empty text was valid. We migrate the blank to a hyphen, so that reviews previously considered final are still considered final

hawc/apps/riskofbias/models.py
```diff
@@ -261,18 +261,18 @@ def deactivate(self):
...
-        is empty, so HTML needs to be stripped out.
+        A review is complete if all scores either:
+        1) have a score not equal to NR (default score when created)
+        2) have an NR score, but notes were added
         """
-        return all(
-            [
-                len(strip_tags(score.notes).strip()) > 0
-                for score in self.scores.all()
-                if score.score not in constants.NA_SCORES
```
shapiromatron, line 273: The logic here has now changed is is consistent with the javascript, which users were already familiar with and therefore takes precedence (and is more relaxed than the prior python rule).

rabstejnek (approved):
Looks good!

PR #649, opened 2022-07-12
remove pickles for celery serialization

shapiromatron:
Updating defaults for serialization of task arguments and outputs to use only JSON, and not pickles.  

This is the [recommended default](https://docs.celeryq.dev/en/stable/history/whatsnew-4.0.html#json-is-now-the-default-serializer) for celery since 4.0; in some deployment environments relying on pickle deserialization can be unsafe, and the option wasn't needed or used by HAWC.


munnsmunns (approved):
LGTM

PR #650, opened 2022-07-12
migrate integration tests to playwright

munnsmunns:
Implement Playwright as our engine for integration tests, replacing selenium/helium. This will make it easier to write future tests, and it runs our current tests slightly faster than before.

Also adds the command `make test-integration-debug`, which runs the integration tests in headed mode with a chrome window and a debugger.

Performance impacts, around 70s vs 130s w/ selenium:

old:
image

new:
Screen Shot 2022-07-21 at 12 50 16 PM |


shapiromatron (commented):
Nice! looking great, nice to see that net-negative in code :)

If it's working locally, but not in CI, I think you'll need to tweak the [github actions](https://github.com/shapiromatron/hawc/blob/main/.github/workflows/main.yml) file, we figured how how to run those tests in LLR if you need a reference

shapiromatron (approved):
Really nice job. I made a few edits:

1. rewrote the directly structure from `frontend/integration` to `integration`, since they're really two separate things
2. update docs, specifically the `integration tests` portion
3. used built in fixtures for setting up the playwright browser

shapiromatron:
@munnsmunns please merge if you're ok w/ my revisions!

shapiromatron:
scratch that; broken in CI.... let me take a look

shapiromatron:
ok, I think it's fixed, ok to merge @munnsmunns if you're happy...

PR #651, opened 2022-07-13
Data pivot updates

shapiromatron:
Two changes to animal data pivots:

1. Add the float-based `DosingRegime.duration_exposure` to endpoint and endpoint-group summary data pivots, with a column header of "duration exposure (days)". The `duration_exposure_text` field already exists in the data pivot, named "duration exposure", but this additional field allows for numeric sorting.
2. Added the ability to plot a data pivot barchart with the `stdev` column for error bars. Previously, we only designed the plot to use the calculated upper/lower confidence interval bars. These are absolute values. Now, if the selected column is `stdev` for confidence intervals, we assume relative difference for plotting. This could be revised in the future as an additional datapivot option if we need more customization/flexibility.

One additional change:
- update [lxml](https://github.com/lxml/lxml/blob/master/CHANGES.txt)==4.8.0--> 4.9.1


casey1173 (approved):
Nice, looks good to me.

PR #653, opened 2022-07-13
Blank IntegerInput

casey1173:
Fixes a bug that did not allow for blank IntegerInputs, i.e., users could not backspace a digit to replace with another.
Also adds a help text pop up so users know the accepted range
Found no other Integer or Float inputs to fix--this seems to be the only one that had a looping value definition


casey1173 (commented):
ignore this comment :-)

frontend/summary/summaryTable/studyEvaluationTable/Table.js
```diff
@@ -338,10 +339,10 @@ class EditColumnForm extends Component {
                                         : store.settings.columns[colIdx].width;
                             store.updateStagedColumn(colIdx, {width});
                         }}
-                        value={col.width}
-                        required={true}
```
shapiromatron, line 342: @casey1173 @rabstejnek it looks like the only change is removing the required attribute. I guess if the width is unspecified, is there any way to assume the width is 1 for the SET table?

casey1173: it only updates the table width if the input is valid, plus when you create a table the initial width value is 1. so if a user removes the width val and presses update, the width will stay 1 (or whatever it was before)

rabstejnek (approved):

shapiromatron (approved):
nice, works well

PR #654, opened 2022-07-13
Tagtree reference filtered modal fix

rabstejnek:
When viewing reference details in a tagtree visual, it used to be that `required_tags` and `pruned_tags` filtering was done on the frontend; however, since this filtering was done on paginated responses, there were times when whole pages would be returned blank, meaning the user would have to sift through all the pages just to find their filtered references.

image

This PR adds `required_tags` and `pruned_tags` as query parameters for the references API endpoint, which allows filtering to be done on the backend and for pagination to work and look as expected.

image

Other changes:
- Brought `PydanticDrfSerializer` over from LLR to validate the new more verbose query parameters that the references API endpoint could take.


hawc/apps/lit/managers.py
```diff
@@ -295,6 +295,41 @@ def with_tag(self, tag, descendants: bool = False):
...
+        Filter references by any of the given tags.
+
+        Mirrors the filtering done in tagtree visuals.
+
+        Args:
+            tags: Required tags.
+        """
+        query = Q()
+        for tag in tags:
+            query |= Q(tags__in=[tag, *tag.get_descendants()])
```
shapiromatron, line 309: haven't completely gotten through this PR yet, but I did find a case where if a user filters by two items in two different trees, it shows both. I think it should be some sort of `&=` instead of `|=`, but it's more complicated than a quick change.

shapiromatron: I tried this, and it would work, but if you're also excluding then you get an error:

```python
matches = []
for tag in tags:
    matches.append(self.filter(tags__in=[tag, *tag.get_descendants()]))
return self.intersection(*matches)
```

shapiromatron: got a little further, but now some other numbers aren't adding up

```python
def require_tags(self, tags, lazy: bool = True):
        """
        Filter references by any of the given tags.

        Mirrors the filtering done in tagtree visuals.

        Args:
            tags: Required tags.
            lazy: Do not evaluate the queryset (intersection can cause issues downstream)
        """

        matches = []
        for tag in tags:
            matches.append(self.filter(tags__in=[tag, *tag.get_descendants()]))
        qs = self.intersection(*matches)
        if lazy:
            return qs
        ids = list(qs.values_list("id", flat=True))
        return self.filter(id__in=ids)
```

hawc/apps/lit/serializers.py
```diff
@@ -486,3 +487,77 @@ def get_exporter(self):
 
     def include_descendants(self):
         return self.validated_data["nested"] == "t"
+
+
+class FilterReferences(PydanticDrfSerializer):
```
shapiromatron, line 492: this is very nice

hawc/apps/lit/managers.py
```diff
@@ -295,6 +295,41 @@ def with_tag(self, tag, descendants: bool = False):
...
+        return self.filter(query).distinct("pk")
+
+    def prune_tags(self, root_tag, pruned_tags):
```
shapiromatron, line 312: I think this is the reason we are getting different results.

The visual tagtree doesn't show the excluded nodes from the tree visual, but I don't think it filters references that have any of the tags in them. The help text currently says:

> Select tags which should be hidden from the tagtree. If a parent-tag is selected, all child-tags will also be hidden.

I think maybe we dont need this method here, and we should update the help text to state that it doesn't filter references which may include these tags?

shapiromatron (changes requested):
This is looking good so far, diagnosing a few edge cases wher multiple filters and excludes are selected.  I think I figured out most of the key issues.  There are a few changes to be made to complete. I've pushed up what I have for now, so if you'd like to continue working feel free, or I can pick it up later...

hawc/apps/lit/managers.py
shapiromatron: Per our discussion, this approach would work for the case I described, but that is not consistent with our tag tree filtering; it currently DOES do an or by all tags, but does not include descendants of the tags, which is likely why we were getting incorrect behavior. 

So, the change, as suggested by @rabstejnek, is likely to remove the descendant checking from the original QuerySet operation

rabstejnek: I parameterized `descendants` in these manager methods, and it looks like `descendants=False` for the required tags does fix the problems we were seeing.

shapiromatron (approved):
LGTM - really complex ORM logic you sorted through with this one...

PR #655, opened 2022-07-14
Django field for Quill input

rabstejnek:
# QuillField / QuillWidget

Added a new `QuillField` field and `QuillWidget` widget for use in Django forms. This removes the need to manually quillify any `textarea`s where we want HTML content, and allows for a centralized place to clean and validate incoming HTML data.

## Cleaning/validating

Cleaning is now performed on passed in HTML to resolve any rendering issues and to make sure any content doesn't look out of place within the app.

Validation is done on links, since this is something that can't be handled automatically through cleaning; if a link is given that doesn't fall within our constraints, then a validation error is given:

image

__Notes__:
- This needs to be added to epi v2.

__Bugs__:
- A console error is given when the field media is first loaded, since it's dependent on jQuery. This same error is given where Django selectable is used.
  - The reason it still works is because we also include the form media within the template, so it is presumably loaded a second time; again, this is a fix also implemented for Django selectable.


rabstejnek:
@shapiromatron I added two more commits that I believe are helpful, but can be safely reverted if you disagree.

- 8360797a36a9b4a3f87560eb5b6b95195d7c4301 brings in form media in `base.html` if the form exists; this way we don't have to remember to include it on a case by case basis, and we can be assured that it is brought in after jQuery (which most of our form media relies on)
- f8d1811d1a5d28636345d04d4f43fc4e62234e05 removes form media from being added by the crispy django tag; this was causing console errors since it brings in javascript thats dependent on jQuery without jQuery being brought in yet

hawc/apps/lit/forms.py
```diff
@@ -95,6 +92,9 @@ def helper(self):
 
 
 class ImportForm(SearchForm):
+    class Meta(SearchForm.Meta):
```
shapiromatron, line 95: wow, that's a nice trick

hawc/templates/base.html
```diff
@@ -107,6 +107,9 @@
     <script type="text/javascript" src="{% static 'vendor/jqueryui/1.10.3/jquery-ui.min.js' %}"></script>
     <script type="text/javascript" src="{% static 'vendor/htmx/1.6.1/htmx.min.js' %}"></script>
     {% render_bundle 'main' %}
+    {% if form %}
```
shapiromatron, line 110: this is very nice. we do have to watch out to make sure that we still need to add `form.media` on any htmx stuff, but that makes sense conceptually so i like this pattern

shapiromatron (approved):
Really well done! Glad to merge this one in; sorry it took so long...

frontend/shared/utils/Quillify.js
```diff
@@ -98,4 +99,7 @@ export default function() {
             q.stc.enableModals();
         }
     });
+
+    // restore original focus
+    $(focusedItem).focus();
```
shapiromatron, line 104: fyi; I added this, now I think quill doesn't mess with which form field is focused on anymore

hawc/apps/common/validators.py
```diff
@@ -36,12 +36,13 @@
 valid_css_properties = {"color", "background-color"}
 valid_scheme = {"", "http", "https"}
 valid_netloc_endings = {
-    ".gov",
     ".edu",
+    ".gov",
     ".who.int",
-    "sciencedirect.com",
+    "doi.org",
```
shapiromatron, line 42: also added doi.org as a valid URL

PR #656, opened 2022-07-15
Study Type Help Text

casey1173:
Adds help text for the study type options in the study form.
image


shapiromatron (approved):
Looks really nice, I like the inset field. I made a few changes:

- wrapped the help-text describing these fields in a form-row, and wrapped the bold span into a p, to make it a block tag instead of an inline tag
- Added padding to the entire `form-row` instead of just the first and last objects; it looks better in mobile and seemed to better match your intent

PR #657, opened 2022-07-18
Change summary table export orientation to landscape

rabstejnek:


shapiromatron (approved):
looks good; I like the height/width swap

PR #658, opened 2022-07-18
Pydantic to Django error context manager

rabstejnek:
Add a context-manager that can be used in a django application to catch pydantic value errors and throw them as standard django validation errors for use in forms or rest framework.


hawc/apps/common/helper.py
```diff
@@ -411,3 +413,23 @@ def event_plot(series: pd.Series) -> Axes:
 
     plt.tight_layout()
     return ax
+
+
+class PydanticToDjangoError:
+    def __init__(self, include_field=True, field=None, msg=None, drf=False):
```
shapiromatron, line 419: add some type annotations and docstring?

shapiromatron (changes requested):
Would it be relatively easy to add a few tests for the different permutations of how to use it? I might just make a dummy pydantic model in the test to keep it simple. TBH, I don't have a good handle of when it might come in handy, so the tests may help with that.

hawc/apps/common/helper.py
shapiromatron: also, if drf is True, does it need to throw a `rest_framework.exceptions.ValidationError` instead of a django one?  I'm not 100% sure, but they dont have the same exception inheritance

rabstejnek: DRF is pretty good about handling Django ValidationErrors, but its a better idea to explicitly throw the correct error; this has been added!

shapiromatron (approved):

PR #660, opened 2022-07-19
set base site to localhost when scrubbing db

shapiromatron:
in scrub db command, set Site to localhost to prevent admin urls going to production site

### DIFF
--- a/hawc/apps/myuser/management/commands/scrub_db.py
+++ b/hawc/apps/myuser/management/commands/scrub_db.py
@@ -3,6 +3,7 @@ from textwrap import dedent
 
 from django.contrib.auth import get_user_model
 from django.contrib.auth.hashers import make_password
+from django.contrib.sites.models import Site
 from django.core.management.base import BaseCommand, CommandError
 from django.db import transaction
 from faker import Faker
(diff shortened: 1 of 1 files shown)


hawc/apps/myuser/management/commands/scrub_db.py
```diff
@@ -33,6 +34,13 @@ def handle(self, *args, **options):
             if input("".join(message)) != "yes":
                 raise CommandError("Scrubbing user data cancelled.")
 
+        self.update_site()
+        self.update_users()
+
+    def update_site(self):
+        Site.objects.update(domain="127.0.0.1:8001", name="localhost")
```
casey1173, line 41: should this be 8000 instead of 8001? I think it's the default case

casey1173 (commented):
Looks good and works well, just one question about the url

hawc/apps/myuser/management/commands/scrub_db.py
shapiromatron: oh, yeah, wow good catch!

casey1173: okay, changed it to 8000! all good now

casey1173 (approved):

PR #662, opened 2022-07-19
Search delete warnings

shapiromatron:
Add more guards to prevent cascading delete of reference imports

1. Force user to write "delete" instead of clicking a button
2. Instead of hypothetical data loss, show exactly what will be cascade deleted if a user proceeds

Make it very clear when a user is about to remove a search that has data related to it. Instead of a hypothetical warning, show exact numbers for data removed from this deletion.

Before:
Screen Shot 2022-07-19 at 8 28 46 AM

After:
image

In addition, add guard on PubMed imports to also not delete removed references which have related studies.


casey1173 (approved):
Changed some wording slightly for personal clarity--feel free to change back if you'd like.
I also changed the comma-separated list to a semicolon-separated list since most studies have commas in the name (i.e., Butterworth, 1998, 1023693).

PR #665, opened 2022-07-20
Filter by User for RoB Assignments

casey1173:
Adds users to RoB filter page. Only shows studies with an active review by the given user: 
image

All reviews are shown for any study which meets these criteria

Additionally, added the filter form to the list view of the riskofbias reviews, in addition to the update list view, so now all team members can use.


shapiromatron:
@casey1173 ready for review. I made two big changes:

1. used a template instead of string manipulation to render the message
2. updated numbers to be exact, to show what will be deleted if a user proceeds. We're using the same calculation in the warning message is what's actually done when a user deletes.

shapiromatron (approved):

PR #666, opened 2022-07-21
normalize and lowercase all DOI 

munnsmunns:
Because DOIs are case insensitive, we now save all DOIs as lowercase. Updates the Identifiers save() function and the Reference form clean function to transform DOIs to lowercase.

In addition, ensure imports from HERO, PubMed, and RIS all import to lowercase in the future.

We also added DOI import functionality to HERO; a feature that was previously missing!


hawc/apps/lit/migrations/0016_doi_lowercase.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+
+
+from django.db import migrations
+
+
+def set_doi_lowercase(apps, schema_editor):
```
shapiromatron, line 7: does this work or something similar?

```python
from django.db.models import F
from django.db.models.functions import Lower

Identifiers.objects.filter(database=4).update(unique_id =F(Lower('unique_id'))
```

That way it's done in the db

hawc/services/nih/pubmed.py
```diff
@@ -310,6 +310,6 @@ def _get_year(cls, tree: ET.Element, dtype) -> Optional[int]:
     def _get_doi(cls, tree: ET.Element, search_string) -> Optional[str]:
         doi = tree.find(search_string)
         if doi is not None:
```
shapiromatron, line 312: maybe better to say `if doi` - that would prevent both None and empty strings

hawc/apps/lit/models.py
```diff
@@ -707,6 +707,11 @@ def existing_doi_map(cls, dois: List[str]) -> Dict[str, int]:
             ).values_list("unique_id", "id")
         }
 
+    def save(self, *args, **kwargs):
```
shapiromatron, line 710: we dont need this anymore right? i guess we can keep it b/c it doesnt hurt, but it does seem redundant

shapiromatron (changes requested):
Looking good, a few more minor requests

1. do we lowercase the dois from the RIS? I didnt see it
2. now that we're extracting DOIs w/ HERO, are DOI identifiers created when we need to?  If not, that might be a follow on task (we dont need to overload this one)

hawc/apps/lit/migrations/0016_doi_lowercase.py
munnsmunns: I'll give that a try, I'm not too familiar with the F function yet so I didn't think it would be possible to do that in the update function

munnsmunns:
> 1. do we lowercase the dois from the RIS? I didnt see it

the utils.doi.py is used for the RIS imports and for HERO imports

shapiromatron:
> > 1. do we lowercase the dois from the RIS? I didnt see it
> 
> 
> 
> the utils.doi.py is used for the RIS imports (in the Identifiers manager) and for HERO imports
> 
> 

perfect, thanks!

hawc/apps/lit/models.py
munnsmunns: Even though it's redundant, I like having it there in case a bug is introduced elsewhere, that way we don't end up creating any duplicate dois in the db (where one is lowercase and one is uppercase)

munnsmunns:
> 2. now that we're extracting DOIs w/ HERO, are DOI identifiers created when we need to?  If not, that might be a follow on task (we dont need to overload this one)

It looks like the code for extracting the DOIs from hero is already being used in the reference manager

hawc/services/nih/pubmed.py
munnsmunns: I agree, but for some reason that change totally breaks it, and no dois end up being created in the unit tests.

hawc/apps/lit/models.py
shapiromatron: sounds good, let's keep it in

hawc/apps/lit/migrations/0016_doi_lowercase.py
```diff
@@ -0,0 +1,52 @@
...
+        for ident in idents:
+            if ident.id == target.id:
+                continue
+            for ref in ident.references.all():
+                ref.identifiers.add(target)
+                ref.identifiers.remove(ident)
+            ident.delete()
+
+
+def set_doi_lowercase(apps, schema_editor):
```
shapiromatron, line 36: we have a db restraint which requires unique_id to be unique, and lowercasing was causing failures in that restraint. I had to update the migration to remove duplicates prior to the transform.

shapiromatron (approved):
Nice job! I updated the migration to fix a case I hit where the doi couldn't be lowercased b/c of a conflict with other identifiers with the same normalized unique value. If you could review my code to confirm it makes sense and works, please merge!

hawc/apps/lit/migrations/0016_doi_lowercase.py
munnsmunns: I tried it out and it worked! looks good

PR #668, opened 2022-07-22
Admin diagnostics API

rabstejnek:
Added a viewset for admin diagnostics purposes: the first being a throttle endpoint.

`/admin/api/diagnostic/throttle/`

You must be an admin in hawc to be able to access this read-only API.


shapiromatron (approved):
LGTM; changed debug to diagnostic and added tests

PR #669, opened 2022-07-22
Change Healthcheck path

casey1173:
- Moves the healthcheck out of the admin view so it's accessible regardless of firewall rules related to admin paths.
- Added new configurable setting, `HEALTHCHECK_URL_PREFIX`, which can be configured via environment variable of the same name, to add a prefix to the location of healthchecks
- Added these URLs to the SWAGGER API generated in admin


shapiromatron (approved):
Looks good! I ended up fully qualifying the `common` URL path and API endpoints, just for consistency across the application.

PR #671, opened 2022-07-25
Prevent exploratory heatmaps that are too large from rendering

rabstejnek:
The computed `matrixDataset` is the first hurdle when it comes to rendering large heatmaps: the fix here is to split the `HeatmapDatastore` constructor so that a few properties set in the constructor are used to determine if the heatmap is too large, while the `initialize` method continues the computation intensive initialization.

image

Note:
- It still hangs a little to perform the basic calculations to determine table size when the table is REALLY large, but it should not freeze the page.

### DIFF
--- a/frontend/summary/summary/ExploreHeatmap.js
+++ b/frontend/summary/summary/ExploreHeatmap.js
@@ -17,16 +17,28 @@ import {NULL_VALUE} from "./constants";
 import HeatmapDatastore from "./heatmap/HeatmapDatastore";
 
 const startupHeatmapAppRender = function(el, settings, datastore, options) {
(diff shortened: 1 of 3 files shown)


shapiromatron (approved):
Looks great to me, but I made a few changes in review that I was hoping you could take a look at.

- I tried to approach this more as a guarding method. Previously, we rendered a no dataset warning before, and I just added this below that
- I moved all the logic into the store
- I also moved the `hasDataset` and `settingsHash` logic into the store instead of the view too

One bonus change, I renamed "X fields" and "Y fields" to  "Columns" and "Rows", because that seems much more intuitive given that this is really just a big 2d table.

@rabstejnek if you're ok with my revisions, feel free to merge, else we can discuss more!

PR #674, opened 2022-07-27
Assessment values

munnsmunns:
Add two new models, both of which are scoped at the assessment-level:

- AssessmentDetail - high level additional details of assessments, commonly used for organizational users of HAWC to capture content around projects and procedures which how the HAWC assessment is integrated in this higher level project

image
image

- AssessmentValue - numerical values for the "answer" for a risk assessment, such as a reference value or reference dose if one is available. There may be multiple values (or no values) for a given assessment, and they can be scoped for cancer and/or noncancer

image
image

Assessment Detail and Values are both available for viewing on the main assessment page, and permission rules are scoped to the standard assessment visibility.

An admin-only report, `/admin/api/reports/values/`, creates a tabular report reporting assessment-level values for all assessments in HAWC.

In addition to this work, the following utility classes were added:

- Added a new URL field validator, `common.validators.validate_hyperlink`, which validates that a URL field is using a URL from a whitelisted set of websites
- Added a new URL field validator, `common.validators.FlatJSON.validate`, for checking a user-provided JSON field to ensure it only contains "flat" values, that is no, object or value arrays
- Added new pandas helper, `common.helpers.map_enum`, which adds or replaces a column with the enum value in the database with the user-label for that value.
- Moved the `optional_table_row` template tag to the common `hawc` namespace.


hawc/apps/assessment/models.py
```diff
@@ -425,6 +425,22 @@ def set_communications(self, text: str):
         Communication.set_message(self, text)
 
 
+class AssessmentValues(models.Model):
+    assessment = models.ForeignKey(Assessment, models.CASCADE, related_name="values_list")
```
shapiromatron, line 429: maybe just `values` instead?

hawc/apps/assessment/views.py
```diff
@@ -412,6 +412,7 @@ def get_context_data(self, **kwargs):
             if context["obj_perms"]["edit"]
             else context["object"].datasets.filter(published=True)
         )
+        context["assessment_values"] = context["object"].values_list.all()
```
shapiromatron, line 415: i dont think this is needed; you can just do this directly in the template

hawc/apps/assessment/templates/assessment/assessmentvalues_detail.html
```diff
@@ -0,0 +1,35 @@
+{% extends 'assessment-rooted.html' %}
+
+{% block content %}
```
shapiromatron, line 3: For now, can we revise this template to look more like the /study/:id template?

That is, there will be an Actions button in the top right to update/delete, and then the attributes of the value will be in a table, where each row is an attribute.

That way it'll be easier to add new values

hawc/apps/assessment/templates/assessment/_assessmentvalue_list.html
```diff
@@ -0,0 +1,33 @@
...
+            <th>Comments</th>
+            <th>Actions</th>
+        </tr>
+    </thead>
+    <tbody>
+        {% for object in object_list %}
+        <tr>
+            <td>{{object.value}}</td>
+            <td>{{object.comments}}</td>
+            <td>
```
shapiromatron, line 21: we need to selectively show/hide that entire column depending on permission - if you can't edit, we shouldn't show the thead either

shapiromatron (changes requested):
This is a great start. Very minimal, just what we need to get started I think.

munnsmunns:
@shapiromatron any idea what's going on in the failing tests? It's not clear to me why they're connected with this new model

shapiromatron:
> @shapiromatron any idea what's going on in the failing tests? It's not clear to me why they're connected with this new model

Agree; this doesnt look related to your work.

Yes; I think we don't have an explicit order on some of these queryset results, so sometimes we get ordering differences. Looping in @rabstejnek in case he has ideas; I think some of this were changes in related work

hawc/apps/assessment/admin.py
```diff
@@ -140,6 +140,13 @@ def migrate_terms(self, request, queryset):
     migrate_terms.short_description = "Migrate endpoint terms"
 
 
+@admin.register(models.Values)
+class ValuesAdmin(admin.ModelAdmin):
```
shapiromatron, line 144: might be good to select_related the assessment too since as it's currently designed it will need to query each assessment in the table as well

hawc/apps/assessment/models.py
```diff
@@ -1038,7 +1057,7 @@ def rendered_page(
 
 
 reversion.register(DSSTox)
-reversion.register(Assessment)
+reversion.register(Assessment, follow=["values"])
```
shapiromatron, line 1060: instead of adding this model to follow, add another row to just register this one by itself.

shapiromatron (changes requested):
A few more minor comments regarding the admin and reversion registration

hawc/apps/assessment/models.py
```diff
@@ -425,6 +425,131 @@ def set_communications(self, text: str):
...
+        choices=constants.Confidence.choices,
+        help_text="Select the overall study level confidence.",
+    )
+    species_studied = models.CharField(max_length=64, blank=True)
+    duration = models.CharField(
+        max_length=64,
+        blank=True,
+        help_text="Describe the duration of the study selected to support the derivation of the toxicity value.",
+    )
+    # TODO: figure out how to avoid circular imports/models aren't loaded yet
```
shapiromatron, line 494: https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.ForeignKey

```python
class Car(models.Model):
    manufacturer = models.ForeignKey(
        'production.Manufacturer',
        on_delete=models.CASCADE,
    )
```

hawc/apps/hawc_admin/serializers.py
```diff
@@ -0,0 +1,10 @@
+from hawc.apps.assessment.models import Values
+from hawc.apps.common.helper import FlatExport
+
+
+class ValuesExportOptions:
```
shapiromatron, line 5: i'd move this to the `assesssment.exports`; there are other common patterns you can check that follow this

hawc/apps/hawc_admin/api.py
```diff
@@ -30,3 +32,15 @@ class DiagnosticViewset(viewsets.ViewSet):
     def throttle(self, request):
         throttle = self.get_throttles()[0]
         return Response({"identity": throttle.get_ident(request)})
+
+
+class ReportsViewset(viewsets.ViewSet):
+    permission_classes = (permissions.IsAdminUser,)
+
+    @action(
+        detail=False, renderer_classes=PandasRenderers, throttle_classes=(FivePerMinuteThrottle,)
```
shapiromatron, line 41: no need for a custom throttle here

setup.cfg
```diff
@@ -19,3 +19,4 @@ python_requires = >=3.9
 [options.entry_points]
 console_scripts =
     manage.py = hawc:manage
+    hawc = hawc:manage
```
munnsmunns, line 22: dependency issue with celery - requires setuptools<59.7

hawc/apps/hawc_admin/api.py
```diff
@@ -1,9 +1,12 @@
 from django.utils import timezone
-from rest_framework import permissions, viewsets
+from rest_framework import permissions, status, viewsets
 from rest_framework.decorators import action
 from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
 
+from hawc.apps.assessment.exports import ValuesListExport
+from hawc.apps.assessment.models import Values
```
shapiromatron, line 8: keep these relative imports, consistent with the rest of hawc, here and elsewhere

shapiromatron (approved):
LGTM - please take a review and merge this in. Thanks!

munnsmunns:
> LGTM - please take a review and merge this in. Thanks!

Everything looks good! Nothing to say on the changes, the code is clear and everything works as it should.

PR #676, opened 2022-08-01
Correctly populate vocabulary terms on animal endpoint form

rabstejnek:
When validation fails on endpoint create, the previous vocabulary terms put into the form were lost. This is because it was relying on values from `form.initial` instead of `form.data`!

This fix uses the `BoundField` value instead of values from `form.initial` when rendering the form; this will get the correct value since it takes into account whether the form is bound or not.


shapiromatron (approved):
Definitely makes more sense than the initial; I didn't know about about the getter on bound forms

PR #678, opened 2022-08-02
SET choice ordering; bug fix in word reports

rabstejnek:
Added ordering for different choices in the SET form. This includes alphabetical ordering for study, experiment, and animal group labels, and domain/metric ordering for the metric choices.

image

A bug fix was also implemented; the word report now shows the correct judgment when the row is an experiment or animal group.


shapiromatron (approved):
Nice job!

PR #681, opened 2022-08-10
Treatment Related Effect

casey1173:
Creates treatment related effect for endpoint groups.
- Adds treatment effect to endpoint group model
- Can be added via endpoint update form
- Displays on endpoint/:id/ detail page
- Updated help text for significance level
- Adds column to exports

image

Results are propagated throughout the application, and can be used in data pivots. We recommend using a white-filled green up triangle, or a white-filled red down triangle to indicate treatment-related effects:

Screen Shot 2022-08-23 at 10 22 06 AM


hawc/apps/common/templates/common/helptext_popup_js.html
```diff
@@ -0,0 +1,9 @@
+<script type="text/javascript">
```
shapiromatron, line 1: this is a nice pattern

frontend/animal/EndpointTable.js
```diff
@@ -32,11 +32,18 @@ class EndpointTable {
...
+            .value();
+    }
+
+    hasData(val) {
         return _.chain(this.endpoint.data.groups)
             .map(function(d) {
                 return d[val];
             })
-            .some($.isNumeric)
+            .some(x => x != null)
```
shapiromatron, line 46: @casey1173 curious - why did this line change?

shapiromatron: can we just delete `hasData` and use `hasValues` everywhere?

shapiromatron (commented):
almost there, one minor question regarding `hasValues` vs `hasData`

frontend/animal/EndpointTable.js
casey1173: hasValues checks if all string values are numeric, and treatment related effect values are 'yes' or 'no'

shapiromatron: can we change to use `!_.isNil` and get rid of the duplicative function? https://lodash.com/docs/4.17.15#isNil

shapiromatron (approved):

PR #682, opened 2022-08-10
update default IRIS study evaluation questions

munnsmunns:
Update IRIS questions using the latest scientific version.

Bonus - add the default year to assessment creation


munnsmunns (commented):
@shapiromatron test_load_approach is failing, is it behaving how we expect it to based on the changes?

shapiromatron:
> @shapiromatron test_load_approach is failing, is it behaving how we expect it to based on the changes?

So I think there can only be one domain marked as "overall"; at least that was an idea we originally tried to enforce. That's b/c the "overall" is handled differently than the other domains. It might be good to try to collapse the animal overall and epi overall into a single domain to see if that fixes things, if you can.

shapiromatron (approved):
Looks good. I collapsed a few more domains since they were the same across animal and epi questions. Also ran into some weirdness on windows with utf8 issues; I fixed a few things, though it probably wouldn't have affected anything in production.

PR #683, opened 2022-08-16
Audit logs

rabstejnek:
Added audit logs for assessment, animal, epidemiology, risk of bias, and summary apps at the assessment level. These logs are retrieved from reversion.

The frontend has been updated to include download links for these audit logs on the assessment logs page, and reversion logs are now included on the object logs page.

|image|
|---|
|Download links are available in the Actions dropdown on the assessment logs page.|

|image|
|---|
|The table on the object logs page has been updated to include the serialized data snapshot that is stored by reversion.|

Run this command to create initial revisions as needed:

```python
manage createinitialrevisions
```

In addition, new initial revisions are created daily if they're needed via a celery task.


shapiromatron (approved):
Whew, this is another big body of work. I made a number of edits, but are happy with the final result, but I made some larger changes so your review @rabstejnek would be nice to confirm I didn't break something.

Changes include:

- Added a daily task to build any missing initial reversions
- Update permissions with API exports to make consistent with log views
- Update tabular exports, merging summary and assessment and adding more assessment details
- Moving the Change log urls to the animal app, since that seems more widely logged
- Restyle the item log list page
- Rewrite the event comparison manager. I noticed some cases where the item log list returned a list of zero items, but at least one item existed on the assessment log page. This seems to have corrected it, but this code was very tricky!

hawc/apps/assessment/managers.py
```diff
@@ -145,10 +145,59 @@ class DatasetManager(BaseManager):
...
+class Event(NamedTuple):
+    """A potentially collapsed changed event between Logs and Reversions"""
+
+    message: str
+    snapshot: str
+    user: Any
+    created: datetime
+
+
+class EventPair:
```
rabstejnek, line 157: This is a much better approach!

rabstejnek (commented):
The changes look good!

PR #684, opened 2022-08-19
Legend, show/hide node counts, and draggable nodes

casey1173:
Lit Tree Updates:
- Adds a legend for lit trees as a default
- Can be turned off in the Update or Create page for a lit tree
- Adds option to turn off lit tree node counts; defaulted to on
- Show assessment name on root nodes for custom visuals

**Legend example with no counts in nodes:**

image

**Bonus: add the ability to drag nodes to custom locations:**. One "bug" is you have to click after dragging the node to get the lines to match with the node locations.

image

**Options are configurable on all visuals, and the help text was updated:** 

image


shapiromatron (approved):
I made a few more changes to the UI and fully created the node dragging feature - could you please check it out and let me know if everything is working well for you @casey1173 ?  

I'm happy if you are - please feel free to merge!

PR #685, opened 2022-08-22
import default legend location position for rob heatmaps and barcharts

shapiromatron:
Instead of placing legends awkwardly at bottom of figure, try to calculate a default position to the right of the heatmap or barchart:

Heatmap:
image

Barchart:
image


casey1173 (approved):

shapiromatron (commented):
Nice job; one change I made was to update the migrations to use [bulk_update](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#bulk-update); originally this wasn't available but this is more performant

PR #686, opened 2022-08-30
Dual Tagging/Conflict Resolution Wireframes

casey1173:
Begin implementing conflict resolution

Wireframes for dual literature tagging and tag conflict resolution

- Adds "Search by Title/Abstract" and keyword highlighting to tagging page
image
(search nonfunctional, keywords hardcoded)

- Adds alert to Literature Review page that tag conflicts need to be resolved
image

- Adds an initial wireframe for tag conflict resolution page; has a collapsible Filter sidebar. Tag conflicts are larger and maroon while identical tags are smaller and gray.
image

- Click on a reference in this list to toggle full abstract and button links
image

- Tags can be displayed in a column (see above) or row format
image


shapiromatron:
@casey1173 - I think this has served it's purpose; should we close this one? I feel like we've implemented everything here!

PR #687, opened 2022-08-31
epiv2 assessment data export

munnsmunns:
Exports all epiv2 data of an assessment as a downloadable excel file. Each row is a Data Extraction object and all of its associated models. Replaces existing Epidemiology download link when assessment is set to Epi version 2.


shapiromatron (approved):
Nice job! Ok to merge if you're ok w/ the changes

shapiromatron:
hey @munnsmunns, I just realized that I think these exports are missing the created/last_updated timestamps - would you mind adding these in as well? After that, it's good to go from my end!

PR #688, opened 2022-08-31
FIPS cryptography compatible deployments

shapiromatron:
Enable options to deploy application with [FIPS mode](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/assembly_installing-a-rhel-8-system-with-fips-mode-enabled_security-hardening) enabled on RHEL8 (tldr; no md5 or insecure cryptographic hashing for security uses)

- Update webpack for [fips compliance](https://github.com/webpack/webpack/issues/13572)
- Add monkeypatch for django 3.2; this should be fixed with [django 4.1](https://github.com/django/django/pull/14763/) - details on [monkeypatch](https://blog.serindu.com/2019/11/12/django-in-fips-mode/) 

Bonus - add environment variables for "staging" environment settings

To enable:

```bash
export HAWC_FEATURE_FLAGS='{"FIPS_MODE":true}'
```

You should see a `FIPS monkeypatch enabled...` (using `print` b/c logging isn't enabled yet), when starting a dev server:

```text
~/dev/hawc  fips ► source venv/bin/activate && manage runserver
FIPS monkeypatch enabled...
Performing system checks...
System check identified no issues (0 silenced).
September 01, 2022 - 23:15:45
Django version 3.2.15, using settings 'hawc.main.settings.dev'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
```


rabstejnek (approved):
Looks good! For some reason it's being applied twice on my dev environment, but judging from the code it shouldn't cause any problems.

image

shapiromatron:
> Looks good! For some reason it's being applied twice on my dev environment, but judging from the code it shouldn't cause any problems.
> 
> image

Good find; that looks like can occur with django runserver. https://stackoverflow.com/questions/33814615/

I fixed with an [enviro variable](https://stackoverflow.com/a/67973819/906385), which only persists for this processes runtime environment.

PR #689, opened 2022-08-31
update developer environment entrypoint

shapiromatron:
Recent updates to setuptools and pip have caused the `manage.py` command to no longer properly function on Windows. This PR fixes that command by also aliasing it to `manage`. The HAWC documentation has now been updated to use `manage` instead of `manage.py`.

In addition, the following changes were made

- Update setuptools to build packages without `setup.py`, using [build](https://pypi.org/project/build/) instead.
- Remove pex exploration code; cannot easily be used in current production environments


munnsmunns (commented):
LGTM! had no issues on my machine

munnsmunns (approved):
LGTM

PR #690, opened 2022-09-01
hotfix - fix endpoint creates

shapiromatron:


shapiromatron:
Closed; handled server side via https://github.com/shapiromatron/hawc/pull/692

PR #691, opened 2022-09-01
update timezone to US Eastern by default

munnsmunns:
Updates the TIME_ZONE setting to match RTP local time (US/Eastern).  This can be modified by changing the `TIME_ZONE` environment variable.

In addition, define the order for rob results for SET tables (fixed test which sometimes failed)


shapiromatron (approved):
Nice job! I had a failing test on one computer, and it was related to a queryset that was unordered; so I've updated to define the order as well.

PR #692, opened 2022-09-04
hotfix  - endpoint/create/  + animal integration tests

shapiromatron:
Add a hotfix with endpoint creates, and an integration test to prevent in the future.


munnsmunns (approved):
LGTM

PR #694, opened 2022-09-06
Update IRIS study eval questions

munnsmunns:
Update default IRIS study eval questions with most recent changes as of 9/2.


shapiromatron (approved):
Nice, I looked through it and it looked really nice!

PR #695, opened 2022-09-06
fix attachment update bug

munnsmunns:
Fix bug where assessment file attachments and attachment links would not change after submitting the update form.


shapiromatron (approved):
Works great! I added an bunch of unrelated code to hopefully make the epiv2 integration test a little more resilient. fyi @munnsmunns 
 & @casey1173

PR #696, opened 2022-09-06
500-errors

munnsmunns:
Fixes several 500 errors discovered in the latest deployment of HAWC:

- Fix 500 error when clicking 'cancel' after uploading an excel for excel to json conversion.
- Fix 500 errors related to animal groups with no dosing regime. Also adds can_delete logic to prevent accidentally deleting a dosing regime that is associated with another animal group.


tests/hawc/apps/animal/test_models.py
```diff
@@ -3,6 +3,13 @@
 from hawc.apps.animal import models
 
 
+@pytest.mark.django_db
+class TestAnimalGroup:
+    def test_can_delete(self):
```
shapiromatron, line 8: would you mind writing a test here? I'm imagining you'd create a new experiment (repro), and then a parent w/ dosing regime and a child w/o. Then you'd run through these checks

```python
assert parent.can_delete() is False
assert child._can_delete() is True
child.delete()
assert parent.can_delete() is True
```

hawc/apps/animal/exports.py
```diff
@@ -115,16 +115,16 @@ def _get_data_rows(self):
...
             for i, eg in enumerate(ser["groups"]):
                 row_copy = copy(row)
+                ser_doses = ser_dosing_regime["doses"] if ser_dosing_regime else None
                 row_copy.extend(
-                    models.DoseGroup.flat_complete_data_row(
-                        ser["animal_group"]["dosing_regime"]["doses"], self.doses, i
-                    )
+                    models.DoseGroup.flat_complete_data_row(ser_doses, self.doses, i)
+                    if ser_doses
+                    else [None for _ in self.doses]
```
shapiromatron, line 127: This is what was causing the export type to look different (and fail for excel exports). Previously, the else case was returning a generator, now I wrapped it in a list so it returns a list.

hawc/apps/animal/views.py
```diff
@@ -201,6 +201,7 @@ def get_context_data(self, **kwargs):
         )
         context["config"] = dict(
             id=self.object.id,
+            has_dosing_regime=self.object.dosing_regime is not None,
```
shapiromatron, line 204: changed to set this variable in the config, rather than the html template, it should be a lot cleaner this way and is more like our "normal" path that has slowly evolved

shapiromatron (changes requested):
One more test and a quick review of my changes. Please update the description as well

tests/hawc/apps/animal/test_models.py
munnsmunns: I ended up testing the parent/child case separately from the no dosing/regime case, since the parent can still be deleted if none of its children have a dosing regime

hawc/apps/animal/templates/animal/animalgroup_confirm_delete.html
```diff
@@ -2,5 +2,10 @@
 
 {% block content %}
   {{ block.super }}
-  {% include "hawc/_delete_block.html" with name="animal-group" notes="This will remove all data associated with the animal-group, including dose-regimes, endpoints, and BMD modeling runs." %}
+
+  {% if object.can_delete %}
+    {% include "hawc/_delete_block.html" with name="animal-group" notes="This will remove all data associated with the animal-group, including dose-regimes, endpoints, and BMD modeling runs." %}
+  {% else %}
+    <div class="alert alert-danger" role="alert">Cannot be deleted. The dosign regime is associated with at least one more animal group; please delete associated animal groups before continuing.</div>
```
shapiromatron, line 9: fix typo ` dosign regime`

hawc/apps/animal/models.py
```diff
@@ -382,6 +383,12 @@ def delete_caches(cls, ids):
             Endpoint.objects.filter(animal_group__in=ids).values_list("id", flat=True)
         )
 
+    def can_delete(self) -> bool:
+        # can only be deleted if dosing regime is not associated with other animal groups
+        if not self.dosing_regime or self.dosing_regime.dosed_animals_id != self.id:
```
shapiromatron, line 388: good find!

shapiromatron (approved):

PR #697, opened 2022-09-07
one-sided relative error bars in data pivot

shapiromatron:
One sided relative error bars using `stdev` are now possible in data pivots:

Using this dataset:
image

One sided:
image

Two sided:
image


munnsmunns (approved):
Looks good!

PR #667, opened 2022-07-21
django-selectable ->  django-autocomplete-light

rabstejnek:
Replace [django-selectable](https://github.com/mlavin/django-selectable) (no longer supported) with [django-autocomplete-light](https://django-autocomplete-light.readthedocs.io/en/master/).

In addition:

- remove [jquery UI](https://jqueryui.com/)
- update [htmx] (https://github.com/bigskysoftware/htmx/blob/master/CHANGELOG.md) 1.6.1 -> 1.8.0
- write a new client-side autocomplete method now that jquery UI is removed

PR #698, opened 2022-09-09
Autocomplete Field Fix

casey1173:
Changes some of the autocomplete fields to widgets to restore requirement and help text info. When using default model settings (required, help_text, label, etc) - overriding the field requires these items to be manually set, while using just the widget allows for these to be carried properly by the django model field.

Fixes regressions in #667


shapiromatron (approved):
LGTM

PR #701, opened 2022-09-19
Make unique id for select2 components more robust

rabstejnek:
Adds a global counter that increments when a select2 component is initialized. This is used in combination with the input element's id and a randomly generated number to create a more robust unique id for each select2 component.


shapiromatron (approved):
Looks good. I changed the random number from 1k to 10k, that coupled with the field name, and the counter, I think this is pretty robust :)

PR #702, opened 2022-09-26
add ecological and other to epiv1 study designs

shapiromatron:
Add ecological and other types to epi v1, consistent with epi v2.


munnsmunns (approved):
LGTM!

PR #704, opened 2022-10-06
Replace endpoint list filter views with django-filters

shapiromatron:
Updated the list views for animal Endpoint, epi Outcome, epimeta MetaResult, and invitro IVEndpoint to use django-filters instead of our own hand rolled solution.


rabstejnek:
@shapiromatron Ready for review!

shapiromatron (commented):
Still checking this out, but I noticed that if I'm unauthenticated, I cannot filter, even on public assessments. The autocomplete fields return 403.  This needs to be fixed so that filters work even if you're not logged in.

shapiromatron:
> Still checking this out, but I noticed that if I'm unauthenticated, I cannot filter, even on public assessments. The autocomplete fields return 403. This needs to be fixed so that filters work even if you're not logged in.

I fixed this by removing the requirement that autocompletes require authentication. We added this with the dal rewrite, but it wasn't in the original spec.

shapiromatron (commented):
I made a number of changes and am really happy with how this turned out! I can assign to you to review my changes, and if so, ok to merge :)

shapiromatron (commented):
LGTM!

PR #705, opened 2022-10-17
HERO error catching; RIS flag enabled

munnsmunns:
* Handle 500 errors from HERO requests. Previously, if HERO returned out error, HAWC would also throw an error. Now we properly handle and report to the user an issue.
* Add feature flag to hide RIS import button. RIS files are not recommend when possible as we lose a lot of control on data quality when using; we can disable in some deployment environments. Users can still import an RIS if they really want to, but they'd have to navigate directly to the URL. To disable:

```bash
export HAWC_FEATURE_FLAGS='{"ALLOW_RIS_IMPORTS":false}'
```


shapiromatron (approved):
looks good!

PR #706, opened 2022-10-19
Change ROB Bulk copy API permissions

munnsmunns:
Change Risk of Bias Bulk Copy API action to require Project Manager or higher for both Assessments, instead of Admin only.


hawc/apps/riskofbias/actions/rob_clone.py
```diff
@@ -209,3 +209,17 @@ def evaluate(self) -> Dict[str, Any]:
             message=json.dumps(src_to_dst),
         )
         return {"mapping": src_to_dst}
+
+    def has_permission(self, request) -> Tuple[bool, str]:
+        """
+        Role of Project Manager or higher is required for both assessments.
+        """
+
+        src_assessment = Assessment.objects.filter(id=self.inputs.src_assessment_id).first()
```
shapiromatron, line 218: this could return None, but that would crash the if statement below

hawc/apps/riskofbias/actions/rob_clone.py
```diff
@@ -209,3 +209,17 @@ def evaluate(self) -> Dict[str, Any]:
...
+        if not src_assessment.user_can_edit_assessment(request.user):
+            return False, "Not a Project Manager or higher for source assessment."
+        dst_assessment = Assessment.objects.filter(id=self.inputs.dst_assessment_id).first()
+        if not dst_assessment.user_can_edit_assessment(request.user):
+            return False, "Not a Project Manager or higher for destination assessment."
+        else:
+            return super().has_permission(request)
```
shapiromatron, line 225: i'd remove this line out of the else; the other two statements are "guard" statements, but this should be returned regardless

shapiromatron (commented):
Requesting a few minor changes, but looks good. The change to the  `BaseApiAction` makes sense and seems appropriate.

shapiromatron (approved):
nice! tweaked language a bit, but looks good!

PR #707, opened 2022-10-19
update /new/ endpoints to /create/

munnsmunns:
Rename URLS that use /new/ to /create/ for consistency across all HAWC modules


shapiromatron (approved):
nice!

PR #711, opened 2022-10-20
Add BMDS v3.3 to database and make prior versions read only

munnsmunns:
Adds the framework for implementing BMDS v3.3 in HAWC. 

- Existing BMD Sessions are now read-only, and any existing assessments without a BMD session are set to the newest version of BMDS. 
- Adds a "Hello World" placeholder for BMDS 3 implementation in HAWC.


hawc/apps/bmd/migrations/0008_add_bmds330.py
```diff
@@ -0,0 +1,62 @@
+# Generated by Django 3.2.15 on 2022-10-20 20:24
+
+from django.db import migrations, models
+from django.db.models import F
+
+
+def update_bmd_version(apps, schema_editor, version: str):
```
shapiromatron, line 7: I rewrote to use less queries

shapiromatron (changes requested):
Looking good, a few more changes

frontend/splits.js
```diff
@@ -10,6 +10,9 @@ const startup = (name, cb) => {
         case "bmds2Startup":
             import("./bmd/bmds2/index.js").then(app => cb(app.default));
             break;
+        case "bmds3Startup":
```
shapiromatron, line 13: nice job with this; figuring this out is tricky...

shapiromatron (approved):
LGTM!

PR #712, opened 2022-10-20
SVG rasterization fixes

shapiromatron:
Add JPEG export format for downloading large visuals where compression is needed:

image

Fixed two additional issues:

1. Fixed cases where the rasterized image was scaling incorrectly (use `getBoundingClientRect`)
2. Fixed cases where extremely large canvas (>32k pixel height/width) cannot be exported because of [browser limits](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/canvas#maximum_canvas_size)


munnsmunns (approved):
Looks good!

PR #713, opened 2022-10-20
hotfix - dose-group count in data pivot

shapiromatron:
Fixed a rare edge case where the number of dose-groups for a given tabular set may be greater than the total number of dose groups for a visual, for endpoints with many dose groups but no data are extracted. This change made the code more confusing, so a comment was added to describe why it's required.


munnsmunns (approved):
LGTM

PR #716, opened 2022-10-26
Create UserReferenceTag objects

munnsmunns:
Create a UserReferenceTag object when applying tags to references. When conflict resolution is disabled the tags are immediately applied to the reference, otherwise a NotImplementedError is thrown.

### DIFF
--- a/hawc/apps/lit/models.py
+++ b/hawc/apps/lit/models.py
@@ -868,6 +868,15 @@ class Reference(models.Model):
 
     BREADCRUMB_PARENT = "assessment"
 
+    def update_tags(self, tag_pks, user):
+        user_ref_tags, created = UserReferenceTag.objects.get_or_create(reference=self, user=user)
+        user_ref_tags.tags.set(tag_pks)
+        user_ref_tags.save()
+        if not self.assessment.literature_settings.conflict_resolution:
+            self.tags.set(tag_pks)
+        else:
+            raise NotImplementedError
+
     def get_absolute_url(self):
         return reverse("lit:ref_detail", args=(self.pk,))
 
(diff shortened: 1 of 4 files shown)


hawc/apps/lit/models.py
```diff
@@ -868,6 +868,15 @@ class Reference(models.Model):
 
     BREADCRUMB_PARENT = "assessment"
 
+    def update_tags(self, tag_pks, user):
+        user_ref_tags, created = UserReferenceTag.objects.get_or_create(reference=self, user=user)
```
shapiromatron, line 872: can we use `update_or_create`? it would require fewer db hits instead of getting, changing, and then updating I think?

hawc/apps/lit/models.py
```diff
@@ -868,6 +868,15 @@ class Reference(models.Model):
...
+        user_ref_tags.tags.set(tag_pks)
+        user_ref_tags.save()
+        if not self.assessment.literature_settings.conflict_resolution:
+            self.tags.set(tag_pks)
+        else:
+            raise NotImplementedError
```
shapiromatron, line 878: I know it wasn't on the JIRA card, but do you think you could take a crack at writing this? 

- Query the UserReferenceTag and get all reviews for this reference. 
- If <2 reviews exist, return; no change
- Get a set of tag ids from all reviews; if all sets of tag ids are not the same; no change
- If all tags are the same, save tags on reference and update last_updated

That's probably what we need for now. I'm thinking ahead though, and when we add a view to find all conflicts, this step would be really expensive. Therefore, I wonder if we may end up adding another field to UserReferenceTag for `resolved` might be good, where basically it's set to False until this operation is complete...

shapiromatron (commented):
You knocked this out! Think we can take this a bit further before merging?

In addition, do you think updating the fixture, creating a new assessment with the screening resolution enabled and three references could be added? One with no reviews, One with resolved reviews, and one with conflicting reviews?

shapiromatron (approved):
Looks good! I did some rewrites, assuming that there are conflicts whenever a new instance is created instead of assuming there are not conflicts. Please give it a thorough review and make sure things look ok to you, but good from my end to merge!

PR #718, opened 2022-10-26
add feature flag to prevent account creation

shapiromatron:
For some deployments, anonymous users shouldn't be able to register and create an account. This PR adds a new feature flag where if enabled, the registration pages are no longer available.


rabstejnek (approved):
Looks good!

PR #719, opened 2022-10-27
Replace reference and study filters with django-filter filtersets

rabstejnek:
Swapped in django-filter on the following pages:
- `/study/assessment/<pk>/`
- `/lit/assessment/<pk>/references/search/`
- `/rob/assessment/<pk>/study-assignments/`
- `/rob/assessment/<pk>/study-assignments/updates/`


shapiromatron:
@rabstejnek is this one ready? it looks like it is from the merge status and tests passing, but wanted to see b/c it's still in draft

rabstejnek:
> @rabstejnek is this one ready? it looks like it is from the merge status and tests passing, but wanted to see b/c it's still in draft

Yep, its ready for review!

hawc/apps/lit/api.py
```diff
@@ -268,27 +268,6 @@ def update_reference_metadata_from_hero(self, request, pk):
         )
         return Response(status=status.HTTP_204_NO_CONTENT)
 
-    @action(
-        detail=True,
-        methods=("post",),
-        url_path="reference-search",
-        permission_classes=(AssessmentReadPermissions,),
-    )
-    def reference_search(self, request, pk):
```
shapiromatron, line 277: I ended up removing this API and the things around it; the only place we were using it was the reference search, and now that's a pure django view. We can potentially add back in the lit sprint as we need to, but for now, since it's not being used, I removed it.

shapiromatron (approved):
Looks great! Ok to merge if you're happy w/ my changes.

PR #721, opened 2022-11-04
Stub lit views

casey1173:


shapiromatron (approved):
Looks good!  I will add a card from this to rewrite how we handle these `TeamMemberOrHigherMixin` type cases where we just want to change the permission check on a single view; the current way we have to do it seems a little strange...

PR #722, opened 2022-11-07
Removed OrderedDict

rabstejnek:
Native dictionaries were guaranteed to preserve order in [python 3.6](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep520), and this was introduced as part of the language specification in [python 3.7](https://mail.python.org/pipermail/python-dev/2017-December/151283.html). Therefore, remove [OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict).


shapiromatron (approved):
strange code around old patterns, glad this is gone!

PR #723, opened 2022-11-07
Replace type hints with their built-ins

rabstejnek:
Replaced List, Dict, Tuple, and Set from the `typing` module with their built-ins: https://peps.python.org/pep-0585/


shapiromatron (approved):
Looks good! We'll need to check in a few months to make sure that our `next` and `literature-conflict-resolution` branches don't add any of these back; I will add task to JIRA

PR #725, opened 2022-11-07
Use regex for test that involves a content type ID

munnsmunns:
Replaces a hard coded value for content type ID in a test with a regex to make that test more resilient.

### DIFF
--- a/tests/hawc/apps/riskofbias/test_models.py
+++ b/tests/hawc/apps/riskofbias/test_models.py
@@ -1,3 +1,4 @@
+import re
 from textwrap import dedent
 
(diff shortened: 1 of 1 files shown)


shapiromatron (commented):
If you change the CT in `api-visual-barchart.json` and api-visual-rob-heatmap.json`, as will happen in the future, this will fail, is that correct?

image

I think we need to update the things that compare the JSON for this and the others in the screenshot to replace `obj_ct=\d+` with `obj_ct=9999` on the write to disk as well. That way when the obj_ct changes the ct will always be the same...

PR #726, opened 2022-11-08
style fixes

casey1173:
Fixes various style issues, most of which were caused by the new epa style sheet. 
Before and after:
- Restore background color for SET tables
image
image

- Remove bottom margin on li elements
image
image

- Maintain original link color for visited links
image
image

- Remove bottom margin for all p elements
image
image
- This centers breadcrumb text
image
image

- Increase min page size from 50vh to 95vh to fix a bug on new assessments/short pages where the action button dropdown was cropped
image
image

- Add margin to page bottom to ensure footer is properly separated from page info
image
image

- Ensure all action button dropdown links are the same color when active (i.e. clicked)
image
image

- Restore previous size and style of submit/apply filter buttons to match cancel/reset buttons
image
image

- Fix management dashboard header alignment
image
image


shapiromatron (approved):
nice, changes look good! A lot of these things were bothering me too, thanks for finding the right way to fix them!

PR #728, opened 2022-11-08
Add eslint plugin for sorting imports

munnsmunns:
`make format` and `make format-js` now sort imports in javascript files.


frontend/.eslintrc
```diff
@@ -55,7 +57,8 @@
     "plugins": [
         "babel",
         "react",
-        "prettier"
+        "prettier",
+        "simple-import-sort"
```
munnsmunns, line 61: It turns out the built in sort-imports rule [won't fix anything automatically](https://github.com/eslint/eslint/issues/11542), so a plugin is required

shapiromatron: very cool!

PR #729, opened 2022-11-09
restore dose-unit endpoint values to endpoint list

shapiromatron:
#704 simplified how endpoint list views were generated, but we removed the JS implementation of the views which simplified the data which was available for the view. This PR restores the original view, where dose-unit specific data (NOEL, LOEL, BMD, and BMDL) are presented for each endpoint. 

We implement by a custom database query, so we don't need to build the whole JS instance of the object, but instead query for what's needed. The end result is a faster view, though we lose the desirable caching properties of the serialized objects.


munnsmunns (approved):
Looks good Andy!

rabstejnek (approved):

PR #730, opened 2022-11-10
update github-actions

shapiromatron:
Assorted updates for refreshing the github actions, which haven't been changed in a while:

- [x] Update from Ubuntu 20.04 LTS to 22.04 LTS.
- [x] Update github actions to latest versions, often changing v2 to v3
- [x] Investigate using postgres installed on system instead of container
- [x] fix reported deprecations

---

This PR continues to use a docker-container instead of the installed postgres on the instance for two reasons. The first, is the installed version is more recent than our target version (14 vs 12). The second reason is it reduced container startup time from around ~23 to ~18 seconds, but then it took an additional 8 seconds to start the local service, so there wasn't much time savings. For completeness, here's the additional step in case we want to revisit in the future:

```yaml
- name: setup postgres
  run: |
    sudo systemctl start postgresql.service
    sudo -u postgres psql --command="CREATE USER hawc PASSWORD 'password'" --command="\du"
    sudo -u postgres createdb hawc-test --owner=hawc
    sudo -u postgres psql --command="\l"
```

And the environment variables are important, I'm not really sure why:

```yaml
- name: test
  env:
    DJANGO_DB_NAME: hawc-test
    DJANGO_DB_USER: hawc
    DJANGO_DB_PW: password
  run: |
    py.test --vcr-record=none
```

---

One deprecation could not be fixed -> actions/cache needs update before we can use:

```yaml
# old
run: echo "{dir}={$(yarn cache dir)}" >> $GITHUB_OUTPUT
# new
run: echo "::set-output name=dir::$(yarn cache dir)"
```        

- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


rabstejnek (approved):
Looks good!

PR #731, opened 2022-11-10
hover tag animation

casey1173:
- Changes the hover style for literature tags on the screening page.
- Corrects vertical alignment for the plus/minus expansion buttons on the tagtree
- Swaps out a custom css class for a bootstrap class

Before:
image

After:
image


shapiromatron (approved):
The animation is really smooth and looks good. Bumping up that little plus sign is really nice too, and the way you've implemented animation is really well done. My concern is that it may be distracting to people when they're focusing on assigning the tags. 

We can try it out and see what people think. 

I ended up removing the animation and liked it if we bump up `--bg-h: 4px;` to make it a little stronger w/o the animation.

shapiromatron:
Also, I don't think it' s a dealbreaker, but saw a few cases where if the tag is long it overlaps:

image

PR #732, opened 2022-11-12
refactor assessment permissions in django views and API

shapiromatron:
There were currently many patterns for checking permissions to view a page, edit content, or interact with the API. This PR begins consolidating permission checking approaches.

For django based views, whenever we can inherit permissions from an Assessment, we can now inherit from of the common base views in HAWC, instead of adding mixins or custom permission checks. These mixins are now removed:

- LegacyAssessmentAdapterMixin
- TeamMemberOrHigherMixin
- ProjectManagerOrHigherMixin
- CanCreateMixin

Renamed this method, it had a misleading name:

- Assessment.user_is_part_of_team -> Assessment.user_is_reviewer_or_higher

Renamed this method in the hawc client, to be consistent with API:

- riskofbias.data -> riskofbias.export
- riskofbias.full_data -> riskofbias.full_export


hawc/apps/assessment/api/pagination.py
```diff
@@ -0,0 +1,5 @@
+from rest_framework.pagination import PageNumberPagination
+
+
+class DisabledPagination(PageNumberPagination):
```
shapiromatron, line 4: move to common?

rabstejnek: Done!

PR #734, opened 2022-11-15
literature tag conflict resolution view

munnsmunns:
Add view for resolving literature tag conflicts.

Todos/issues:

- Creates a lot of queries due to being unable to prefetch tag names
- Should the api endpoint ignore non-htmx requests or handle them differently?
- Tagging a reference that already has tags would replace the existing tags, no option to add to the tags instead
- When adding tags after a conflict has already been resolved, the resolved user tags are shown as well as the unresolved ones

Reference with some conflicting tags, and some the same:
image

Reference with existing tags, and a new tag being applied by a user:
image


rabstejnek:
Added a new filter that allows a user to choose which tag types to filter against:
> image

This new filter has been added to existing reference filter forms, along with the new filter form on the conflict resolution view:
|Collapsed filter form|Expanded filter form|
|-----|-----|
|image|image|

Note: on filter forms where conflict resolution is not enabled for an assessment, this new filter is disabled:
|Reference search filter form where conflict resolution is not enabled|
|-----|
|image|

hawc/apps/lit/api.py
```diff
@@ -385,3 +386,20 @@ def tag(self, request, pk):
...
+        template = "lit/_reference_tag_conflict.html"
+        errors = None
+        ref = self.get_object()
+        assessment = ref.assessment
+        if assessment.user_can_edit_object(self.request.user):
+            try:
+                user_tag_id = request.POST.get("user_tag_id")
+                ref.resolve_user_tag_conflicts(user_tag_id)
+                template = "lit/_conflict_resolved.html"
+            except Exception:
```
shapiromatron, line 401: global catch all exceptions are generally a pattern we should avoid if possible

shapiromatron: I simplified this to just throw 403 or 404 instead of nicely showing exceptions - these issues wouldn't be shown by users in general

PR #735, opened 2022-11-15
User tag screening

casey1173:
Adds user tag distinction to screening view
- Resolved tags are shown as solid when conflict resolution is enabled; when disabled, it is the default tag style
- User tags are shown with dotted borders when conflict resolution is enabled; you will only see your own user tags
- Users can no longer remove "resolved" tags when conflict resolution is enabled

TODO:
- Decide how to manage resolved tag removal (currently no way to remove resolved tags)
- Fix resolved tag overwrite; i.e., when tags are marked as resolved and moved to reference model tags, previously resolved tags are removed
- Decide if Browse page should also show references tagged under user tags when filtering
- Decide if reference detail pages should show user tags
image


frontend/lit/components/Reference.js
```diff
@@ -155,6 +155,7 @@ class Reference extends Component {
                 {data.abstract ? (
                     <div
                         className="abstracts resize-y p-2"
+                        style={data.abstract.length > 1500 ? {height: "45vh"} : null}
```
shapiromatron, line 158: that's nice - so it only sets a bound if it's a long abstract?

hawc/apps/lit/managers.py
```diff
@@ -351,6 +352,18 @@ def prune_tags(self, root_tag, pruned_tags, descendants: bool = False):
         )
         return self.exclude(query).distinct("pk")
 
+    def user_tags(self, user_id: int) -> dict[int, list[int]]:
```
shapiromatron, line 355: I changed how we're building this dictionary for the serializer since it looked like the old way required a query for each reference if user references were found

hawc/apps/lit/templates/lit/overview.html
```diff
@@ -107,7 +107,7 @@ <h3>Manually added references</h3>
 
   window.app.startup("litStartup", function(lit){
     let tagtree = new lit.TagTree(config.tags[0], config.assessment_id, null);
-    tagtree.render(document.getElementById('tags'))
+    tagtree.render(document.getElementById('tags'), {handleTagClick: (x) => {window.location.href = x.get_list_link()}})
```
shapiromatron, line 110: nice!

hawc/apps/lit/views.py
```diff
@@ -286,14 +286,19 @@ def get_context_data(self, **kwargs):
         return context
 
     def get_app_config(self, context) -> WebappConfig:
+        references = [ref.to_dict() for ref in context["object_list"]]
```
shapiromatron, line 289: ... and we add to the dictionaries here instead of in the `to_dict()` view

shapiromatron (approved):
LGTM - I changed how we fetch user tags. switching to use [ArrayAgg](https://docs.djangoproject.com/en/4.1/ref/contrib/postgres/aggregates/#django.contrib.postgres.aggregates.ArrayAgg), but in the future when we update to django 4, using [ArraySubquery](https://docs.djangoproject.com/en/4.1/ref/contrib/postgres/expressions/#arraysubquery-expressions) might be a better way to go. I suspect we can do lots of other QS optimizations with that's available.

frontend/lit/components/Reference.js
casey1173: yeah that way longer abstracts aren't overwhelming on page load, but users can still expand to show the full abstract if they wish. and there's no weird oversized box with shorter abstracts

PR #736, opened 2022-11-15
Tagging filters

rabstejnek:
Added filtering to the reference tagging view.

| On page load the accordion is collapsed |
|-----|
|image|

| When toggled, the filter form is shown |
|-----|
|image|

***Other changes***:
- All tagging views have been condensed down into one view, since filtering now covers their individual edge cases
- `include_descendants` has been added as a filter, which allows tag filtering to include descendants
  - This used to be baseline behavior, but the now deleted `TagByTag` view did **not** include descendants, so this flag was added to allow us to replicate this view using filtering
- `search` and `untagged` have been added as filters
- Updated the reference search filter form with these new filters


shapiromatron (approved):
Love it! Introducing more functionality, but a net result in fewer lines of code. I think we hit the mark with this filterset abstraction.

casey1173 (approved):

PR #737, opened 2022-11-16
bmds 2022.10 - add execution and structure

shapiromatron:
Add `bmds==22.1`, using BMDS version 2022.10 to HAWC. 

This is the first PR of a few; it implements the backend work required to execute analysis, configure inputs, store outputs, and pass information to the frontend. It only does a first pass of the user-interface; additional information work is required to improve the user-interface, output reporting, and exports to Word and Excel.

Analysis was implemented both for continuous and dichotomous datasets in HAWC.

As part of using the new code, a few notable changes were made:

- Removed the BMDS remote execution configuration required with older versions of BMDS
- Updated github actions on linux to ensure required dependencies to shared object are available; these are not required with windows builds but the bundled linux shared object has them missing.
- Remove bmd2 react components related to editing old analyses
- Simplified data storage to store inputs, outputs,and selected as JSON fields for a modeling session
- Removed the bmd LogicField and customization. Surveyed the existing apps and saw the feature was never used.
- Tests were also implemented for permissions and functionality.

This PR also adds crude, huge single components for the BMD analysis. It is expected as work is continued to revise the and update user-interface, this large component is refactored into smaller components for readability.

**Settings**

**Execution**

**Results**


rabstejnek (approved):
Looks good! Tested it out locally and runs well too. There were some minor frontend concerns but since we're focused on the backend for this PR I think its good to merge!

PR #742, opened 2022-11-29
Summary filtersets

rabstejnek:


hawc/apps/summary/filterset.py
```diff
@@ -0,0 +1,147 @@
...
+        queryset = super().filter_queryset(queryset)
+        query = Q(assessment=self.assessment)
+        if not self.perms["edit"]:
+            query &= Q(published=True)
+        return queryset.filter(query)
+
+    def create_form(self):
+        form = super().create_form()
+
+        if not self.perms["edit"]:
```
shapiromatron, line 133: changed this to hide, instead of disabling. This reduces the UI complexity for browsers, who don't need to see this information and may confuse them.

hawc/apps/summary/filterset.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+        choices = [constants.VisualType(choice) for choice in sorted(set(choices))]
+        return [(choice.value, choice.label) for choice in choices]
+
+    def create_form(self):
+        form = super().create_form()
+        form.fields["type"].choices = self.get_type_choices()
+        return form
+
+    def pop_published(self, form):
```
shapiromatron, line 53: do the pop of published at the end

hawc/apps/summary/filterset.py
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+    def filter_queryset(self, queryset):
+        queryset = super().filter_queryset(queryset)
+        query = Q(assessment=self.assessment)
+        if not self.perms["edit"]:
+            query &= Q(published=True)
+        return queryset.filter(query)
+
+    def get_type_choices(self):
```
shapiromatron, line 39: I moved this into it's own function so it's simpler to override, and makes the `super()` calls simpler

hawc/apps/summary/filterset.py
```diff
@@ -0,0 +1,147 @@
...
+        label="Visualization type",
+        help_text="Type of visualization to display",
+        empty_label="<All>",
+    )
+
+    class Meta(VisualFilterSet.Meta):
+        model = models.DataPivot
+
+    def filter_queryset(self, queryset):
+        return super().filter_queryset(queryset).select_related("datapivotquery", "datapivotupload")
```
shapiromatron, line 74: this fixes an N+1 query issue; required for the get type choices

hawc/apps/summary/filterset.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def get_type_choices(self):
```
shapiromatron, line 76: confusing, but had to set a special case for DataPivotUpload

hawc/apps/summary/filterset.py
```diff
@@ -0,0 +1,147 @@
...
+            query &= Q(published=True)
+        return queryset.filter(query)
+
+    def create_form(self):
+        form = super().create_form()
+
+        if not self.perms["edit"]:
+            # hide published filter and modify layout
+            form.fields.pop("published")
+            form.grid_layout.rows[0].columns[0].width = 8
```
shapiromatron, line 136: general thought - i wonder if we could have a refactoring on the BaseFilterSet to have a `get_grid_layout` method, so it'd be an easy override to either default to what's defined in your meta or do something like if published not in form.fields, then pop off the final width before build the pydantic grid object. It's out of scope for this PR, but I think we have this same pattern in a few other places.

rabstejnek:
Changes look good! Merging

PR #745, opened 2022-12-01
add ids to export

shapiromatron:
Added HERO, PubMed, and DOI values to risk of bias, animal, epi, epi meta, and in vitro exports.


hawc/apps/animal/serializers.py
```diff
@@ -11,13 +11,13 @@
...
 from ..study.models import Study
-from ..study.serializers import StudySerializer
+from ..study.serializers import VerboseStudySerializer
 from ..vocab.constants import VocabularyTermType
 from . import forms, models
 
 
 class ExperimentSerializer(serializers.ModelSerializer):
-    study = StudySerializer(required=False, read_only=True)
+    study = VerboseStudySerializer(required=False, read_only=True)
```
shapiromatron, line 20: my preference would be to not change to this; the verbose serializer pulls in many more models that aren't required and since they're only used for the exporters which get all the data, it makes the query much more expensive.

shapiromatron (commented):
I think it'd be better if we could keep the serializers as is, and then just insert the columns after they export had been created. There's a method here that I think gives all the data we need, on a reference queryset (which could be the ids from the study queryset:

models.Reference.objects.identifiers_dataframe(qs)

shapiromatron (commented):
Awesome job Matt. I had grand ideas to refactor what you did to make the study evaluations and invitro to use the same approach as the other exports, but once I started, I realized it would take a lot more work to complete. I would have done it exactly as you did.

The only change I made was moving the identifier data frame creation from the common app to the Study app, since its more a property of studies.

munnsmunns (approved):
Thanks Andy! The changes look good

PR #747, opened 2022-12-06
Paginated tagging and updated save alerts

rabstejnek:
# Paginated tagging view

Limited the tagging view to only show 100 references at a time

>image

# Updated save alerts

## Standard save alert

The save alert has been updated and moved from beside the "Save and next" button to right below that header. It fades in and then out whenever tags are saved.

Screenshot 2022-12-08 at 10 31 19 PM

## Autoresolution save alert

A custom save alert pops up if tags where automatically resolved when saved. For example:

The only unresolved user tag here is from Project Manager (Exclusion -> Tier I)

>image

If another user goes to the tagging view, applies the same tags as the other unresolved user tags, and clicks "Save and next"...

>image

Then the tags are automatically resolved and added as consensus tags, and the user is notified as such

>image

# Other changes
- Removed js ordering and added to the form
- Fixed tag autoresolution
- Changed tagging view so that only unresolved user tags are used
  - This makes sure that old user tags aren't the default after a resolution; instead it will be the resolved tags

### DIFF
--- a/frontend/lit/TagReferences/Main.js
+++ b/frontend/lit/TagReferences/Main.js
@@ -2,12 +2,12 @@ import {toJS} from "mobx";
 import {inject, observer} from "mobx-react";
 import PropTypes from "prop-types";
 import React, {Component} from "react";
+import Alert from "shared/components/Alert";
 import HelpTextPopup from "shared/components/HelpTextPopup";
 import Modal from "shared/components/Modal";
 import {LocalStorageBoolean} from "shared/utils/LocalStorage";
 
 import Reference from "../components/Reference";
-import ReferenceSortSelector from "../components/ReferenceSortSelector";
 import TagTree from "../components/TagTree";
 
 @inject("store")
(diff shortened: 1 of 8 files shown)


rabstejnek:
@shapiromatron In regards to the alert changes: we originally had it so that the save alert would have to finish fading out before it moved on to the next reference. Since we want the alert to linger a little longer if it informs the user that tags were autoresolved, some of the store and React logic was changed to allow for the store to change its selected reference state without waiting for the alert to go away.

frontend/lit/TagReferences/Main.js
```diff
@@ -117,59 +103,63 @@ class TagReferencesMain extends Component {
...
-                                    the error persists please contact HAWC staff.
-                                </div>
+                                <Alert
+                                    className="alert-danger mt-2"
+                                    message="An error occurred in saving; please wait a moment and retry. If the error persists please contact HAWC staff."
+                                />
                             ) : null}
+                            <div className="well" style={{minHeight: "50px"}}>
+                                {store.successMessage ? (
+                                    <Alert
```
shapiromatron, line 121: moved it into the well to replace the tags so the add/delete wasn't as jarring

frontend/lit/TagReferences/store.js
```diff
@@ -53,51 +53,52 @@ class Store {
...
-                            reference = this.references[nextIndex];
-                        this.setReference(reference);
-                    },
-                });
-            },
-            failure = data => {
-                console.error(data);
-                this.errorOnSave = true;
-            };
-        $.post(url, payload, v => (v.status === "success" ? success() : failure())).fail(failure);
```
shapiromatron, line 90: modernized the jquery post in favor of our fetch api used elsewhere

hawc/static/css/hawc.css
```diff
@@ -638,6 +638,15 @@ table.bordered td {
     transition: opacity 1s ease-out;
 }
 
+/* fade in/out */
+.fade-in-out {
```
shapiromatron, line 642: we can apply this style to approximate the $.fadeIn.fadeOut

shapiromatron (approved):
Pagination is wonderful. I made a few other updates; mainly trying to reduce the DOM layout changes to insert/remove the message box

PR #748, opened 2022-12-07
show comparison set name for group on the result form

shapiromatron:
Users reported difficulty in data entry for `epi.Result` groups for cases where there are multiple groups named the same thing across comparison sets. Therefore, we add the comparison set name in addition to the group name for the data entry label.

| Old | New |
| ------------- | ------------- |
| image  | image  |


shapiromatron:
@munnsmunns please review when you have a minute; a good test might be http://127.0.0.1:8000/epi/result/100501471/update/

munnsmunns (approved):
Looks good! Works as expected on the link you shared

PR #749, opened 2022-12-07
Improvements to the Conflict Resolution View

munnsmunns:
* On conflicts resolution page, filter to only show cases where the count of is_applied=False is > 1
 * Paginate conflict resolution page, 100 at a time
 * Add Actions button to references in the conflict page, with options to:
   * Tag any reference if you want to edit/create your own tags on that reference
   * Edit reference details
   * view reference tag history

image

 * Add an Assesssment.Log for whenever conflicts are resolved, log the ref id, and tag id set which includes all includes and excludes

### DIFF
--- a/hawc/apps/lit/api.py
+++ b/hawc/apps/lit/api.py
@@ -5,7 +5,7 @@ import plotly.express as px
 from django.conf import settings
 from django.core.cache import cache
 from django.db import transaction
-from django.shortcuts import get_object_or_404, render
+from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from rest_framework import exceptions, mixins, status, viewsets
 from rest_framework.decorators import action
--- a/hawc/apps/lit/api.py
+++ b/hawc/apps/lit/api.py
@@ -399,5 +399,10 @@ class ReferenceViewset(
         if not assessment.user_can_edit_object(self.request.user):
             raise PermissionDenied()
         reference.resolve_user_tag_conflicts(selected_user_tag)
-        create_object_log(f"Tag conflict resolved with user_tag {user_tag_id}", reference, assessment.pk, request.user.id)
+        create_object_log(
+            f"Tag conflict resolved with user_tag {user_tag_id}",
+            reference,
+            assessment.pk,
+            request.user.id,
+        )
         return render(request, "lit/_conflict_resolved.html", {"ref": reference})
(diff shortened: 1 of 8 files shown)


hawc/apps/lit/api.py
```diff
@@ -379,26 +379,28 @@ def get_queryset(self):
...
             except ValueError:
                 return Response({"tags": "Array of tags must be valid primary keys"}, status=400)
-            resolved = ref.update_tags(request.user, tags)
+            resolved = instance.update_tags(request.user, tags)
             response["status"] = "success"
             response["resolved"] = resolved
         return Response(response)
 
     @action(detail=True, methods=("post",))
     def resolve_conflict(self, request, pk):
```
shapiromatron, line 395: I modified this; REST APIs should generally return JSON unless otherwise specified by the content type via content negotiation, generally we'd want to build a custom renderer that could return json or html depending on what's requested.

hawc/apps/lit/templates/lit/_reference_tag_conflict.html
```diff
@@ -1,7 +1,15 @@
 <li class="list-group-item list-group-item-action conflict-reference-li {% if first %} pb-4 {% else %} py-4 {% endif %}"
     id="ref-form-{{ref.pk}}"
     hx-target="#ref-form-{{ref.pk}}"
-    hx-swap="outerHTML">
+    hx-swap="delete swap:1s">
```
shapiromatron, line 4: changed the method here to just delete this `li` and give a fade out animation

hawc/apps/lit/templates/lit/_reference_tag_conflict.html
```diff
@@ -13,36 +21,40 @@
...
-                            {% endif %}
-                        {% endfor %}
-                        {% for tag in diff.difference %}
-                            {% include 'lit/_nested_tag.html' with tag=tag extra_classes='refUserTag' %}
-                        {% endfor %}
-                    {% endwith %}
-                </form>
+            {% with ref.tags.all as tags%}
+            {% if tags %}
+            <div class="tag-conflict">
```
shapiromatron, line 26: changed some CSS here to horizontally align the existing consensus tags with the suggested tags; makes it easier to review

hawc/apps/lit/views.py
```diff
@@ -355,13 +356,15 @@ class ConflictResolution(BaseFilterList):
             ]
         },
     )
-    paginate_by = None
+    paginate_by = 100
```
shapiromatron, line 359: awesome!

shapiromatron (approved):
Looks good. Pagination here is really nice; I'm seeing a constant <20 queries to render the response regardless of the number of conflicts.  Made a few UI tweaks, including just removing the `li` if a user resolves a conflict instead of swapping with an alert; makes it cleaner.

PR #752, opened 2022-12-10
Lit review summary page

casey1173:
Updates literature review page to reflect new conflict resolution models & stats.
- Includes a potential visual rework for funsies. Can be easily reverted to existing table view if preferred (while including new stats)
- Clicking a tile will redirect to a filtered list view matching the stat (if possible)
- potential for more graphs

image
image

- Works without conflict resolution too
image


shapiromatron (approved):
awesome job! I made some changes, but nothing huge:

- fix error w/ unauthenticated users
- create utility classes for widths instead of putting into the main class
- reorganized the forms for vertical space
- tried to provide a pydantic default instead of handling multiple types (eg., provide an empty string instead of Optional[str])

PR #754, opened 2022-12-14
Configurable session cookies

shapiromatron:
Make cookie domain/name configurable via environment variables. Use the django defaults by default

- [Session Cookie Domain](https://docs.djangoproject.com/en/4.1/ref/settings/#session-cookie-domain) - defaults to None; a standard domain cookie (django default)
- [Session Cookie Name](https://docs.djangoproject.com/en/4.1/ref/settings/#session-cookie-name) - defaults to `sessionid` (django default)


munnsmunns (approved):
LGTM!

PR #755, opened 2022-12-14
Fix associated identifiers on import

rabstejnek:
Added associated PubMed identifiers on HERO import and improved the handling of associated identifiers on HERO/PubMed imports through bulk operations.

>image

>image

Also updated the `/lit/api/assessment/{assessment_id}/update-reference-metadata-from-hero/` endpoint to set associated identifiers as well, so this can be used to fix existing references in an assessment.

>image

>image

>image


shapiromatron (approved):
What a tough task. The generic methods on the queryset you created seems like a great way to do this, and it made it easy to integrated into this tangle of code.

PR #756, opened 2022-12-14
WIP: Add view for tag state

rabstejnek:


rabstejnek:
@shapiromatron I updated the wording from "tag history" to "tag state", since I believe that is more indicative of what's being shown; I'm open to other suggestions of wording, since it would just be an easy search and replace.

frontend/lit/components/Reference.js
```diff
@@ -114,7 +114,7 @@ class Reference extends Component {
                 <ActionLink key={0} label="Edit reference tags" href={data.editTagUrl} />,
                 <ActionLink key={1} label="Edit reference" href={data.editReferenceUrl} />,
                 <ActionLink key={2} label="Delete reference" href={data.deleteReferenceUrl} />,
-                <ActionLink key={3} label="Tag history" href={data.tagHistoryUrl} />,
+                <ActionLink key={3} label="Tag state" href={data.tagStateUrl} />,
```
shapiromatron, line 117: Technically, I think you're correct that this is the tag state. However, for our users, generally the term state may be confusing as it's a concept they're likely not very familiar with. Consider changing back to "Tag history" or "User tags"

hawc/apps/lit/templates/lit/_reference_with_tags.html
```diff
@@ -0,0 +1,82 @@
+<div id="tagged-ref-{{ref.pk}}">
+    {% if show_actions %}
+    <div class="dropdown btn-group float-right">
+        <a class="btn btn-sm btn-secondary dropdown-toggle" data-toggle="dropdown">Actions</a>
+        <div class="dropdown-menu dropdown-menu-right">
+            <a href="{% url 'lit:tag' assessment.pk%}?id={{ref.pk}}" class="dropdown-item" target="_blank"
```
shapiromatron, line 6: remove the `target="_blank" rel="noreferrer"` from these; move edit reference tags to be after edit reference

shapiromatron (changes requested):
Nice job! Some minor tweaks requested to the language for state vs history; we can discuss in person if you'd like.

shapiromatron (approved):
Looks good! I made two minor, substantive changes: a) renamed "tag state" to "tag status", and b) for the tag status page, we should all user_tags, regardless of if they've been resolved or not. It's just available for the historical record.

PR #761, opened 2023-01-03
add Tableau filters to embedded Visuals

shapiromatron:
Add the ability to filter Tableau data via the API. The screenshots below just copy and pasted the help-text examples. Examples were taken from the latest Tableau embedding documentation: https://help.tableau.com/current/api/embedding_api/en-us/docs/embedding_api_filter.html#filtering-example

image

Only shows two states:
image

---

Also updated `wheel`, which had a [CVE](https://nvd.nist.gov/vuln/detail/CVE-2022-40898), though it's not installed in production so we don't anticipate this to be an issue in production.


rabstejnek (approved):
Looks good!

PR #762, opened 2023-01-04
Add token authentication for django views

munnsmunns:
Token based authentication previously allowed access for REST API views. This has been expanded, so that Token-based authorization is also enabled for standard django views. 

In order to use, you'll need to create a django session and the appropriate cookies, which can be done by submitting the correct token to the `/user/api/validate-token/` endpoint (done by default with the hawc client). After doing so, you can access content using a token.


hawc/main/settings/base.py
```diff
@@ -185,6 +185,10 @@
 # Session and authentication
 AUTH_USER_MODEL = "myuser.HAWCUser"
 AUTH_PROVIDERS = {AuthProvider(p) for p in os.getenv("HAWC_AUTH_PROVIDERS", "django").split("|")}
+AUTHENTICATION_BACKENDS = [
+    "hawc.apps.common.backends.TokenBackend",
```
shapiromatron, line 189: swap order; we will want to use the model first, and fallback to a token

hawc/apps/myuser/api.py
```diff
@@ -25,4 +26,8 @@ class HawcValidateAuthToken(APIView):
     permission_classes = [IsAuthenticated]
 
     def get(self, request, format=None):
+        # also login for use of other views
```
shapiromatron, line 29: why is this needed? suggest removing

tests/client/test_client.py
```diff
@@ -52,6 +52,10 @@ def test_set_authentication_token(self):
         resp = client.set_authentication_token(token)
         assert resp == {"valid": True}
 
+        # can also access regular Django views that are private
```
shapiromatron, line 55: nice! can you also write a check prior to setting the token that an anon going to the same url results in a redirect to login page?

hawc/apps/common/backends.py
```diff
@@ -0,0 +1,40 @@
+from django.contrib.auth.backends import BaseBackend
```
shapiromatron, line 1: rename this file to `auth` instead of `backends`; backends could refer to other things such as db or cache

hawc/apps/common/backends.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class TokenBackend(BaseBackend):
+    keyword = "Token"
+
+    def get_token_from_header(self, request):
+        auth = get_authorization_header(request).split()
+        if not auth or auth[0].lower() != self.keyword.lower().encode():
+            return
+        if len(auth) != 2:
```
shapiromatron, line 15: add this to the prior statement with another `or`

hawc/apps/common/backends.py
```diff
@@ -0,0 +1,40 @@
+from django.contrib.auth.backends import BaseBackend
+from rest_framework.authentication import get_authorization_header
+from rest_framework.authtoken.models import Token
+
+from hawc.apps.myuser.models import HAWCUser
+
+
+class TokenBackend(BaseBackend):
```
shapiromatron, line 8: perhaps better to inherit from the `ModelBackend` and just override the authenticate method? That way we can remove the `get_user` method and the other permissions fields will already be implemented

shapiromatron (changes requested):
Excited to get this in so quickly - a few comments but I think they should be easy to implement

hawc/apps/common/backends.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def authenticate(self, request, token=None, **kwargs):
+        if token is None:
+            token = self.get_token_from_header(request)
+        try:
+            token_obj = Token.objects.select_related("user").get(key=token)
+        except Token.DoesNotExist:
+            return
+        user = token_obj.user
+        if not user.is_active:
```
shapiromatron, line 32: if use inherit from the other class, you can also do a nice one-liner:

`return user if self.user_can_authenticate(user) else None`

hawc/apps/myuser/api.py
munnsmunns: authenticate/login functions aren't called automatically when visiting a page. The easiest path was adding them to this API endpoint. I tried doing this with a middleware but I couldn't get it to work

hawc/apps/common/auth.py
```diff
@@ -0,0 +1,27 @@
+from django.contrib.auth.backends import ModelBackend
+from rest_framework.authentication import get_authorization_header
+from rest_framework.authtoken.models import Token
+
+
+class TokenBackend(ModelBackend):
```
shapiromatron, line 6: if there's no token, this would still query the Token object I think with every request as it's currently written. Suggest changing to bail early if the token is None or empty string.  For example:

```python
class TokenBackend(ModelBackend):
    keyword = "Token"

    def get_token_from_header(self, request) -> str:
        auth = get_authorization_header(request).split()
        token = ""
        if auth and auth[0].lower() == self.keyword.lower().encode() and len(auth) == 2:
            try:
                token = auth[1].decode()
            except UnicodeError:
                pass
        return token

    def authenticate(self, request, token=None, **kwargs):
        if request is not None and token is None:
            token = self.get_token_from_header(request)
        if not token:
            return None
        try:
            token_obj = Token.objects.select_related("user").get(key=token)
        except Token.DoesNotExist:
            return
        return token_obj.user if self.user_can_authenticate(token_obj.user) else None

```

shapiromatron (approved):
Nice job!

hawc/apps/myuser/api.py
```diff
@@ -25,4 +26,6 @@ class HawcValidateAuthToken(APIView):
     permission_classes = [IsAuthenticated]
 
     def get(self, request, format=None):
+        # create a django session for django view
+        login(request, request.user)
```
shapiromatron, line 30: @munnsmunns fyi, I simplified to just log the user in, since they're already confirmed to be authenticated

PR #763, opened 2023-01-09
Assessment admin notes

rabstejnek:
Added an `admin_notes` field to the `Assessment` model. As the name implies this field is meant for admins only, so as such it is not viewable or editable anywhere besides the admin dashboard.

**Admin assessment detail view**

> image


shapiromatron (approved):
Looks good! I tweaked the help text, but checked all APIs and code paths to confirm field was only shown in the admin.

PR #766, opened 2023-01-11
Epi v2 metadata API endpoint

casey1173:
Adds a metadata API endpoint for epiv2, similar to the animal and epiv1 endpoints.

Example endpoints:

- Animal (v1): `/ani/api/metadata/` (not assessment specific)
- Epi (v1): `/epi/api/metadata/:id/` (assessment specific; requires read-access to assessment)
- Epi (v2): `/epidemiology/api/metadata/` (not assessment specific)


shapiromatron (approved):
Looks good! 

I added a few tests, and made the view look a little more like the animal metadata api view instead of the the epi v1. Therefore

- No assessment metadata is shown, its global
- No authentication is required

I made a few other changes along the way, but nothing really substantive. Nice job!

PR #767, opened 2023-01-13
remove copy across study across assessment

shapiromatron:
Remove old `copy_across_assessments` code - it was untested and likely had many errors. If rewritten in the future, this may be useful as a resources. However:

- instead of putting into the models.py, move to a [services](https://github.com/HackSoftware/Django-Styleguide#services) style structure?
- consider `clone(parent=...)` as a different structure?
- build some sort of non-dict data structure to properly store the state changes
- unit tests for each component!


munnsmunns (approved):
Looks good to merge!

PR #770, opened 2023-01-15
fix data pivot initial settings

shapiromatron:
Fixed a bug where a user attempts to start with an initial data pivot template with an invalid settings. 

This PR more defensively gets and updates settings, so that dictionary key errors aren't passed.


casey1173 (approved):
Looks good!

PR #772, opened 2023-01-19
add caption to summary tables

shapiromatron:
A caption can be helpful to explain a complex table, much like the a figure. This PR adds a caption table to `summary.SummaryTable`, in the same style as visuals. 

Captions are added to to tables both in preview mode, in the HTML, and Microsoft Word reports.

In addition, also added the Table title and a hyperlink to the Word Report.


rabstejnek (approved):
Looks good! The "View Online" link in the Word report is a nice touch.

PR #744, opened 2022-11-30
Swapped EPT position of increase/decrease factors

rabstejnek:
Swapped the positions of increased/decreased certainty factors in evidence profile tables.

---

**EPT form**

image

---

**EPT web report**

image

---

**EPT word report**

image

PR #773, opened 2023-01-19
Revert "Swapped EPT position of increase/decrease factors (#744)"

shapiromatron:
Per updated scientist recommendations, revert previous change #744.


rabstejnek (approved):
I tacked on another commit that removes an unused stylesheet; looks good to merge!

PR #774, opened 2023-01-24
Conflict Resolution updates from review

casey1173:
Screening revisions from review:
- Change 'Tag' button
- Add user tag list view
- Bug fixes
- Resolved -> Consensus tags
- Add tag counts to review page tagtree
- Add user tag icon to screening view
- Show full tag length on tag status page
- Fix grammatical error on literature review page
- Button spacing on literature review page


frontend/lit/Reference.js
```diff
@@ -8,7 +8,7 @@ class Reference {
         this.data = data;
         this._quickSearchText = `${data.title}-${data.year}-${data.authors}-${data.authors_short}`.toLowerCase();
         this.tags = data.tags.map(tagId => tagtree.dict[tagId]);
-        this.userTags = data.user_tags ? data.user_tags.map(tagId => tagtree.dict[tagId]) : null;
+        this.userTags = data.user_tags ? data.user_tags.map(tagId => tagtree.dict[tagId]) : [];
```
shapiromatron, line 11: @casey1173 do you remember why you made this change? 

This makes it so that if resolved tags exist, it suggests to the user to remove them?
image

I bet you added to fix something else, just curious if you remember what it was.

casey1173: > @casey1173 do you remember why you made this change?
> 
> This makes it so that if resolved tags exist, it suggests to the user to remove them? image
> 
> I bet you added to fix something else, just curious if you remember what it was.

wow, interesting bug, nice catch! I made that change because I also started calling reference.userTags.length in TagReferences/Main.js to determine whether to add the new user tags icon. Accessing the length property of a null object caused an error

shapiromatron: ok, thanks, that makes sense. I think I've already fix that, so we should be good. The dual icons looks 💯 !

hawc/apps/lit/templates/lit/overview.html
```diff
@@ -130,7 +130,13 @@ <h2 class="d-inline-block">Literature Review</h2>
     </div>
     <div class="col-md-4">
       <div class="lit-review-box box-shadow" style="width: 100%;">
-        <p class="h4 m-0">Tagtree</p>
+        <p class="h4 m-0">Tagtree
```
shapiromatron, line 133: @casey1173 what do you think of using bootstrap badges? 
image

casey1173: Love it!!

swatford: Those look great

shapiromatron (approved):
Great job! I'm really excited to merge these final few in! 

I made a number of minor changes, the only larger one was moving some of the logic to get the parent tags which was sprinkled in the views to a class-method on the `models.Reference` model. I'm not sure that's ultimately where it belongs, but it's at least a little more DRY

PR #776, opened 2023-01-27
hotfix - fix cookie vs token based client sessions

shapiromatron:
Fix a regression from #762. 

We added the ability to login a request via the `/myuser/api/validate-token/` endpoint. This can be useful with browsing the non-API website, but it breaks functionality on the API sdide because CSRF tokens are required (since it finds a valid django session, it uses that first, before falling back to tokens). While inconvenient in the context of this PR, this is desired behavior, (that is, checking session before token).
 
This PR adds a new optional query param to log the user in with a django session if specified, and the corresponding client parameter to specify it. By default, it's disabled and API token based sessions will be used.


munnsmunns (approved):
Looks good!

PR #780, opened 2023-01-30
Update animal dose response plots

shapiromatron:
A number of minor changes to the animal bioassay dose response plots.

Plots are generally shown on endpoint detail page and modals, `/ani/endpoint/:id/`

1. For dichotomous data, the barchart is shown initially instead of the dose-response
2. For dichotomous data, the barchart no longer shows error bars
3. Long titles will new wrap if needed
4. Added buffer around dose-response plot so values at min/max no longer hit axes
5. Fix legend dragging
6. Update plot style for a more modern look
7. Improve dose-response BMD/BMDL rendering

| OLD  | NEW |
| ------------- | ------------- |
| image  | image  |
| image | image |


caseyhans (approved):
LGTM!

PR #781, opened 2023-01-30
Order people

shapiromatron:
A few unrelated user-interface improvements observed when conducting demos for users.

First, we updated the user-select input to sort users in the assessment alphabetically by last name instead of a random order which made it difficult to find people:
image

Second, simplified the user-profile layout:
image

Finally, a few minor type annotation updates for `assessment.models`.


casey1173 (approved):
Love these changes!

PR #783, opened 2023-02-03
Eco updates

casey1173:
Updates to eco:

- Adjustments to various form fields on cause and result
- Added filterset for nestedterm. This requires some python processing, but efficiently returns all ancestors and descendants for a search in a single query, or a few additional

Two additional changes beyond the eco form updates:

- Added a new helper function for `MP_Node` querysets to get all ancestors and descendants given a queryset (see `hawc.apps.common.models.include_related`)
- Refactored `hawc.apps.common.views.BaseFilterList` to separate all the `FilterSet` logic into a self-contained `FilterSetMixin` that isn't tied to an assessment


hawc/apps/eco/views.py
```diff
@@ -76,6 +76,50 @@ def get_success_url(self):
 # Term preview
 class NestedTermList(ListView):
     model = models.NestedTerm
+    paginate_by = 2000
+    filterset_class = filterset.NestedTermFilterset
+
+    def include_ancestors(self, queryset=None):
```
shapiromatron, line 82: This is brilliant. I'm going to refactor a little bit but this approach is perfect; I thought we were going to have hundreds of queries but this puts it all into one; nice job

shapiromatron (approved):
Great job. The only substantive changes I changed were to lift up some of the changes you made into the common modules, since I think they may be useful beyond eco. I added an `include_related` function so other MP_Node can get ancestors or descendants, and refactored the BaseFilterList to move all the non-assessment specific FilterSet into a FilterSetMixin so we didn't have to duplicate the code in the eco views

PR #784, opened 2023-02-09
Eco exports

shapiromatron:
Add two new export formats for ecological term values and data exports:

- export: `/eco/api/assessment/:assessment_id/export/`
- terms: `/eco/api/terms/nested/`

In addition, one item was added to this PR that may be generally useful. I implemented a different strategy for pulling data from the database into a tabular dataset. This approach builds a pandas dataframe for each table (and their related fields), and then a single method combines the data frames into a full export. As an example, the `Study.objects.filter(assessment_id=123).flat_df()` will return a dataframe of study data, but it also includes the matched reference identifiers for a pubmed column and a HERO column, etc.  Doing this work in the database level will likely be very performant compared to the other approach which required django-rest-framework serialization of the nested data structures.


casey1173 (approved):
Looks good so far!
Edit: nevermind I broke a test

casey1173 (approved):

PR #786, opened 2023-02-17
interactive generic summary tables

shapiromatron:
Add optional [datatables.net](https://datatables.net/) interactive to generic tables.

Screenshot 2023-02-16 at 11 02 01 PM

(data generated using [Faker](https://pypi.org/project/Faker/)).


rabstejnek (changes requested):
I noticed some weirdness when sorting by columns on a generic summary table:

image

Also on the heatmap visual, if you fetch one dataset and then fetch a different kind of dataset, you get an error and the page breaks:

image

Besides that it looks good! We should try to address these errors, but after that its good to merge.

rabstejnek:
I believe the sorting problem is because the user copy/pasted HTML into the example I'm looking at. I tried explicitly setting the [column types](https://datatables.net/reference/option/columns.type) to `html` and it was still an issue, so I'm guessing there's nothing we can do about sorting weird HTML fragments that were copied in by a user.

shapiromatron:
> I noticed some weirdness when sorting by columns on a generic summary table:
> 
> image
> 
> Also on the heatmap visual, if you fetch one dataset and then fetch a different kind of dataset, you get an error and the page breaks:
> 
> image
> 
> Besides that it looks good! We should try to address these errors, but after that its good to merge.

Fixed the issue w/ the heatmap visual - ok to merge?

rabstejnek (approved):

PR #787, opened 2023-02-21
django update

shapiromatron:
Update django to latest patch version:
https://docs.djangoproject.com/en/3.2/releases/3.2.18/


munnsmunns (approved):
Looks good! had no issues

PR #789, opened 2023-02-23
User tagging export

shapiromatron:
Add downloadable exports for user tagging, and refine the implement for how we add reference tags.

Available on the `/assessment/:id/downloads/` download page: 

image

---

As part of adding the user-tag exports, we did some substantial API refactoring of literature reference exports. Previously, the same export function was called from three endpoints, depending on if a tag was selected, if a search was selected, or overall for an assessment. This has been modified to have a single route for downloading, and the ability to specify query parameters via django_filters to filter by search or tag. 

Finally, we improved the speed of report generation optimizing the queryset for the models, as well as writing a new set-based method for determining if a parent tag should be marked True in the export as a result of a child tag being used.


casey1173 (approved):
Hefty PR but it all looks good to me!

PR #792, opened 2023-02-27
Assessment values update

munnsmunns:


shapiromatron (changes requested):
some relatively minor changes; I may take this over so you can stay focused on the current work and turn it back to you after changes are done - is that ok?

shapiromatron (approved):
Looks great; I made a bunch of text updates and some form layouts on other assessment things, but happy to get this one merged in.

PR #793, opened 2023-02-27
add plotly template tag

shapiromatron:
Add a plotly template tag for easily adding new figures into our application.  The template tag by default assumes that the plot should be rendered when the DOM is ready, and it should not be resizable, but these are options that can be passed to the template tag.

Refactor plotly figures on the admin to use the new template tag, and make them resizable so that we can shape the figure to fit powerpoint presentations or other reports.

On the view side, you'll just generate a plotly Figure and make sure it's available in the view context. On the template side, you'll pass the figure to our custom tag:

```html
{% load bs4 %}

<-- common usage; wrap a figure inside a bootstrap cell -->
<div class="col-md-4">{% plotly fig %}</div>

<-- render plot from an htmx event; make it resizable by the user -->
{% plotly fig event="htmx" resizable=True %}
```

When resizable; you can change the parent div dimensions (see handle in bottom-right):
image


casey1173 (approved):
Looks good!

PR #794, opened 2023-02-28
get filterset class dynamically

shapiromatron:
Searching and filtering options on the reference tagging views are different when conflict resolution is enabled versus disabled. This work modifies the class-based view workflow to dynamically build the appropriate FilterSet class, based on the view request context. 

Like form views, we now have a `get_filterset_class` method that defaults to  `filterset_class`, but can be overridden to dynamically get the correct class.

This was then used on the `TagReferences` class to build the correct form based on assessment configuration.

With conflict resolution:
image

Without conflict resolution:
image


casey1173 (approved):
@shapiromatron I fixed the filterset fields and the grid layout, changed the 'Tag' button links to match the expected workflow, and made a minor style change to the tagtree. Back to you for final check + merge

PR #799, opened 2023-03-02
standardize copy views

shapiromatron:
Our copy a selected object as a new object was done ad hoc with lots of different patterns throughout HAWC. This PR standardizes the views and uses two base clasess:

- For the view: `hawc.common.views.BaseCopyForm`
- For the form: `hawc.common.forms.CopyForm`


caseyhans (approved):
LGTM

PR #800, opened 2023-03-02
browsable API (in debug) for pandas renderers

shapiromatron:
The browsable API is useful in debug mode because you can use to get information on database query usage. This PR adds this API view to our standard PandasRenderers in debug mode. Previously, we had no easy mechanism to profile database performance with these.


munnsmunns (approved):
LGTM!

PR #804, opened 2023-04-06
Add Custom Plotly visual type

munnsmunns:
Add a new visual type that allows the user to create a custom plotly visual from JSON. This visual type is still a work in progress and is hidden behind the `ENABLE_PLOTLY_VISUAL` feature flag.


shapiromatron (changes requested):
This is exciting! Can you put the visual selector and the form view behind a new feature flag? maybe we can call it `JSON_VISUALS`, and default it to False.

See here: https://github.com/shapiromatron/hawc/pull/802

- if the flag is True, show the new option on the visual selector page
- if the page is False and someone tries to create a version of this, raise a 404 or 403

shapiromatron (approved):
Looks good! We need to think through a good way to remove this feature flag for general release, after people try it out for a little bit... @swatford

PR #805, opened 2023-04-06
updated python, yarn and postgresql versions in docs

dannypeterson:
Updated development documentation.
- kept python version 3.9
- yarn version --> 1
- postgresql >=12

- tested with python 3.9, latest version of postgresql (v15.2), and yarn v1.22.19

### DIFF
--- a/docs/docs/development.md
+++ b/docs/docs/development.md
@@ -5,7 +5,7 @@ Assessment Workspace Collaborative project.  To begin you should have the
 following applications installed on your local development system:
 
 - [Git](https://git-scm.com/)
-- [Python](https://www.python.org/) == 3.10
+- [Python](https://www.python.org/) == 3.9
 - [Node.js](https://nodejs.org)
 - [Yarn](https://yarnpkg.com/) < 2
 - [PostgreSQL](https://www.postgresql.org/) >= 12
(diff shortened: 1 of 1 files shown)


docs/docs/development.md
```diff
@@ -5,10 +5,10 @@ Assessment Workspace Collaborative project.  To begin you should have the
 following applications installed on your local development system:
 
 - [Git](https://git-scm.com/)
-- [Python](https://www.python.org/) == 3.9
+- [Python](https://www.python.org/) == 3.10
```
shapiromatron, line 8: the current python version we have deployed is 3.9; can we keep this one back on 3.9 for now? there's a PR we'll merge in next quarter to move to 3.11, see https://github.com/shapiromatron/hawc/pull/769

shapiromatron (commented):
Can we keep python back on version 3.9? i know it works with our build tools, CI/CD etc, I wonder if this is just a weird conda issue?

docs/docs/development.md
dannypeterson: just tested with python 3.9 and it still works! going to send in a new pull request with the change

shapiromatron (approved):

PR #807, opened 2023-04-12
Downloads page revamp

casey1173:
Prettify the assessment downloads page.
- Use card columns for more horizontal real estate
- Let users choose file type
- Adds simple help text for all downloads, based off endpoint help text
- Update/add correlating endpoint help text
- Move a previous css style fix so it applies to PRIME flavor

Before:
image

After:
image

Downloads without data are now shown as disabled instead of completely hidden.
image


hawc/apps/assessment/templates/assessment/assessment_downloads.html
```diff
@@ -6,91 +6,156 @@ <h2>{{object}} downloads</h2>
   <p class='form-text text-muted'>
     All data from HAWC are exportable into Excel. Developer exports in JSON
     format are also available (please <a target="_blank" href="{% url 'contact' %}">contact us</a> for more information).
+    Visualizations can be exported into SVG and PNG formats;
+    you can download each individual visualization when being viewed.
   </p>
-
-  <ul>
+  <div class="card-columns">
     {% if assessment.enable_literature_review and assessment.has_lit_data %}
```
shapiromatron, line 13: @casey1173 one thought that came out of our meeting was to show the cards even if there's no data, so users know there's no data for this domain. So I think this would be moving the if login inside the card-body, to basically show either the body with buttons or a "no data available" body

shapiromatron (approved):
Looks great. I made the "no data available" more muted and hide the button (but not the data type). I like the changes a lot, plus we have additional functionality for other export types.

PR #810, opened 2023-04-18
Download visuals using client

munnsmunns:
Adds a method to the client Summary class that uses Playwright to download all images from a given assessment id. The images are saved as individual pngs.

Methods to download a single visual or data pivot are accessed within the new `InteractiveHawcClient` context manager class.

Example use:

```python
from hawc_client import HawcClient

client = HawcClient("https://hawc.com")
client.set_authentication_token(token, login=True)

with client.interactive(headless=False) as iclient:
    # return visual PNG as bytes; also save to the `visual.png` file
    png_bytes = iclient.download_visual(visual_id, 'visual.png')

    # just return the bytes for a data pivot
    png_bytes = iclient.download_data_pivot(data_pivot_id)
```

This adds `playwright` to the dependencies, and now requires a minimum Python version of 3.10.


client/hawc_client/summary.py
```diff
@@ -49,3 +54,115 @@ def table_list(self, assessment_id: int) -> pd.DataFrame:
...
+    def download_visuals(
+        self, assessment_id: int, dest: str = ".", file_type: str = "docx", headless: bool = True
+    ):
+        """
+        Retrieves all visuals for the given assessment and saves them in the given path.
+        Session must be authenticated with a token to access private assessment visuals.
+
+        Args:
+            assessment_id (int): Assessment ID
+            dest (str): Folder where images will be saved. Defaults to current working directory.
```
shapiromatron, line 67: If we could make this either a string or a `bytes.IO` instance, then we could write to disk if string, and use memory if `bytes.IO`. That would fix the temp file directory issue too I think

shapiromatron:
I started on a revision for this, but never finished it. I think it'd be goo to reuse the playwright instance with multiple functional calls, so it would be possible to either use a batch call or a series of individual calls interchangeably. This might be best accomplished by a [class-based context manager](https://realpython.com/python-with-statement/) that spins up a playwright instance for this case.  I was thinking it'd look something like this:

```
# pass in the non-interactive client to take login credentials
with InteractiveHawcClient(client) as iclient:
    iclient.download_visual(123)
    iclient.download_data_pivot(123)
```

munnsmunns:
I tested it in headed mode and it looks like the login/token is not set properly on the playwright page, so its only getting the 403 page

shapiromatron (approved):
I spent some time refactoring things and wrote a pretty extensive test script that attempts to download images for all visualization types we currently support. I ended up figuring out a way to determine which approach to use to download a figure without having to know from the client perspective; it just looks for common components currently on the website to determine which approach to use.

Ready for your testing and review @munnsmunns - if you're ok with how things look, ok to merge!

PR #811, opened 2023-04-20
fix reference -> study form

shapiromatron:
Fixes a regression where when literature are tagged with the inclusion tag with a bulk convert to study form. To test, you'll need an assessment that has been tagged with inclusion tag, and then you can convert to study via the form: `/lit/assessment/:id/references/extraction-ready/`

This also writes some unit tests to check views work as expected.


munnsmunns (approved):
Looks good!

PR #813, opened 2023-04-22
return a 403 instead of stopping views at API level

shapiromatron:
Previously for visuals and data pivots, if a visual was not published, then the page would load, but then the user would get an error when the API fetched the data, since the data wouldn't be returned since it was unpublished. This is confusing and made it hard for users to determine if there was a HAWC error or if they shouldn't have access. This PR makes it so that if a visual is not published and you don't have access, the appropriate login redirect/403 errors are returned.


shapiromatron:
@rabstejnek per your feedback, renamed `object` to `obj`. This is consistent with naming in django CBV: https://ccbv.co.uk/projects/Django/4.2/django.views.generic.detail/DetailView/#get_object

PR #814, opened 2023-04-22
rename all Read views to Detail views for consistency

shapiromatron:
Consistently name Read views Detail views. This is more standard with CRUD (create/read/update/delete) and hopefully removes an older pattern to make hawc more consistent.


dannypeterson (approved):
went through each app and found no other instances of read views

PR #815, opened 2023-04-22
rename viewset -> ViewSet

shapiromatron:
We inconsistently named classes as Viewset and ViewSet, case insensitive. This PR makes the naming consistent through hawc with the DRF preferred name, ViewSet:

https://www.django-rest-framework.org/api-guide/viewsets/


dannypeterson (approved):
could not find any more instances of Viewset good job

shapiromatron:
> could not find any more instances of Viewset good job

Thanks! You can do a case-sensitive search in vscode that makes it pretty simple:

image

Ended up finding one more in the screenshot :)

PR #816, opened 2023-04-23
Tabular export - metadata worksheet

shapiromatron:
Add new optional worksheet which can be added to excel exports - a metadata tab. This can be used to describe the data in the data tab. To add a metadata tab, you'll need to add a new method `build_metadata` to an exporter, which returns a pandas data frame to be used for the metadata.

Additionally, automatically format excel exports with a few updates:

- freeze header row
- enable filters
- set header to a nice dark blue+white text; update column width 

**Data tab** - color, filter, and freezing on all data exports:
image

**Metadata tab** - currently implemented for risofbias exports only:
image


casey1173 (approved):
Love this, looks good to me! Snuck in an unrelated fix for some weird text alignment on the downloads page.

PR #819, opened 2023-05-02
Added alert in manual study entry page

dannypeterson:
Prevent users from attempting to manually enter study data:
- shown for: /study/assessment/:id/new-study/ (manually entered)

Even though the same template is used, we hide for these views:
- hidden for: /study/assessment/:id/create-from-identifier/
- hidden for: /study/:id/add-details/

image


shapiromatron (approved):
Nice job! Just tweaked the language/layout a little.

PR #820, opened 2023-05-02
implement epiv2 api

DookTibs:
* CRUD api endpoints for design, exposure measurements, chemicals, exposure_levels, outcomes, adjustment factors, data extraction

* test cases for all of the above

* a new ArrayedFlexibleChoiceField (plus tests), for FlexibleChoiceField-style behavior when the underlying field is wrapped in an ArrayField.

* new BelongsToSameDesign mixin to validate that related objects belong to same overall design, e.g. when associating a chemical with an exposure level.

* add environment variable (SKIP_BMDS_TESTS=1) to disable bmds testing.

### DIFF
--- a/hawc/apps/epiv2/api.py
+++ b/hawc/apps/epiv2/api.py
@@ -47,3 +47,45 @@ class Design(EditPermissionsCheckMixin, AssessmentEditViewset):
 class Metadata(viewsets.ViewSet):
     def list(self, request):
         return EpiV2Metadata.handle_request(request)
+
+
+class Chemical(EditPermissionsCheckMixin, AssessmentEditViewset):
+    edit_check_keys = ["design"]
+    assessment_filter_args = "design__study__assessment"
+    model = models.Chemical
+    serializer_class = serializers.ChemicalSerializer
+
+
+class Exposure(EditPermissionsCheckMixin, AssessmentEditViewset):
+    edit_check_keys = ["design"]
+    assessment_filter_args = "design__study__assessment"
+    model = models.Exposure
+    serializer_class = serializers.ExposureSerializer
+
+
+class ExposureLevel(EditPermissionsCheckMixin, AssessmentEditViewset):
+    edit_check_keys = ["design"]
+    assessment_filter_args = "design__study__assessment"
+    model = models.ExposureLevel
+    serializer_class = serializers.ExposureLevelSerializer
+
+
+class Outcome(EditPermissionsCheckMixin, AssessmentEditViewset):
+    edit_check_keys = ["design"]
+    assessment_filter_args = "design__study__assessment"
+    model = models.Outcome
+    serializer_class = serializers.OutcomeSerializer
+
+
+class AdjustmentFactor(EditPermissionsCheckMixin, AssessmentEditViewset):
+    edit_check_keys = ["design"]
+    assessment_filter_args = "design__study__assessment"
+    model = models.AdjustmentFactor
+    serializer_class = serializers.AdjustmentFactorSerializer
+
+
+class DataExtraction(EditPermissionsCheckMixin, AssessmentEditViewset):
+    edit_check_keys = ["design"]
+    assessment_filter_args = "design__study__assessment"
+    model = models.DataExtraction
+    serializer_class = serializers.DataExtractionSerializer
--- a/hawc/apps/epiv2/mixins.py
+++ b/hawc/apps/epiv2/mixins.py
@@ -1,0 +1,69 @@
+from rest_framework import exceptions, serializers
+
+from .models import Design
+
+
+class BelongsToSameDesignMixin:
+    """
+    ex: when updating an ExposureLevel, there's a related Chemical and Exposure sometimes passed into the payload as
+    "chemical_id" and "exposure_level_id". Basic DRF functionality verifies that the supplied values are valid
+    items in the database, but it's a common requirement that they also belong to the same parent object; in this
+    case, the same design. This mixin performs that type of validation logic.
+
+    Could potentially generalize this even further (so that it's not just useful for "Design"); this type of
+    relationship is not uncommon.
+    """
+
+    def validate(self, data):
+        fields_to_check = []
+        try:
+            fields_to_check = self.belongs_to_same_design_fields
+
+            # first off - were any of the parameters we're concerned with in the payload? If not, there's
+            # no need to perform any custom validation.
+            perform_custom_validation = False
+            for field_to_check in fields_to_check:
+                param_name = field_to_check[0]
+                if param_name in self.initial_data:
+                    perform_custom_validation = True
+                    break
+
+            if perform_custom_validation is True:
+                # figure out what design we're concerned with...
+                if "design" in self.initial_data:
+                    # load from the supplied payload...
+                    design_id = self.initial_data.get("design")
+                    design = Design.objects.get(id=design_id)
+                else:
+                    # load from the previously saved state; the APIException here should never happen but just in case...
+                    if self.instance is not None:
+                        design = self.instance.design
+                    else:
+                        raise exceptions.APIException(
+                            "validation failed; could not determine relevant design id"
+                        )
+
+            invalid_fields = []
+            for field_to_check in fields_to_check:
+                param_name = field_to_check[0]
+                model_klass = field_to_check[1]
+
+                if param_name in self.initial_data:
+                    candidate_id = self.initial_data.get(param_name)
+                    candidate_obj = model_klass.objects.get(id=candidate_id)
+
+                    if candidate_obj.design.id != design.id:
+                        invalid_fields.append(
+                            (
+                                param_name,
+                                f"object with id={candidate_id} does not belong to the correct design.",
+                            )
+                        )
+            if len(invalid_fields) > 0:
+                raise serializers.ValidationError({x[0]: x[1] for x in invalid_fields})
+        except AttributeError:
+            raise Exception(
+                f"{type(self).__name__} is configured to mixin BelongsToSameDesignMixin but lacks a properly configured property 'belongs_to_same_design_fields'"
+            )
+
+        return super().validate(data)
--- a/hawc/apps/epiv2/serializers.py
+++ b/hawc/apps/epiv2/serializers.py
@@ -1,77 +1,168 @@
 from django.db import transaction
 from rest_framework import serializers
 
-from ..common.serializers import FlexibleChoiceField, IdLookupMixin
+from ..assessment.models import DSSTox
+from ..assessment.serializers import DSSToxSerializer
+from ..common.serializers import ArrayedFlexibleChoiceField, FlexibleChoiceField, IdLookupMixin
 from ..epi.serializers import StudyPopulationCountrySerializer
+from ..study.models import Study
 from ..study.serializers import StudySerializer
-from . import constants, models
+from . import constants, mixins, models
 
 
 class ChemicalSerializer(serializers.ModelSerializer):
+    dsstox_id = serializers.PrimaryKeyRelatedField(
+        write_only=True,
+        source="dsstox",
+        queryset=DSSTox.objects.all(),
+        required=False,
+        allow_null=True,
+    )
+    dsstox = DSSToxSerializer(read_only=True)
+
     class Meta:
         model = models.Chemical
-        fields = ["id", "name", "dsstox"]
+        fields = ["id", "design", "name", "dsstox_id", "dsstox"]
 
 
 class ExposureSerializer(serializers.ModelSerializer):
     exposure_route = FlexibleChoiceField(choices=constants.ExposureRoute.choices)
+    biomonitoring_matrix = FlexibleChoiceField(
+        choices=constants.BiomonitoringMatrix.choices, allow_blank=True
+    )
+    biomonitoring_source = FlexibleChoiceField(
+        choices=constants.BiomonitoringSource.choices, allow_blank=True
+    )
 
     class Meta:
         model = models.Exposure
-        exclude = ["design", "created", "last_updated"]
+        exclude = ["created", "last_updated"]
 
 
-class ExposureLevelSerializer(serializers.ModelSerializer):
+class ExposureLevelSerializer(mixins.BelongsToSameDesignMixin, serializers.ModelSerializer):
+    variance_type = FlexibleChoiceField(choices=constants.VarianceType.choices)
+    ci_type = FlexibleChoiceField(choices=constants.ConfidenceIntervalType.choices)
+
+    belongs_to_same_design_fields = [
+        ("chemical_id", models.Chemical),
+        ("exposure_measurement_id", models.Exposure),
+    ]
+
+    chemical_id = serializers.PrimaryKeyRelatedField(
+        write_only=True,
+        source="chemical",
+        queryset=models.Chemical.objects.all(),
+        required=True,
+        allow_null=False,
+    )
+    chemical = ChemicalSerializer(read_only=True)
+    exposure_measurement_id = serializers.PrimaryKeyRelatedField(
+        write_only=True,
+        source="exposure_measurement",
+        queryset=models.Exposure.objects.all(),
+        required=True,
+        allow_null=False,
+    )
+    exposure_measurement = ExposureSerializer(read_only=True)
+
     class Meta:
         model = models.ExposureLevel
-        exclude = ["design", "created", "last_updated"]
+        exclude = ["created", "last_updated"]
 
 
 class OutcomeSerializer(serializers.ModelSerializer):
+    system = FlexibleChoiceField(choices=constants.HealthOutcomeSystem.choices)
+
     class Meta:
         model = models.Outcome
-        exclude = ["design", "created", "last_updated"]
+        exclude = ["created", "last_updated"]
 
 
 class AdjustmentFactorSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.AdjustmentFactor
-        exclude = ["design", "created", "last_updated"]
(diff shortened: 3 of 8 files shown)


hawc/apps/epiv2/api.py
```diff
@@ -47,3 +47,45 @@ def get_queryset(self, *args, **kwargs):
 class Metadata(viewsets.ViewSet):
     def list(self, request):
         return EpiV2Metadata.handle_request(request)
+
+
+class Chemical(EditPermissionsCheckMixin, AssessmentEditViewset):
```
shapiromatron, line 52: rename this and others to `ChemicalViewSet`, or whatever the model is

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    ex: when updating an ExposureLevel, there's a related Chemical and Exposure sometimes passed into the payload as
+    "chemical_id" and "exposure_level_id". Basic DRF functionality verifies that the supplied values are valid
+    items in the database, but it's a common requirement that they also belong to the same parent object; in this
+    case, the same design. This mixin performs that type of validation logic.
+
+    Could potentially generalize this even further (so that it's not just useful for "Design"); this type of
+    relationship is not uncommon.
+    """
+
```
shapiromatron, line 16: add a type annotation here: `belongs_to_same_design_fields: list[tuple[str, type[models.Model]]`

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,69 @@
...
+                        invalid_fields.append(
+                            (
+                                param_name,
+                                f"object with id={candidate_id} does not belong to the correct design.",
+                            )
+                        )
+            if len(invalid_fields) > 0:
+                raise serializers.ValidationError({x[0]: x[1] for x in invalid_fields})
+        except AttributeError:
+            raise Exception(
```
shapiromatron, line 65: remove this whole try/except? or only worry about the case where `belongs_to_same_design_fields` isn't defined, so that the whole block isn't in a big try; if it's configured incorrectly it's ok to throw a standard excpetion

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,69 @@
+from rest_framework import exceptions, serializers
+
+from .models import Design
+
+
+class BelongsToSameDesignMixin:
```
shapiromatron, line 6: rename to `SameDesignSerializerMixin`?

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+
+    def validate(self, data):
+        fields_to_check = []
+        try:
+            fields_to_check = self.belongs_to_same_design_fields
+
+            # first off - were any of the parameters we're concerned with in the payload? If not, there's
+            # no need to perform any custom validation.
+            perform_custom_validation = False
```
shapiromatron, line 24: can you do something like

```python
perform_custom_validation = any(field_name in self.initial_data for field_name, _ in fields_to check)
```

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,69 @@
...
+                            )
+                        )
+            if len(invalid_fields) > 0:
+                raise serializers.ValidationError({x[0]: x[1] for x in invalid_fields})
+        except AttributeError:
+            raise Exception(
+                f"{type(self).__name__} is configured to mixin BelongsToSameDesignMixin but lacks a properly configured property 'belongs_to_same_design_fields'"
+            )
+
+        return super().validate(data)
```
shapiromatron, line 69: can we call the super first? and then if perform_custom_validation is false, just return early?

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,69 @@
...
+                    candidate_obj = model_klass.objects.get(id=candidate_id)
+
+                    if candidate_obj.design.id != design.id:
+                        invalid_fields.append(
+                            (
+                                param_name,
+                                f"object with id={candidate_id} does not belong to the correct design.",
+                            )
+                        )
+            if len(invalid_fields) > 0:
```
shapiromatron, line 62: you can just say `if invalid_fields:`

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,69 @@
...
+                    # load from the previously saved state; the APIException here should never happen but just in case...
+                    if self.instance is not None:
+                        design = self.instance.design
+                    else:
+                        raise exceptions.APIException(
+                            "validation failed; could not determine relevant design id"
+                        )
+
+            invalid_fields = []
+            for field_to_check in fields_to_check:
```
shapiromatron, line 47: use tuple unpacking

```python
for param_name, cls in fields_to_check
```

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,69 @@
...
+                if "design" in self.initial_data:
+                    # load from the supplied payload...
+                    design_id = self.initial_data.get("design")
+                    design = Design.objects.get(id=design_id)
+                else:
+                    # load from the previously saved state; the APIException here should never happen but just in case...
+                    if self.instance is not None:
+                        design = self.instance.design
+                    else:
+                        raise exceptions.APIException(
```
shapiromatron, line 42: maybe rewrite?; I'd probably put all this logic in another method for named `_get_design` or something

```python

design = getattr(self.instance, 'design', None) if self.instance else None
design_id = self.initial_data.get("design")
if design_id:
    try:
        design = Design.objects.get(id=design_id)
    except not found  # catch if id doesnt exist
        design = None
if design is None:
    raise ValidationError()
return design
```

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,69 @@
...
+                        )
+
+            invalid_fields = []
+            for field_to_check in fields_to_check:
+                param_name = field_to_check[0]
+                model_klass = field_to_check[1]
+
+                if param_name in self.initial_data:
+                    candidate_id = self.initial_data.get(param_name)
+                    candidate_obj = model_klass.objects.get(id=candidate_id)
```
shapiromatron, line 53: catch a object not found and raise validation error

hawc/apps/epiv2/serializers.py
```diff
@@ -1,77 +1,168 @@
...
-        exclude = ["design", "created", "last_updated"]
-
+        exclude = ["created", "last_updated"]
+
+
+class ExposureLevelSerializer(mixins.BelongsToSameDesignMixin, serializers.ModelSerializer):
+    variance_type = FlexibleChoiceField(choices=constants.VarianceType.choices)
+    ci_type = FlexibleChoiceField(choices=constants.ConfidenceIntervalType.choices)
+
+    belongs_to_same_design_fields = [
```
shapiromatron, line 46: move the `belongs_to_same_design_fields` down below all the other serializer fields, and should we rename this to `same_design_fields`. Do this for other serializer modes as well

hawc/apps/epiv2/urls.py
```diff
@@ -7,6 +7,12 @@
 router.register(r"assessment", api.EpiAssessmentViewset, basename="assessment")
 router.register(r"design", api.Design, basename="design")
 router.register(r"metadata", api.Metadata, basename="metadata")
+router.register(r"chemical", api.Chemical, basename="chemical")
+router.register(r"exposure", api.Exposure, basename="exposure")
+router.register(r"exposurelevel", api.ExposureLevel, basename="exposurelevel")
```
shapiromatron, line 12: use hyphens `exposure-level` for this and other multiwords?  i don't think they need to be raw strings

tests/hawc/apps/bmd/test_models.py
```diff
@@ -1,11 +1,16 @@
...
 
 from hawc.apps.animal.models import Endpoint
 from hawc.apps.bmd.constants import SelectedModel
 from hawc.apps.bmd.models import Session
 
+SKIP_BMDS_TESTS = os.environ.get("SKIP_BMDS_TESTS", "0") == "1"
+
 
 @pytest.mark.django_db
+@pytest.mark.skipif(SKIP_BMDS_TESTS is True, reason="skippyjon")
```
shapiromatron, line 13: change reason :)

tests/hawc/apps/bmd/test_models.py
```diff
@@ -1,11 +1,16 @@
+import os
+
 import pytest
 
 from hawc.apps.animal.models import Endpoint
 from hawc.apps.bmd.constants import SelectedModel
 from hawc.apps.bmd.models import Session
 
+SKIP_BMDS_TESTS = os.environ.get("SKIP_BMDS_TESTS", "0") == "1"
```
shapiromatron, line 9: its not great, but the standard pattern we use for this is

```python
SKIP_BMDS_TESTS = bool(os.environ.get("SKIP_BMDS_TESTS", "False") == "True")
```

hawc/apps/epiv2/mixins.py
DookTibs: added type annotation in the mixin, e.g. 

```
class SameDesignSerializerMixin:
   def validate(self, data):
      fields_to_check: list[tuple[str, type[models.Model]]] = self.belongs_to_same_design_fields
```

hawc/apps/epiv2/mixins.py
DookTibs: the Java programmer in me forgets the pythonic way to do things sometimes :) -- done, took it out of the try/catch block

hawc/apps/epiv2/mixins.py
DookTibs: done

hawc/apps/epiv2/mixins.py
DookTibs: that's much nicer yea, done

hawc/apps/epiv2/mixins.py
DookTibs: done, broke it out into funtion _get_contextual_design in the mixin, and reworked it a bit

hawc/apps/epiv2/mixins.py
DookTibs: neat, I either didn't know you could do that with a list, or I forgot.

hawc/apps/epiv2/mixins.py
shapiromatron: I was thinking more like this?

```python
class BelongsToSameDesignMixin:
    """
    ex: when updating an ExposureLevel, there's a related Chemical and Exposure sometimes passed into the payload as
    "chemical_id" and "exposure_level_id". Basic DRF functionality verifies that the supplied values are valid
    items in the database, but it's a common requirement that they also belong to the same parent object; in this
    case, the same design. This mixin performs that type of validation logic.

    Could potentially generalize this even further (so that it's not just useful for "Design"); this type of
    relationship is not uncommon.
    """ 

    fields_to_check: list[tuple[str, type[models.Model]]]

    def validate(self, data):
        ...

```

hawc/apps/epiv2/api.py
shapiromatron: let's leave it be for v1 for now, we'll just fix it moving forward

hawc/apps/epiv2/urls.py
DookTibs: I went back and forth on this when setting them up - if I recall, my rationale was that in the urlpatterns defined later in that file, they weren't hyphen delimited, and I wanted to keep the API ones consistent with that? Anyway - for the new API routes, switched "exposurelevel" -> "exposure-level", etc.

```
    # exposure level
    path(
        "exposurelevel/<int:pk>/create/",
        views.ExposureLevelViewset.as_view(),
        {"action": "create"},
        name="exposurelevel-create",
    ),
# .....
    # adjustment factor
    path(
        "adjustmentfactor/<int:pk>/create/",
        views.AdjustmentFactorViewset.as_view(),
        {"action": "create"},
        name="adjustmentfactor-create",
    ),
 # etc.
```

tests/hawc/apps/bmd/test_models.py
DookTibs: whoops placeholder slipped by :)

DookTibs:
@shapiromatron went through feedback, just had outstanding questions on a few items so I left those unresolved. Let me know what you think

hawc/apps/epiv2/mixins.py
shapiromatron: ok great! just make sure that doesn't throw a 500 error; I guess it checks when it's building the serializer up prior the custom validation here

shapiromatron:
> @shapiromatron went through feedback, just had outstanding questions on a few items so I left those unresolved. Let me know what you think

Thanks - could you push up the latest version of code w/ updates? I only see a single commit

DookTibs:
trying again; I am not getting any complaints when I run black locally, not sure if I have an out of date config?

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,75 @@
...
+    ex: when updating an ExposureLevel, there's a related Chemical and Exposure sometimes passed into the payload as
+    "chemical_id" and "exposure_level_id". Basic DRF functionality verifies that the supplied values are valid
+    items in the database, but it's a common requirement that they also belong to the same parent object; in this
+    case, the same design. This mixin performs that type of validation logic.
+
+    Could potentially generalize this even further (so that it's not just useful for "Design"); this type of
+    relationship is not uncommon.
+    """
+
+    _fields_to_check: list[tuple[str, type[models.Model]]]
```
shapiromatron, line 19: I think this is a public attribute since you need to specify it on the class that uses it; remove underscore?

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,75 @@
...
+        # passed in initial_data. But just in case:
+        if design is None:
+            raise serializers.ValidationError({"design": "could not determine contextual design"})
+
+        return design
+
+    def validate(self, data):
+        validated_data = super().validate(data)
+
+        self._fields_to_check = self.same_design_fields
```
shapiromatron, line 46: there's no reason to set this - remove this line?

hawc/apps/epiv2/mixins.py
```diff
@@ -0,0 +1,75 @@
...
+        perform_custom_validation = any(
+            [field_name in self.initial_data for field_name, _ in self._fields_to_check]
+        )
+
+        if perform_custom_validation is False:
+            return validated_data
+        else:
+            design = self._get_contextual_design()
+
+            invalid_fields = []
```
shapiromatron, line 59: use a dictionary as a data structure instead? then you dont have to convert back to a dictionary in like 73?

shapiromatron (changes requested):
A few more very minor changes/renames. Happy to chat if it's helpful to discuss.

hawc/apps/epiv2/api.py
```diff
@@ -34,7 +34,7 @@ def export(self, request, pk):
         return Response(exporter.build_export())
 
 
-class Design(EditPermissionsCheckMixin, AssessmentEditViewset):
+class DesignViewset(EditPermissionsCheckMixin, AssessmentEditViewset):
```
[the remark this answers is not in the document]
DookTibs: I modified this on all the ones I touched (DesignViewset -> DesignViewSet) but not on existing untouched ones (AssessmentEditViewset, etc.) since the other PR seems to cover those.

hawc/apps/epiv2/mixins.py
DookTibs: reworked. re: formatting, I had been using flake8, didn't realize you'd switched to ruff. When I add that to my setup it catches everything (I think!)

tests/hawc/apps/common/test_serializer.py
```diff
@@ -171,3 +172,49 @@ def test_flexible_db_linked_choice_field(db_keys):
...
+        converted = numeric_choice.to_internal_value(valid_input)
+        assert len(converted) == 2 and converted[0] == 1 and converted[1] == 0
+
+    # Works for text or numeric keys
+    for valid_input in (["job", "occupation", "OcCuPaTiOn"],):
+        converted = text_choice.to_internal_value(valid_input)
+        assert len(set(converted)) == 1 and converted[0] == "job"
+
+    for invalid_input in (["example", "bad input"], [1, "another bad input"]):
+        try:
```
shapiromatron, line 202: fyi, I changed this to use pytest.raises

```python
for invalid_input in (["example", "bad input"], [1, "another bad input"]):
    with pytest.raises(ValidationError):
        numeric_choice.to_internal_value(invalid_input)
```

tests/hawc/apps/common/test_serializer.py
```diff
@@ -171,3 +172,49 @@ def test_flexible_db_linked_choice_field(db_keys):
     assert len(serialized_metrics) == len(result_metrics_list)
     assert serialized_metrics[0]["id"] == first_metric.id
     assert serialized_metrics[1]["id"] == second_metric.id
+
+
+@pytest.mark.django_db
+def test_arrayed_flexible_choice_field(db_keys):
```
shapiromatron, line 178: converted to a class-based test to check each method separately

tests/hawc/apps/epiv2/test_api.py
shapiromatron, line 1: amazing test suite; thank you for writing all of these

shapiromatron (approved):
Fantastic work. I realized we forgot to update the hawc client to create these items. Perhaps it's something to chat about at our next meeting, since it sounds like ya'll will be using these APIs pretty soon...

PR #821, opened 2023-05-03
added alert to study eval page if metrics do not exist

dannypeterson:
### Added info alert using if statement to check if metrics exist for assessment
study eval


hawc/apps/riskofbias/views.py
```diff
@@ -73,7 +73,7 @@ def grouped(metrics: Iterable) -> list[list[models.RiskOfBiasMetric]]:
         )
         context.update(
             metrics=metrics,
-            no_data=metrics == len(metrics) == 0,
+            no_data=len(metrics) == 0,
```
shapiromatron, line 76: I ended up changing this - which fixed the existing message we had I think it was trying to compare `[] == True` which must have been an error

shapiromatron (approved):
I took a different approach; it turns out we had a message already but it was broken because we had a typo in the logic passed from the template - if its ok, feel free to merge.

PR #822, opened 2023-05-03
Only renders 'no data available' when user is a team member or higher

dannypeterson:
We previously showed "none available" to the public for assessments if no data are available. This has been revised to only show no data available if you're able to edit content, otherwise, the header isn't shown.

**If data is available for a given extraction domain, no changes is made; it's displayed in either case.**

|Unauthenticated|Team Member|
|---|---|
|(empty)image|image|


shapiromatron (approved):
Back in your court @dannypeterson - I had to change things because with the initial implementation, even if there was data, it wasn't presented to the public unless a user had edit permission. I had to update the individual templates, many haven't been touched in years, so I ended up editing them a bit more to use some of the helper utilities we have in HAWC now...

PR #827, opened 2023-05-16
New sidebar display options

shapiromatron:
Add new sidebar display options to show/hide tables, visuals, and downloads. These can be enabled or disabled like the existing options for a given assessment.  By default, values are shown, but we do run a data migration to check if an assessment hasn't been updated in 6 months and data does not exist, to disable since it's unlikely that data will be added.

Screenshot 2023-05-16 at 4 22 51 PM


hawc/apps/assessment/forms.py
```diff
@@ -38,20 +38,22 @@ class AssessmentForm(forms.ModelForm):
     )
 
     class Meta:
-        exclude = (
```
shapiromatron, line 41: switched from exclude to fields (include), since it's easier to understand what's available.

hawc/apps/assessment/migrations/0036_assessment_enable_downloads_and_more.py
```diff
@@ -0,0 +1,54 @@
+# Generated by Django 4.2.1 on 2023-05-11 01:25
+
+from datetime import timedelta
+
+from django.db import migrations, models
+from django.db.models import Count
+from django.utils.timezone import now
+
+
+def fwd(apps, schema_editor):
```
shapiromatron, line 10: this sets older assessments to turn on/off the sidebar options, depending on if they have data in these comains

dannypeterson (approved):

PR #831, opened 2023-05-30
added header to Update Assessment Value page

dannypeterson:
assessment value header

### DIFF
--- a/hawc/apps/assessment/forms.py
+++ b/hawc/apps/assessment/forms.py
@@ -180,7 +180,7 @@ class AssessmentDetailForm(forms.ModelForm):
 
 
 class AssessmentValueForm(forms.ModelForm):
-    CREATE_LEGEND = "Create Assessment values"
+    CREATE_LEGEND = "Create Assessment Value"
     UPDATE_LEGEND = "Update Assessment Value"
     CREATE_HELP_TEXT = ""
     UPDATE_HELP_TEXT = "Update current assessment value."
--- a/hawc/apps/assessment/forms.py
+++ b/hawc/apps/assessment/forms.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 from textwrap import dedent
 
+import numpy as np
 from django import forms
 from django.conf import settings
 from django.contrib import admin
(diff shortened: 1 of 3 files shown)


hawc/apps/assessment/templates/assessment/assessmentvalue_form.html
```diff
@@ -3,7 +3,8 @@
 {% load crispy_forms_tags %}
 
 {% block content %}
-<div class="d-flex justify-content-end">
+<div class="d-flex justify-content-between">
+    <h3>Assessment Value for {{ assessment }}</h3>
```
shapiromatron, line 7: Just for consistency with other forms we typically do at the form-level. It might be better to add here: https://github.com/shapiromatron/hawc/blob/bfb056dffb26bfe9e66191ae254fe10b18e725c6/hawc/apps/assessment/forms.py#L254-L267


In addition, I think we'll need the same for the AssessmentDetails: https://github.com/shapiromatron/hawc/blob/bfb056dffb26bfe9e66191ae254fe10b18e725c6/hawc/apps/assessment/forms.py#L160

I think it should be as simple as adding `legend_text` for the update fields.

hawc/apps/assessment/forms.py
```diff
@@ -127,9 +128,8 @@ def helper(self):
 
 
 class AssessmentDetailForm(forms.ModelForm):
-    CREATE_LEGEND = "Add additional Assessment details"
-    CREATE_HELP_TEXT = ""
-    UPDATE_HELP_TEXT = "Update additional details for this Assessment."
+    LEGEND = "Additional Assessment Details"
```
shapiromatron, line 131: I ended up adding the same here as well; the text is generic enough it should work for creates and updates

hawc/apps/assessment/forms.py
```diff
@@ -245,6 +240,18 @@ def clean(self):
             if not cleaned_data.get("uncertainty"):
                 msg = "Required for Noncancer evaluation types."
                 self.add_error("uncertainty", msg)
+        if (
```
shapiromatron, line 243: added an additional data-check and a test - minor scope creep for this PR, but it was quicker to add than write another card :).

hawc/apps/assessment/templates/assessment/assessmentvalue_form.html
```diff
@@ -3,7 +3,7 @@
 {% load crispy_forms_tags %}
 
 {% block content %}
-<div class="d-flex justify-content-end">
+<div class="float-right">
```
shapiromatron, line 6: this fixes the extra vertical padding by floating to the right instead of a flex container

shapiromatron (approved):
Looks good! Made a few more edits for related changes (instead of making additional jira cards)  - if you're ok with the changes, feel free to merge.

PR #833, opened 2023-05-31
add flexible keyword highlighting

casey1173:
Adds flexible keyword highlighting, pulled from LLR.


hawc/apps/lit/migrations/0019_alter_literatureassessment_keyword_list_1_and_more.py
shapiromatron, line 1: Instead of making a new migration which doesn't change anything in the database, can you copy/paste the help text into the lit.0017 migration and then delete this migration?  Sorry about this; I wish django didn't make migrations like this.

frontend/lit/components/Reference.js
```diff
@@ -12,8 +12,7 @@ const markKeywords = (text, allSettings) => {
         const all_tokens = allSettings.set1.keywords
                 .concat(allSettings.set2.keywords)
                 .concat(allSettings.set3.keywords),
-            all_re = new RegExp(all_tokens.join("|"), "gim");
-
+            all_re = new RegExp(all_tokens.join("|").replace(/\*/g, ".*?"), "gim");
```
shapiromatron, line 15: I'm not sold on this idea, but what do you think about wrapping word boundaries? It's nice because it'll highlight the whole word, not just the phrase, and we may get fewer false-positives.

image

Without the word boundary, we're basically getting star-style matches all the time: 

image

Looping in @swatford  as well for thoughts...

shapiromatron (changes requested):
One comment for a change and a quick discussion. 

It'd be nice to be able to write a test for this too, I'm not sure how hard that would be, but I dont think it'd be hard; we'd have to move the regex parsing to frontend/shared/utils/helpers? and then you could write a test in frontend/tests/shared/utils/helpers.test.js

frontend/tests/shared/utils/helpers.test.js
```diff
@@ -1,4 +1,5 @@
 import {addOuterTag} from "shared/utils/_helpers";
+import helper from "shared/utils/helpers";
```
shapiromatron, line 2: I saw the CI error; you may need to add this to the `_helpers` file since I think the JS testing framework doesnt have access to a browser, and therefore doesnt have the external jQuery that we vendor in... @casey1173

shapiromatron (commented):
I think you may have to move the file for the tests to work, its a little hacky but if you move the function to _helpers and use that for the tests, and then in our helpers.js just import the function from _helpers, that should do it.  Since we're using jQuery as an "external" resource in webpack, our testing framework doesnt have access to it...

frontend/tests/shared/utils/helpers.test.js
casey1173: ah okay will do, thanks! Got confused by that one

frontend/shared/utils/_helpers.js
```diff
@@ -5,3 +5,38 @@ export const addOuterTag = function(html, tag) {
...
+};
+
+export const markText = function(text, settings) {
+    if (settings.keywords.length === 0) {
+        return text;
+    }
+    const re = new RegExp(
+        `<mark>(?<token>\\b${settings.keywords
+            .join("\\b|\\b")
+            .replace(/\*/g, "[\\S]*?")}\\b)</mark>`,
```
shapiromatron, line 34: This pattern is different than the all_re pattern above - I think that above one that concats the \b at the beginning and end seems better?   Could we make a n few function that build the regex given an array of keywords, that's used everywhere?

shapiromatron (commented):
One more hopefully minor change. We're building different regular expressions in different code, I think it should be refactored so that the same regex is constructed everywhere

frontend/shared/utils/_helpers.js
casey1173: the first regex matches just the keywords, and it's used to add the <mark> elements. this regex matches keywords with the <mark> element (and the beginning and end \b tokens are present, just bookended by the <mark> elements). we did this so users could use 'mark', 'positive', 'negative', etc. as keywords without creating regex conflicts. So I think these regex need to be different

shapiromatron (approved):
Looks great!

PR #839, opened 2023-06-01
update assessment filtering

casey1173:
## Changes
Updates assessment lists to use the new inline filter forms, and to show more data in the table.

FilterSet updates:
- BaseFilterset now allows for fields to specify an initial value; this is NOT recommended for the vast majority of filter options, but it is useful for order by, since a default/initial order is unavoidable. [See django-filter docs for more info](https://django-filter.readthedocs.io/en/main/guide/tips.html#using-initial-values-as-defaults)
- 'Select' widget fields in the inline form now show help text when hovered over (if available)
image

### Before
image

### After
- Inline filter form with increased filter capacity (name, author, year, assessment values, project type, report id, dtxsid, casrn)
- CASRN shown in table, when available
- Assessment objectives rendered as safe, allowing for formatting

image

### Before
image

### After
- Same changes as public assessment list
- Combined three separate lists based on assessment role into one list
- Users can also filter by assessment role on this page (manager, team member, reviewer)

image


shapiromatron (approved):
LGTM - @casey1173 I ended up making a number of additional minor revisions, could you please give it one more pass through prior to merging to make sure things look ok?

PR #840, opened 2023-06-04
epiv2 export - fix header/data column count mismatch

shapiromatron:
Fix 500 error when generating an export - instead of hard-coding 4 columns - generate number of columns based on header length (should be 6).


shapiromatron:
@casey1173 another quick review please - I think you independently fixed this one :)

PR #842, opened 2023-06-13
added dependencies to fix failed migrations

dannypeterson:
make sync-dev failed on manage migrate due to some migrations being out of order. Danny R helped fix the dependencies


shapiromatron (approved):
nice find @dannypeterson! we hit this last week too, LGTM (after checks are fixed)!

PR #843, opened 2023-06-14
added rob help text to page

dannypeterson:
- added help text to rob study eval update page
rob help text


shapiromatron (commented):
LGTM - I made it a collapsable div so you can view it when needed; when I tested it on a few pages, the instructions were really long

shapiromatron (approved):

PR #844, opened 2023-06-20
Better message when there is no study type

dannypeterson:
- added alert to www.haw.epa.gov/study/x/ page if no data metric is chosen
empty study type


shapiromatron (approved):
LGTM - I made a few minor tweaks - let me know if you're ok with the changes

PR #846, opened 2023-07-10
Exploratory heatmap changes

rabstejnek:
Made several changes to improve the exploratory heatmap experience:
- Used flexbox in the filter widget rows to create a more uniform look
- Overhauled filter widget selection process
  - When no selections are made, no filtering is done on that filter widget
  - Selections on a filter widget row now shows as an outline instead of a checkbox
  - Changed logic that hid filter widget rows without endpoints
    - Now the logic is if a selection is made in a widget, show all rows (needed for mutually exclusive filter widget rows). If a selection hasn't been made in a widget, automatically filter out the rows with no endpoints.
  - Removed select-all / select-none buttons from header
- Swapped position of row / column components on heatmap form

image


casey1173:
> @munnsmunns @casey1173 @swatford
> 
> Thoughts on any of these as an alternative to the eye icon? As a reminder, that icon will open a modal to show more details on something. I dont think we need to show the disabled icon anymore, just the active one.
> 
> * https://fontawesome.com/v4/icon/plus-circle
> * https://fontawesome.com/v4/icon/share-square-o
> * https://fontawesome.com/v4/icon/external-link

I ended up using the external link icon

shapiromatron (approved):
looks great @rabstejnek and @casey1173 . I think our users will find this much more intuitive.  Excluded all the fa-eye are gone and more consistent too.

PR #848, opened 2023-07-11
added minor critical sensitivity to constants

dannypeterson:
### Added Minor/Critical concerns (sensitivity) to judgement choices in study eval metric
- Created riskofbias custom template tag to load score options
- Added color shade preview in judgement choices
Screenshot 2023-07-12 133552


shapiromatron (approved):
LGTM - I fixed the typo and then added the new choices to the migrations. You could also type `makemigrations` and that would create a new migration, but since these changes dont actually alter the database at all, I just add them to existing migrations instead. That's just a strange behavior of mine; creating a new migration would be fine too :)

PR #849, opened 2023-07-11
remove source change for literature import update

shapiromatron:
A user shouldn't be able to change the source (HERO, PubMed) after an import has been executed. Also update the help text to explain why the fields are missing.

image


dannypeterson (approved):
LGTM! - ready to merge

PR #851, opened 2023-07-14
remove blog and myst-parser

shapiromatron:
Remove assessment.Blog and the myst-parser from requirements.  This is done to primarily to reduce [dependencies](https://github.com/executablebooks/MyST-Parser/blob/f4afeef1734a4e908670f67120640e6f9bf7308d/pyproject.toml#L36-L44) which are brought into HAWC from using the parser; and removing functional code that was unfortunately never really used.


dannypeterson (approved):
LGTM - no more instances of Blog or myst_parser found

PR #852, opened 2023-07-14
remove topic models

shapiromatron:
Remove topic models and related dependencies from hawc (scikit-learn, nltk, pyarrow). We incubated the topic model for a number of years, but there wasn't a large amount of interest, and significant work would be required to make it ready for production use.


dannypeterson (approved):
LGTM - ready to merge!

PR #853, opened 2023-07-14
remove R client

shapiromatron:
The R client hasn't been used or updated since it was originally written. It is no longer current and isn't tested, so it's been removed.


dannypeterson (approved):
LGTM - ready to merge

PR #855, opened 2023-07-14
epiv2 bulk cleanup

shapiromatron:
- added epiv2 models to bulk cleanup
- organized bulk cleanup page by app
- ui improvements
bulk cleanup
- removed html fields from text cleanup
- added an epiv2 migration; some CharFields could be null, we only want them to be empty strings
- added select related queryset methods, dramatically improving performance for API methods


frontend/shared/textCleanup/App.js
```diff
@@ -86,15 +88,31 @@ class App extends Component {
...
-                                        {d.count}&nbsp;{d.title}
-                                    </a>
-                                </li>
+                                <div key={app}>
+                                    <h4 className="mt-3">{app}:</h4>
+                                    <ul className="list-group">
+                                        {items
+                                            .filter(item => item.count > 0)
+                                            .map(d => {
+                                                return (
```
shapiromatron, line 101: I like this change, but I think it's breaking our integration test. To fix, you'll have to run the integration tests locally. I think you can do something like this:

```bash
playwright install --with-deps chromium
set INTEGRATION_TESTS=1
set PWDEBUG=0
py.test -sv tests/integration/test_cleanup.py
```

Then you can record the new thing; or just hand-edit the text to make edits. I don't really worry about re-recording the whole test; just the thing that failed.  In this case, I think the "text" is different than what it was before, since you switched the count to a badge

Adapted from https://github.com/shapiromatron/hawc/blob/aae68b1cf6fb746efeaf0a57a9611862cc27ae62/make.bat#L102-L107

shapiromatron (commented):
One more UI change, could we make the list sorted case insensitive. A good example might be the epi outcomes for 100500308; right now capital is first, lower case at the bottom. 

It might just be as easy as updating the sort on the store.groupedObjects

shapiromatron:
@dannypeterson LGTM - I made a few more edits:

- removed html fields from text cleanup
- added an epiv2 migration; some CharFields could be null, we only want them to be empty strings
- A few additional user-interface and ordering updates

If you're ok with the changes, please approve (I started this branch so I can't approve) - and then you can merge!

dannypeterson (approved):
LGTM, ready to merge!

PR #845, opened 2023-06-27
eco data pivots and exploratory heatmaps

casey1173:
Adds eco data pivots and exploratory heatmaps.
image

image

Adds eco result endpoint list
image

PR #832, opened 2023-05-30
epiv2 data pivots and exploratory heatmaps

shapiromatron:
Add the ability to create data pivots and exploratory heatmaps using data extracted in the epiv2 module.    

* We used the basic data export for data extraction data, and created a simple systematic evidence map (SEM) style study summary visual. It's likely that these exports will change in the future, and we'll need to make sure existing visuals are migrated
* Added an outcome list view for epiv2 data s well
* Added some API test helper utility methods to make testing datasets simpler

A future PR will improve the default exploratory heatmaps which are created on the "endpoint list" page. This PR implements the initial versions.

PR #861, opened 2023-07-22
refactor DataPivotExtension -> Interactivity package

shapiromatron:
Refactor DataPivotExtension to reduce the dependency with DataPivot visualizations. From an API perspective, multiple columns can now be bound to a single action.


shapiromatron:
Sorry for so many reviews @casey1173; I added this one so that I could do #832 ; but figured you may need it as well for #845

casey1173 (approved):

PR #862, opened 2023-07-25
Add API for creating Assessments

munnsmunns:
Add an admin only API endpoint for creating, updating, and deleting assessments.

### DIFF
--- a/hawc/apps/assessment/api/viewsets.py
+++ b/hawc/apps/assessment/api/viewsets.py
@@ -254,7 +254,6 @@ class DoseUnitsViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
 
 
 class Assessment(AssessmentEditViewSet):
-    http_method_names = METHODS_NO_PUT
     model = models.Assessment
     serializer_class = serializers.AssessmentSerializer
     assessment_filter_args = "id"
--- a/hawc/apps/assessment/api/viewsets.py
+++ b/hawc/apps/assessment/api/viewsets.py
@@ -253,10 +253,17 @@ class DoseUnitsViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
         return self.model.objects.all()
 
 
-class Assessment(AssessmentViewSet):
+class Assessment(AssessmentViewSet, AssessmentEditViewSet):
+    http_method_names = METHODS_NO_PUT
     model = models.Assessment
     serializer_class = serializers.AssessmentSerializer
     assessment_filter_args = "id"
+
+    def get_permissions(self):
+        if self.action in ["create", "update", "delete"]:
+            return [permissions.IsAdminUser()]
+        else:
+            return super().get_permissions()
 
     @action(detail=False, permission_classes=(permissions.AllowAny,))
     def public(self, request):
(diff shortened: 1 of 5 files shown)


hawc/apps/assessment/api/viewsets.py
```diff
@@ -440,6 +440,40 @@ def logs(self, request: Request, pk: int, type: str):
         return Response(export)
 
 
+class AssessmentAdmin(viewsets.ModelViewSet):
```
shapiromatron, line 443: can this be removed? I think it'd be better if we could change the 
`class Assessment(AssessmentViewSet)` to inherit from the `AssessmentEditViewSet` instead of the current `AssessmentViewSet` class?

And then you could just check on the create/update/delete methods that the user passes the isAdminUser check, otherwise 403.

That'd be much better since we wouldn't have an additional API ViewSet, it would be more DRY. 

Just riffing here (untested):


```python

class Assessment(AssessmentEditViewSet):
    def get_permissions(self):
        if self.action in ['create', 'update', delete']:
            permission_classes = [IsAdminUser]
        else:
            permission_classes = [AssessmentLevelPermissions]
        return [permission() for permission in permission_classes]
```

shapiromatron (commented):
Hoping that we can use the existing `Assessment` ViewSet instead of creating another - let me know if this approach wont' work

hawc/apps/assessment/api/viewsets.py
```diff
@@ -253,11 +253,18 @@ def get_queryset(self):
         return self.model.objects.all()
 
 
-class Assessment(AssessmentViewSet):
+class Assessment(AssessmentViewSet, AssessmentEditViewSet):
```
shapiromatron, line 256: remove AssessmentViewSet inheritance?

hawc/apps/assessment/serializers.py
```diff
@@ -21,8 +21,8 @@ class Meta:
 
 
 class AssessmentSerializer(serializers.ModelSerializer):
-    rob_name = serializers.CharField(source="get_rob_name_display")
-    dtxsids = DSSToxSerializer(many=True)
+    rob_name = serializers.CharField(source="get_rob_name_display", read_only=True)
+    dtxsids = DSSToxSerializer(many=True, read_only=True)
```
shapiromatron, line 25: does this mean that you can't add dtxsids? Could we do that trick where this one is read_only and then we have another `dtxsids_ids` which is write_only?

```python
dtxsids_ids = serializers.PrimaryKeyRelatedField(many=True, write_only=True)
```

tests/hawc/apps/assessment/test_api.py
```diff
@@ -175,3 +175,53 @@ def test_bad_request(self, db_keys):
...
+            "assessment_objective": "<p>Test.</p>",
+            "authors": "<p>Test.</p>",
+            "noel_name": 0,
+            "rob_name": 0,
+            "editable": "on",
+            "project_manager": ["2"],
+            "team_members": ["1", "2"],
+            "reviewers": ["1"],
+            "epi_version": 1,
+            "dtxsids": ["DTXSID7020970"],
```
shapiromatron, line 197: I think the dtxsids is ignored with this request, is that right?

shapiromatron (changes requested):
Looking good! A few revisions please

hawc/apps/assessment/api/viewsets.py
munnsmunns: isn't it used for the other actions (public, endpoints, logs)?

shapiromatron: No, I dont think so, the AssessmentEditViewSet has all the same configuration. https://github.com/shapiromatron/hawc/blob/eb4aafd3c8eeb1a35e609479f1a20420266d5d93/hawc/apps/assessment/api/viewsets.py#L182 

The inheritance/naming pattern with these is confusing for sure.

hawc/apps/assessment/api/viewsets.py
```diff
@@ -254,7 +254,6 @@ def get_queryset(self):
 
 
 class Assessment(AssessmentEditViewSet):
-    http_method_names = METHODS_NO_PUT
```
shapiromatron: I removed this; it was already in the parent class

tests/hawc/apps/assessment/test_api.py
```diff
@@ -190,9 +190,10 @@ def test_permissions(self, db_keys):
             "noel_name": 0,
             "rob_name": 0,
             "editable": "on",
-            "project_manager": ["2"],
-            "team_members": ["1", "2"],
-            "reviewers": ["1"],
+            "creator": 1,
```
shapiromatron, line 193: changed from strings to ints

shapiromatron (approved):
looks good!

PR #863, opened 2023-07-26
removed float right from <ActionsButton />

dannypeterson:
- removed class float right from actions button js component


shapiromatron (approved):
Looks good, I only made one change to remove the default btn-primary since that is modified in some cases

PR #864, opened 2023-07-26
added pubertal fields to ani constants

dannypeterson:
- added pubertal field to ExerimentType
- added Juvenile field to Lifestage exposed/assessed
- updated migration


shapiromatron (approved):
LGTM!

PR #847, opened 2023-07-10
update python/js packages

shapiromatron:
Update python and JS packages based to recent versions where possible and relatively simple.

A few minor additional code changes:

* updated ruff made some minor code formatting improvements
* updated code in a few places to remove deprecation warnings

---

After this is merged into main, you'll need to update your developer environment:

* update yarn to version 3: https://yarnpkg.com/getting-started/migration#step-by-step
* run `make sync-dev` to install the latest versions of python and JS packages

PR #867, opened 2023-07-28
rollback yarn

shapiromatron:
Rollback yarn from 3.6.1 -> 1.22.19; users were having issues setting up the developer environment, as well as issues in CI and test deployments using version 3 of the package manager.

This PR reverses yarn update in #847. It doesn't change any installed package versions.


munnsmunns (approved):
LGTM

PR #869, opened 2023-07-30
epiv2 - exploratory heatmap defaults

shapiromatron:
Add a few new default visualizations for the epiv2 exploratory heatmaps. This replaces the placeholders that were previously for epi results. In the future, this will be further updated once we have more data in an assessment to better understand which visuals may be valuable.

TIL - vsocde has a command to "convert to camel case" on any text selection - this was helpful in making all the python snake case variable names to javascript camel case. https://github.com/microsoft/vscode/issues/149824


casey1173 (approved):
Looks good!

PR #870, opened 2023-07-31
rewrite epiv2 exports

dannypeterson:
Rewrite how the epiv2 export is calculated to combine multiple django ORM queries into a single request. 

No changes were made to the export format of the data itself.

This PR created two new base classes, `ModelExport`, and `Exporter`, that we'll use throughout the application to update our modernize our existing export formats.


hawc/apps/epiv2/api.py
```diff
@@ -30,6 +30,22 @@ class EpiAssessmentViewSet(BaseAssessmentViewSet):
...
+        export = FlatExport(df=qs.complete_df(), filename=f"epidemiological-export-{assessment.id}")
+        return Response(export)
+
+    @action(
+        detail=True,
+        url_path="export2",
+        action_perms=AssessmentViewSetPermissions.CAN_VIEW_OBJECT,
+        renderer_classes=PandasRenderers,
+    )
+    def export2(self, request, pk):
```
rabstejnek, line 48: Remove this old path?

hawc/apps/epiv2/managers.py
```diff
@@ -193,6 +194,148 @@ def complete(self):
...
+            "chemical-name": "exposure_level__chemical__name",
+            "chemical-DTSXID": "exposure_level__chemical__dsstox__dtxsid",
+            "chemical-created": "exposure_level__chemical__created",
+            "chemical-last_updated": "exposure_level__chemical__last_updated",
+            # exposure
+            "exposure-pk": "exposure_level__exposure_measurement__pk",
+            "exposure-name": "exposure_level__exposure_measurement__name",
+            "exposure-measurement_type": "exposure_level__exposure_measurement__measurement_type",
+            "exposure-biomonitoring_matrix": sql_display(
+                "exposure_level__exposure_measurement__biomonitoring_matrix",  # todo fix default display "?"
```
rabstejnek, line 233: Fix this TODO? Maybe have an optional `default` argument for `sql_display` that defaults to "?", and then specify it here as whatever you think is more appropriate (empty string maybe).

hawc/apps/epiv2/managers.py
```diff
@@ -193,6 +194,148 @@ def complete(self):
             "factors",
         ).prefetch_related("design__countries")
 
+    def complete_df(self) -> pd.DataFrame:
+        mapping = {
+            # study
+            **study_df_mapping("study-", "design__study__"),
+            # design
+            "design-pk": "design__pk",
+            # "design-url": "design__get_absolute_url", #todo make a sql function or remove?
```
rabstejnek, line 203: If we want to include this and build it in our SQL query, [CONCAT](https://docs.djangoproject.com/en/4.2/ref/models/database-functions/#concat) is probably our best best.

rabstejnek (changes requested):
Looking good! There were some commented TODOs left in here that I think should be addressed before merge, which makes up the bulk of the comments I left. Also once those are wrapped up you should regenerate the VCR fixture for the epi api test so that it passes.

hawc/apps/epiv2/managers.py
rabstejnek: Maybe a function that could be called like `sql_format("/path/{}/{}/","field1","field2")`, where the first argument is a format string, and subsequent arguments are the fields to use in those placeholders?

shapiromatron: I like that idea and the name.

hawc/apps/common/exports.py
```diff
@@ -0,0 +1,108 @@
+import pandas as pd
+
+
+class Module:
```
shapiromatron, line 4: rename base class? Maybe `ModelExportFactory`?

hawc/apps/common/exports.py
```diff
@@ -0,0 +1,108 @@
...
+        # handle any includes/excludes
+        if self.include or self.exclude:
+            annotation_map = {
+                k: v for k, v in annotation_map.items() if k in self.value_map.values()
+            }
+
+        self._annotation_map = annotation_map
+        return self._annotation_map
+
+    def _get_value_map(self):
```
shapiromatron, line 58: I think these are our public API for this class, so I'd remove the underscores. We should also add type annotations for the public methods too

hawc/apps/common/exports.py
```diff
@@ -0,0 +1,108 @@
...
+        # if self.query_prefix:
+        #    return qs.select_related(self.query_prefix)
+        return qs
+
+    def prepare_queryset(self, qs):
+        qs = self._select_related(qs)
+        qs = self._annotate(qs)
+        return qs
+
+    def prepare_df(self, df):
```
shapiromatron, line 82: can you convert docstrings to google style? or at least triple strings quotes

hawc/apps/common/models.py
```diff
@@ -534,6 +534,17 @@ def sql_display(name: str, Choice: type[Choices]) -> Case:
     )
 
 
+def sql_format(format_str: str, *field_params):
```
shapiromatron, line 537: needs docstring and type annotations

hawc/apps/study/exports.py
```diff
@@ -0,0 +1,60 @@
...
+            "doi": str_m2m(
+                query_prefix + "identifiers__unique_id",
+                filter=Q(**{query_prefix + "identifiers__database": ReferenceDatabase.DOI}),
+            ),
+            "coi_reported_display": sql_display(query_prefix + "coi_reported", CoiReported),
+            "url": sql_format("/study/{}/", query_prefix + "pk"),
+        }
+
+    def prepare_df(self, df):
+        for key in [f"{self.key_prefix}pubmed_id", f"{self.key_prefix}hero_id"]:
```
shapiromatron, line 56: add a helper method to base class `get_column_name()` that returns `f"{self.key_prefix}{name}` or something similar?

hawc/apps/common/exports.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def get_df(self, qs):
+        qs = self.prepare_queryset(qs)
+        df = pd.DataFrame(
+            data=qs.values_list(*self.value_map.values()), columns=list(self.value_map.keys())
+        )
+        return self.prepare_df(df)
+
+
+class Exporter:
```
shapiromatron, line 96: Instead of constructing instances of the class in the init, should we just require subclasses to implement directly? As an example:


```python
class Exporter:
    def build_modules(self) -> list[Module]:
        ...

    def get_df(self, qs: QuerySet) -> pd.DataFrame:
        self._modules = self.build_modules()
        for module in self._modules:
            qs = module.prepare_queryset(qs)
        values = [value for module in self._modules for value in module.value_map.values()]
        keys = [key for module in self._modules for key in module.value_map.keys()]
        df = pd.DataFrame(data=qs.values_list(*values), columns=keys)
        for module in self._modules:
            df = module.prepare_df(df)
        return df

class EpiExporter(Exporter):
    def build_modules(self) -> list[Module]:
        return [
            StudyModule("study", "design__study"),
            DesignModule("design", "design"),
            ChemicalModule("chemical", "exposure_level__chemical"),
            ExposureModule("exposure", "exposure_level__exposure_measurement"),
            ExposureLevelModule("exposure_level", "exposure_level"),
            OutcomeModule("outcome", "outcome"),
            AdjustmentFactorModule("adjustment_factor", "factors"),
            DataExtractionModule("data_extraction", ""),
        ]
```

Rename, maybe `ExportFactory`?

shapiromatron (commented):
Ok, you've convinced me, this is a nice pattern! Some feedback on the API and layout

shapiromatron (approved):
LGTM! I made two minor changes, if you're ok with these. I figure in the future we'll get rid of the FlatFileExporter and just use this Export class directly; you can see I was able to do so in the epiv2 api (but couldn't get rid of it b/c the DataPivotQuery still uses the old one).

* 0c2d9755 add special comment (17 seconds ago) {Andy Shapiro}
* 8f2abed4 updates (11 minutes ago) {Andy Shapiro}

rabstejnek (approved):

PR #871, opened 2023-08-01
squash assorted bugs

shapiromatron:
A collection of small bugfixes and minor changes from recent application review. It may be best to review this PR commit by commit; each commit fixes a different bug.

- Update template for newly required environment variables (`HAWC_CSRF_TRUSTED_ORIGINS`)
- Fix bug in parsing of assessment log report type
- Remove hawc exception `ClassConfigurationException`; use built in `django.core.exceptions.ImproperlyConfigured` instead
- Fix bugs in assessment form validation to soft-check required fields if they're not submitted (`AssessmentValueForm`, `AssessmentDetailForm`, `lit.ImportForm`)
- If assessment cannot be resolved for logs; prevent access
- Catch errors where invalid filterset requests are made
- Check that admin dashboard htmx views only return for valid HTTP method types
- Further restrict href validation - for local links, require trailing slash and no file extensions
- Catch error where admin dashboard matrix has no data and cannot be returned
- For PubMed parsing, make author initials optional
- Only allow creation of summary tables when a valid table type is passed
- Require `jinja2` explicitly for pandas html generation; was previously a transitive dependency
- Add tests for admin dashboard views
- fix issue with troublesome `bmd` test in CI?

---

In addition, I kept having issues in VSCode where I'd be writing a method, and hit save, and then ruff [F841](https://beta.ruff.rs/docs/rules/unused-variable/) would delete my variable.  I updated our pyproject.toml so that our linter still requires F841 to pass, but it won't try to autofix these issues.


hawc/apps/summary/views.py
```diff
@@ -163,7 +163,11 @@ class SummaryTableCreate(BaseCreate):
 
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
-        kwargs["table_type"] = self.kwargs["table_type"]
+        try:
+            table_type = constants.TableType(self.kwargs["table_type"])
```
rabstejnek, line 167: Use `.get` or catch `KeyError`? Not sure if this is a potential bug in the security scans.

shapiromatron: We know that `table_type` exists b/c it's in the url pattern, so the only way it would match this route is if it's valid. The issue we observed were cases where an id of 999 or something like that was passed.

hawc/apps/assessment/actions/audit.py
```diff
@@ -239,10 +239,11 @@ def get_riskofbias_queryset(self):
         return domain_qs | metric_qs | score_qs
 
     def get_queryset(self):
-        audit_type = self.type
-        if audit_type == AuditType.EPI:
+        audit_type = self.type.value
+        if audit_type == "epi":
             audit_type = "epiv1" if self.assessment.epi_version == EpiVersion.V1 else "epiv2"
```
rabstejnek, line 244: Was the issue here that we were trying to grab the `value` attribute from a string instead of the enum?

shapiromatron: yes, we previously just grabbed the enum.value, but with this one case, we actually need another check to determine which string to use

hawc/apps/assessment/models.py
```diff
@@ -746,6 +747,8 @@ def get_dict(self):
         }
 
     def get_assessment(self):
+        if self.content_object is None:
```
rabstejnek, line 750: It may be better to do `hasattr(self.content_object,"get_assessment")`

hawc/apps/common/validators.py
```diff
@@ -113,6 +113,8 @@ def validate_hyperlink(value: str, raise_exception: bool = True) -> bool:
     valid = url.scheme in valid_scheme and (
         url.netloc == "" or any(url.netloc.endswith(ending) for ending in valid_netloc_endings)
     )
+    if url.netloc == "" and ("." in url.path or not url.path.endswith("/")):
```
rabstejnek, line 116: What's the reasoning for disallowing filename extensions? At the very least should we allow for `.html`?

hawc/apps/assessment/models.py
shapiromatron: I tried that! It's failing when in some cases self.content_object is None (if the object is deleted I think). I think this way may be better b/c it gives us more data that we can't retrieve a `content_object`

hawc/apps/common/validators.py
shapiromatron: There's no valid URL on the hawc website that doesn't end with a trailing slash or a filename extension. Many times security bots try to add paths like this based on old php applications that weren't secured well and would allow arbitrary execution or retrieval of files on disk.

rabstejnek (approved):
LGTM!

PR #872, opened 2023-08-02
swap log form to inline

casey1173:
Swap log filter form to a filterset, with expandable inline UI

before:
image

after: 
image


shapiromatron (approved):
LGTM; nice find

PR #874, opened 2023-08-05
Data Pivot visualization updates

shapiromatron:
Improve the flexibility of the data pivot plotting component to allow a few new styles for data pivots.

## 1. Arrows when confidence intervals extend beyond plot domain.  

If a x-axis domain is restricted to be less than the entire domain, arrows are shown. Hovering over the bar will show upper and lower values.

**Screenshots:**

image

image

image

## 2. Configure plot background colors and gridlines.

 Added more flexibility to show/hide background color, gridlines, and text grid. The default layout is unchanged.

Default layout (gridlines and background); unchanged from previous versions:
image

Minimal (no gridlines, no background):
image

Extended text background (text background, x-axis gridlines):
image

## 3. `x-axis` customization

Previously, when setting an x-axis range, we'd automatically assume that you'd want to show around 10 ticks (numbers on the x-axis), but would be flexible to show more or less depending on the range. Now, if you'd like more control, you can set the domain to be forced, and set the number of ticks.  This will allow full control of setting the x-axis bounds and layout.

**With force domain:**

image

image

**Without force domain:**

image
image


casey1173:
the x-axis force domain option does not work with log scale FWIW--is this a bug or as intended? if intended, should we include a help text pop up mentioning it?

shapiromatron:
> the x-axis force domain option does not work with log scale FWIW--is this a bug or as intended? if intended, should we include a help text pop up mentioning it?

Updated in c0baf0c8619ee0ee29d447463907349ff1ecd07d? I don't think it's an easy fix to enforce w/ a logscale, now it just toggles show/hide the force depending on if logscale is selected. Does that work?

casey1173 (approved):

PR #875, opened 2023-08-09
fixes from security scan

munnsmunns:
* Use soft .get for accessing cleaned_data
 * Use TypedChoiceField instead of IntegerField with a select widget
 * Wrap autocomplete in a try/except for value error


shapiromatron:
@munnsmunns - is this PR ready?

shapiromatron (approved):
LGTM!

PR #877, opened 2023-08-11
Improve reference full text search

munnsmunns:
Adds use of an index and search vectors to the Reference model and filterset to improve reference searches.

Example: searching by author name and year at the same time
Before:
image

After:
image

### DIFF
--- a/hawc/apps/animal/models.py
+++ b/hawc/apps/animal/models.py
@@ -1235,7 +1235,7 @@ class ConfidenceIntervalsMixin:
 
             elif data_type == constants.DataType.DICHOTOMOUS:
                 if i == 0:
-                    i_0 = eg["incidence"]
+                    i_0 = eg.get("incidence", 0)
                     n_0 = eg["n"]
                 if i_0 is not None and i_0 > 0:
                     i_1 = eg["incidence"]


hawc/apps/lit/managers.py
```diff
@@ -559,6 +559,11 @@ def global_df(self) -> pd.DataFrame:
         ).values_list(*mapping.values())
         return pd.DataFrame(list(qs), columns=list(mapping.keys()))
 
+    def full_text_search(self, search_text):
```
shapiromatron, line 562: I think we need to convert to a SearchQuery too, so the words can be stemmed like the index: https://docs.djangoproject.com/en/4.2/ref/contrib/postgres/search/#searchquery.

hawc/apps/lit/managers.py
```diff
@@ -559,6 +559,11 @@ def global_df(self) -> pd.DataFrame:
...
+        return self.annotate(search=constants.REFERENCE_SEARCH_VECTOR).filter(
+            Q(search=search_text) | Q(identifiers__unique_id=search_text)
```
shapiromatron, line 564: I'd pull the `| Q(identifiers__unique_id=search_text)`, and maybe add a **filters argument to the func, and then add this additional item via a func parameter

hawc/apps/lit/constants.py
```diff
@@ -32,3 +33,7 @@ class SearchType(models.TextChoices):
 DOI_EXACT = re.compile(r"^10\.\d{4,9}/[^\s]+$")
 DOI_EXTRACT = re.compile(r"10\.\d{4,9}/[^\s]+")
 DOI_EXAMPLE = "10.1234/s123456"
+
+REFERENCE_SEARCH_VECTOR = SearchVector(
+    "abstract", "title", "authors", "year", "journal", config="english"
```
shapiromatron, line 38: its redundant, but I think we should add `authors_short` too; sometimes people put other stuff in that one

shapiromatron (commented):
Looking great so far! Can you add a few simple unit tests that just demonstrate it working?

hawc/apps/lit/managers.py
munnsmunns: should we make the search_type="websearch" so that users can add AND/OR operators to the search?

munnsmunns: We also have the option to order by the relevance in the search filter. Would that be good to add in this card?

hawc/apps/lit/managers.py
shapiromatron: I was thinking something like this:

```python
def full_text_search(self, search_text, filters=None):
    if filters is None:
         filters = !()
    filters |= Q(search ...)
         
    return self.annotate(search=constants.REFERENCE_SEARCH_VECTOR).filter(filters)
```

Since i would expect a full_text_search to also look at IDs necessarily

hawc/apps/lit/managers.py
shapiromatron: we can try websearch, as long as the a simple "foo 2012" search still works like we'd expect.  I'd try to order by relevance too if you can; though i think the order by may interfere w/ the order by in our filterset

shapiromatron (changes requested):
It looks good, but when I tried it out, it's really slow. I don't think the query is using the index - did you investigate to see?

You can tell I think by looking at the django debug toolbar and looking at the SQL query plan.   I wonder if we need to separate the two queries for text and identifiers and then run qs.union to pull them together?

shapiromatron (approved):
Got it working using the index by removing the reference identifiers. We have a custom field for them to search for those, so I think it's ok.

For example, this searches 30k references in <10ms; you can see from the django debug toolbar the query plan uses the index for the search:

```

Aggregate  (cost=108.86..108.87 rows=1 width=8) (actual time=0.928..0.929 rows=1 loops=1)
--
->  Unique  (cost=108.81..108.85 rows=1 width=1357) (actual time=0.910..0.922 rows=4 loops=1)
->  Sort  (cost=108.81..108.82 rows=1 width=1357) (actual time=0.909..0.910 rows=4 loops=1)
Sort Key: lit_reference.id, lit_reference.title, lit_reference.authors_short, lit_reference.authors, lit_reference.year, lit_reference.journal, lit_reference.abstract, lit_reference.full_text_url, lit_reference.created, lit_reference.last_updated, lit_reference.block_id, (to_tsvector('english'::regconfig, ((((((((((COALESCE(lit_reference.abstract, ''::text) \|\| ' '::text) \|\| COALESCE(lit_reference.title, ''::text)) \|\| ' '::text) \|\| COALESCE(lit_reference.authors, ''::text)) \|\| ' '::text) \|\| COALESCE(lit_reference.authors_short, ''::text)) \|\| ' '::text) \|\| COALESCE((lit_reference.year)::text, ''::text)) \|\| ' '::text) \|\| COALESCE(lit_reference.journal, ''::text))))
Sort Method: quicksort  Memory: 27kB
->  Bitmap Heap Scan on lit_reference  (cost=36.13..108.80 rows=1 width=1357) (actual time=0.605..0.816 rows=4 loops=1)
Recheck Cond: (to_tsvector('english'::regconfig, ((((((((((COALESCE(abstract, ''::text) \|\| ' '::text) \|\| COALESCE(title, ''::text)) \|\| ' '::text) \|\| COALESCE(authors, ''::text)) \|\| ' '::text) \|\| COALESCE(authors_short, ''::text)) \|\| ' '::text) \|\| COALESCE((year)::text, ''::text)) \|\| ' '::text) \|\| COALESCE(journal, ''::text))) @@ '''coeffici'' & ''monsanto'''::tsquery)
Filter: (assessment_id = 1053)
Heap Blocks: exact=4
->  Bitmap Index Scan on search_vector_idx  (cost=0.00..36.12 rows=17 width=0) (actual time=0.488..0.488 rows=4 loops=1)
Index Cond: (to_tsvector('english'::regconfig, ((((((((((COALESCE(abstract, ''::text) \|\| ' '::text) \|\| COALESCE(title, ''::text)) \|\| ' '::text) \|\| COALESCE(authors, ''::text)) \|\| ' '::text) \|\| COALESCE(authors_short, ''::text)) \|\| ' '::text) \|\| COALESCE((year)::text, ''::text)) \|\| ' '::text) \|\| COALESCE(journal, ''::text))) @@ '''coeffici'' & ''monsanto'''::tsquery)
Planning Time: 3.465 ms
Execution Time: 1.109 ms
```

PR #879, opened 2023-08-21
disable login on some hosts via settings

shapiromatron:
Disable myuser.HawcLoginView and myuser.ExternalAuth routes if specified in `HAWC_DISABLED_LOGIN_HOSTS`. Some domains in a production environment may not be configured for login with external authentication providers, so this prevents attempting to login which results in failure. It redirects back to the home page with a django message describing the reason for failure.

image


munnsmunns (approved):
Looks good to merge!

PR #881, opened 2023-08-23
server reloading in dev - django-browser-reload

shapiromatron:


shapiromatron:
@casey1173 - might be good to see how this one works on the PR you're currently working on...

PR #887, opened 2023-08-29
full text urls view - improve validation

shapiromatron:
Add new validation rules for full text upload. Spent some time writing some validation tests for the excel file, and add a ` df_to_form_data` helper method for testing general reuse.


dannypeterson (approved):
lgtm!

PR #891, opened 2023-08-31
UDF - CRUD + schema preview

munnsmunns:
Adds a preview for dynamic form schemas to the UDF form, and adds basic Detail and List views for UDFs.


hawc/apps/udf/views.py
```diff
@@ -1,9 +1,31 @@
...
+class UDFDetailView(LoginRequiredMixin, DetailView):
+    template_name = "udf/udf_detail.html"
+    model = models.UserDefinedForm
+
+    def get_context_data(self, **kwargs):
+        form = dynamic_forms.Schema.parse_obj(self.object.schema).to_form()
+        kwargs.update(form=form)
+        return super().get_context_data(**kwargs)
+
+    def user_can_edit(self):
```
shapiromatron, line 27: I didn't know that `view` is available in a template context, thanks for showing me this!

hawc/apps/udf/forms.py
```diff
@@ -23,19 +35,38 @@ def __init__(self, *args, **kwargs):
...
-            cancel_url=cancel_url,
-            submit_text="Create Form",
+        cancel_url = reverse("udf:udf_list")
+        form_actions = [
+            cfl.Submit("save", "Save"),
+            cfl.HTML(f'<a role="button" class="btn btn-light" href="{cancel_url}">Cancel</a>'),
+        ]
+        legend_text = "Update a custom form" if self.instance.id else "Create a custom form"
+        helper = BaseFormHelper(self)
+        helper.layout = cfl.Layout(
```
shapiromatron, line 53: this looks very unstyled - that's by design right? Are you thinking we'll come back later and add UI polish + tests?  Makes sense to me, just want to confirm that's what you're thinking...

shapiromatron (approved):
LGTM - assuming I'm interpreting this work as you intended (see comments)

hawc/apps/udf/forms.py
munnsmunns: Yup that's what I had in mind

hawc/apps/udf/views.py
munnsmunns: I happened to notice it when looking at the context in the debug menu for the template, it turned out to be really useful here!

hawc/apps/udf/views.py
```diff
@@ -31,6 +31,7 @@ def user_can_edit(self):
 class CreateUDFView(LoginRequiredMixin, MessageMixin, CreateView):
     template_name = "udf/udf_form.html"
     form_class = UDFForm
+    success_url = reverse_lazy("udf:udf_list")
```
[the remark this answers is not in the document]
munnsmunns: ah yes I meant to

PR #894, opened 2023-09-11
Fix data pivot file upload form javascript

rabstejnek:
There was a bug in the Django template for the data pivot file upload form that was breaking the javascript that added auto-slugification and quillified the caption field. This PR fixes it by adding a default value when declaring the javascript variables instead of having the template leave the declared value blank.

### DIFF
--- a/hawc/apps/summary/templates/summary/datapivot_form.html
+++ b/hawc/apps/summary/templates/summary/datapivot_form.html
@@ -13,7 +13,7 @@
   {{ smart_tag_form.media }}
   <script type="text/javascript">
     $(document).ready(function () {
-      let js_units_choices = {{form.js_units_choices|safe}},
+      let js_units_choices = {% autoescape off %}{% firstof form.js_units_choices "[]" %}{% endautoescape %},
         epi_version = {{assessment.epi_version}},
         printf = window.app.HAWCUtils.printf;
 


hawc/apps/summary/templates/summary/datapivot_form.html
```diff
@@ -13,7 +13,7 @@
   {{ smart_tag_form.media }}
   <script type="text/javascript">
     $(document).ready(function () {
-      let js_units_choices = {{form.js_units_choices|safe}},
+      let js_units_choices = {% autoescape off %}{% firstof form.js_units_choices "[]" %}{% endautoescape %},
```
shapiromatron, line 16: would this work instead? A little less noisy w/ the autoescape on/off.

`{% firstof form.js_units_choices|safe "[]" %}`

https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#firstof

shapiromatron (approved):
One proposed change to simplify template - nice job finding root-cause

hawc/apps/summary/templates/summary/datapivot_form.html
shapiromatron: Actually, cancel that. After testing it doesn't work. The way you did actually works!

PR #895, opened 2023-09-12
add download button to filter list UIs

casey1173:
image


shapiromatron (approved):
LGTM - I changed the button to an actions pull-down - if you're ok w/ the changes, ok to merge!

I should mention, I hope in the future w/ our dashboardification we may have other "dashboards" that you could use in addition to data exports, which might be good in the actions menu. That's my thinking as to why @casey1173

PR #896, opened 2023-09-14
Visual and Data Pivot - read API

casey1173:
Previously, API for Visual and DataPivotQuery were read-only. Now, you can create, update, and delete as well. Validation is not performed on the settings or prefilters, so it is possible to submit visuals that may not work.


casey1173:
@shapiromatron passing the buck!

shapiromatron (approved):
This looks great! However, I made some changes upstream in #901 with how we save the Visual settings to save it as a JSON field instead of a string field. Please review that PR first and merge it in if you're ok with it. Assuming it looks good and you're ok with the changes I made here, we can merge this one too.

casey1173:
Ok, changes LGTM @shapiromatron

PR #898, opened 2023-09-18
add Ns to endpoint level export

shapiromatron:
Add an `N` column to the endpoint row level exports in HAWC:

image


rabstejnek (approved):

hawc/apps/animal/exports.py
```diff
@@ -647,6 +648,9 @@ def getDoses(doses, unit):
             doses = [d["dose"] for d in doses if d["dose_units"]["name"] == unit]
             return [f"{d:g}" for d in doses]
 
+        def getNs(groups):
+            return [f"{grp['n'] or ''}" for grp in groups]
```
rabstejnek, line 652: Should we do `f"{grp['n'] if grp['n'] is not None else ''}"` instead, to preserve the case where `n` is 0?

shapiromatron: thanks, I'll change.

PR #899, opened 2023-09-21
Create models/views for binding UDFs to models and tags

munnsmunns:
Adds basic CRUD views for two new models, ModelBinding and TagBinding, which link UDFs to a model or tag.


hawc/apps/udf/constants.py
```diff
@@ -0,0 +1,5 @@
+from django.db import models
+
+
+class ContentTypeChoices(models.IntegerChoices):
```
shapiromatron, line 4: I think it may be better to just filter the ContentType queryset in the form; we can add additional models over time.

```python

from django.db.models import Case, F, Q, QuerySet, Value, When
from django.db.models.functions import Concat

ContentType.objects.annotate(
   full_name=Concat(F('app_label'), Value('.'), F('model'))
).filter(
   full_name__in=['animal.endpoint', 'study.study']
)
```

shapiromatron (approved):
LGTM - I added a button in the assessment details actions menu, and a list view to view all the bindings for a given assessment. If you're ok with the changes, I"m ok merging!

PR #900, opened 2023-09-21
remove unused code

shapiromatron:
Remove old code; we previously had untested code that would attempt to clone data across assessments, but this has largely removed. Found a few pieces that haven't yet been removed.


shapiromatron:
@munnsmunns - this should be a quick review I hope; I just found some dead code to remove...

munnsmunns (approved):
LGTM!

PR #904, opened 2023-09-26
Update experiment label

dannypeterson:
Changed "comments" key in column name:
image

Updated "help text":
image

### DIFF
--- a/frontend/animal/Experiment.js
+++ b/frontend/animal/Experiment.js
@@ -70,7 +70,7 @@ class Experiment {
             .add_tbody_tr("Animal diet", this.data.diet)
             .add_tbody_tr("Guideline compliance", this.data.guideline_compliance)
             .add_tbody_tr(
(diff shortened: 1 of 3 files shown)


frontend/animal/Experiment.js
```diff
@@ -69,7 +69,10 @@ class Experiment {
             .add_tbody_tr("Vehicle", this.data.vehicle)
             .add_tbody_tr("Animal diet", this.data.diet)
             .add_tbody_tr("Guideline compliance", this.data.guideline_compliance)
-            .add_tbody_tr("Description and animal husbandry", this.data.description);
+            .add_tbody_tr(
+                "Comments (eg. description, animal husbandry, etc.)",
```
shapiromatron, line 73: I think we should still call this one comments, but we want the form view to have the extra help text.

shapiromatron (commented):
I think we'll also need to edit the update experiment form view to add the example information

shapiromatron (approved):

PR #905, opened 2023-09-26
improve view/form test code coverage

BerkowitzMaxICF:
Write basic "smoke" tests for most GET views in the application. These tests assert that the views render a 200 when accessed by a user with an authorized user, which helps raise confidence that the majority of "happy" path code is being covered by our test suite.

With this PR, overall code coverage in the application was increased by around 8% (to 88%).


shapiromatron (commented):
This looks like a good start, we'll definitely keep the current work that you started. However, with the remaining hours, I think we should pivot to develop tests that are more similar to [smoke tests](https://en.wikipedia.org/wiki/Smoke_testing_(software)). For smoke tests, our main goal is to just ensure that the test doesn't throw any HTTP errors essentially. This will give us confidence that when we make changes, any view that previously rendered will continue to render and we're not introducing any broad regressions into the software.

To that end I propose that we add checks using the [check_200](https://github.com/shapiromatron/hawc/blob/03ec1ee3dfbc886d4d604cd180d8dbbafcbfeb49/tests/hawc/apps/test_utils.py#L74) function for views throughout the site to ensure that it properly returns a valid html response with a 200 code. We could try it using the project manager user (falling back to admin where apprpriate), to simulate a standard user.

I looked at the coverage report, and and saw the most of the views in the animal app were not ever executed in our test suite. So, I wrote a simple test that would go to a URL for each page type as a project manager, and make sure that the response was a valid 200 response. The test looks like this:

```python
import pytest
from django.urls import reverse

from ..test_utils import check_200, get_client


@pytest.mark.django_db
def test_smoke_get():
    client = get_client("pm")
    urls = [
        reverse("animal:experiment_new", args=(1,)),
        reverse("animal:experiment_copy_selector", args=(1,)),
        reverse("animal:experiment_detail", args=(1,)),
        reverse("animal:experiment_update", args=(1,)),
        reverse("animal:experiment_delete", args=(1,)),
        reverse("animal:animal_group_new", args=(1,)),
        reverse("animal:animal_group_copy_selector", args=(1,)),
        reverse("animal:animal_group_detail", args=(1,)),
        reverse("animal:animal_group_update", args=(1,)),
        reverse("animal:animal_group_delete", args=(1,)),
        reverse("animal:endpoint_copy_selector", args=(1,)),
        reverse("animal:dosing_regime_update", args=(1,)),
        reverse("animal:endpoint_list", args=(1,)),
        reverse("animal:endpoint_new", args=(1,)),
        reverse("animal:endpoint_detail", args=(1,)),
        reverse("animal:endpoint_update", args=(1,)),
        reverse("animal:endpoint_delete", args=(1,)),
    ]
    for url in urls:
        check_200(client, url)
```

It's ok to skip views if they give us trouble; the goal is to just maximize code that's at least touched for regressions. This will boost our test coverage by a large percentage, with very little work (the test above took ~30min to write and improves coverage from 48% to 67% for animal/views.py, and from 59% to 85% for animal/forms.py)

I'd review the coverage reports, and essentially write the same for each of the following apps (in order of priority):

- study
- myuser
- summary
- lit
- riskofbias
- epi
- assessment
- vocab
- epimeta
- invitro
- bmd

If any particular view is giving you trouble with such a simple test, that's ok, I'd just skip it. We don't need 100% coverage, just trying to get the maximum increase in a time efficient manner.

With this PR, I wouldn't try to check POST or other verbs; we can do that in another PR. If you have time remaining, I'd explore adding a `test_smoke_post` method and writing tests cases for some views using it, just with a simple payload for a valid form.

If you need help finding the correct IDs so that the routes are not 404, I would load the hawc test fixture database, login as the pm user, and browse around to see what pages you have access to.

shapiromatron:
@BerkowitzMaxICF - this is looking great! working on anything else or is this one ready for review?

shapiromatron (approved):
Looks great, really nice job!! This will go a long way in just ensuring confidence that our application doesn't have any regressions as we continue to make updates.  Thank you!

PR #906, opened 2023-09-27
Add UDF to study and endpoint forms

munnsmunns:
- Add model for saved UDF data, `ModelUDFContent`
- Add UDF functionality to the Animal Endpoint form and Study form
- UDF data is saved/initialized properly
- UDF data is displayed on the detail pages. Adds a Mixin which adds UDF data to the view context


hawc/apps/udf/views.py
```diff
@@ -200,3 +201,17 @@ class DeleteTagBindingView(BaseDelete):
 
     def get_success_url(self):
         return self.assessment.get_udf_list_url()
+
+
+class UDFDetailMixin:
```
munnsmunns, line 206: I tried to add this mixin (instead of the changing the view's get_context_data directly) but it didn't seem to work. Any ideas?

shapiromatron: That should work fine, as long as the other classes call `super()`  Did you put the mixin on the far-left of the class inheritance?

shapiromatron:
I think this needs a little more work - can this be updated and completed? I made a few edits in a scratch branch to get the conditional form logic working for the demo, but it may need to be added elsewhere too:

* f105ecd9 wip (3 days ago) {Andy Shapiro}

hawc/apps/study/templates/study/study_detail.html
```diff
@@ -133,6 +133,10 @@ <h3>{{assessment.get_rob_name_display}}</h3>
       {% include "eco/fragments/_design_list.html" with object_list=object.eco_designs.all %}
     {% endif %}
 
+    {% if udf_content %}
```
shapiromatron, line 136: move the if statement inside the fragment template?

hawc/apps/udf/templates/udf/fragments/_udf_content.html
```diff
@@ -0,0 +1,21 @@
+<h3>User defined fields</h3>
+<table class="table table-sm m-0 noborder">
```
shapiromatron, line 2: use our bs4_thead and bs4_colgroup items

shapiromatron (commented):
Let's try to get this one wrapped up! We can start here and then do the UDF tags next.

shapiromatron:
I haven't had time to do the work, but I did a code review of this PR and it's looking great so far. I think one thing we need to add is some caching mechanism to cache the UDF for each assessment, so we don't have as many database hits to see if each model type has a related UDF or not. 

It might be as simple as following a pattern similar to `assessment/permissions.py` to make a new class with some of this information. It could perhaps have a cache key like `assessment-123-udf`, and it has a property for content_ids = [1,2,3] and tag_ids = [3,4,5] that lists the content types and tags that have related UDF. We could add a signal to clear the cache whenever ModelBinding or TagBinding changes for a given assessment.

shapiromatron (approved):
I made a number of changes; see below. In general, try to generalize methods where possible; it's good that we had a Mixin for adding UDF on the detail view, but we needed the same thing to add to any Model Form; the way it was implemented as proposed required us to make lots of repetitive inline changes to each model form.

I also figured out how to auto-add the JS to the form pages; this was worth a read - [Media](https://docs.djangoproject.com/en/5.0/topics/forms/media/).

* 10cb86341 add tests
* 0a85e9dbd refactor content list
* 1f369636f refactor cache
* 518b4f581 bonus - improve type annotations for cacheable
* 780ef15c7 fix signals
* 6cc928db3 fix unique-together constraint
* 2fe1f621d fix table template layout
* b667ecba1 rename UDF table fragment
* 5534d617e add db constraint
* e445f724a load JS with DynamicForm or DynamicFormWidget
* 9a2f73056 refactor common code into a form Mixin
* c242a7dad reformat test html

PR #907, opened 2023-09-28
Fix NOEL/LOEL form and display

shapiromatron:
Fix a bug where users couldn't set or view NOEL/LOEL for endpoints where data was not extracted.  

This bug would only occur if a user created an endpoint, but didn't extract any dose group data. Then, when they tried to update the endpoint, it didn't have dose units available so the LOEL select input didn't have any options.

It also failed on the read-only view of view.


munnsmunns (approved):
Took a look at it on my machine, looks good to merge!

PR #908, opened 2023-09-28
merge tags

casey1173:
Adds a "Apply all tags below" option to the conflict resolution view. This merges all UserReferenceTags for a reference then applies the tags to the reference and the UserReferenceTag of the user making this change.

Unmerged
image

After merge
image

Updated "Approve All" button:

image


shapiromatron (approved):
LGTM - I made some edits to the flow and the button, so it says "Approve all" to be consistent with the current "Approve" buttons. I also added some help text to describe what happens when you press.

image
image

If you're happy with the flow and changes, ok to merge!

PR #909, opened 2023-09-28
added dich % change calc

dannypeterson:


hawc/apps/animal/models.py
```diff
@@ -1235,6 +1235,13 @@ def percentControl(data_type, egs):
                 low = eg["lower_ci"]
                 high = eg["upper_ci"]
 
+            elif data_type == constants.DataType.DICHOTOMOUS:
+                if i == 0:
+                    control_incidence = eg["incidence"]
+                    n_1 = eg["n"]
+                if control_incidence is not None and control_incidence > 0:
+                    incidence = eg["incidence"]
+                    mean = ((incidence - control_incidence) / n_1) * 100
```
shapiromatron, line 1244: i think this need to be ((i1/n1)-(i0/n0))/(i0/n0), where i is incidence and n is num animals

shapiromatron (changes requested):
one minor change to how change is calculated; we have to normalize to fraction incidence instead of raw unnormalized counts

dannypeterson:
percent change calc
Updated the calc to ((i1/n1)-(i0/n0))/(i0/n0), but the calculations seem weird... (0.7 - 0.1) / (0.1) = 0.6/0.1 = 6 so not sure why we are getting a 5.99 float. Also shouldnt the result be 0.6 instead bc we multiply by 100 for % change

shapiromatron:
> percent change calc Updated the calc to ((i1/n1)-(i0/n0))/(i0/n0), but the calculations seem weird... (0.7 - 0.1) / (0.1) = 0.6/0.1 = 6 so not sure why we are getting a 5.99 float. Also shouldnt the result be 0.6 instead bc we multiply by 100 for % change

That seems correct! I think it's a 600% percent chagne relative to control

hawc/apps/animal/models.py
```diff
@@ -1242,7 +1242,7 @@ def percentControl(data_type, egs):
                 if i_0 is not None and i_0 > 0:
                     i_1 = eg["incidence"]
                     n_1 = eg["n"]
-                    mean = ((i_1 / n_1) - (i_0 / n_0)) / (i_0 / n_0)
+                    mean = ((i_1 / n_1) - (i_0 / n_0)) / (i_0 / n_0) * 100
```
shapiromatron, line 1245: looks correct! you can see we do the same on line 1219 `  mean = (mu_2 - mu_1) / mu_1 * 100.0`

shapiromatron (approved):

PR #911, opened 2023-10-03
Epi export rewrite

rabstejnek:
Convert tabular exports from DRF serializer based conversions to django ORM-based conversions for the `epi` application.


shapiromatron (approved):
Code looks great, I tested it on a number of data pivots and they load ~5-20x faster than before, this will be an awesome feature available when it's ready for prime time. 

I made a few minor changes:

* b62e853b6 minor formatting (2 minutes ago) {Andy Shapiro}
* 6c9abd326 use vectorized timestamp conversion (20 minutes ago) {Andy Shapiro}

I didn't check the custom methods for data pivot columns in depth; let me know if there's anything in particular you'd like a deep review on, otherwise, LGTM!

PR #914, opened 2023-10-04
fake imports

caseyhans:
Adds a feature flag to fake HERO, PubMed, and DTXSID imports by returning fake data. Used for security scans so that when the scans inject bad data, we don't proliferate requests to 3rd party systems.

To set, use use the boolean flag: `settings.HAWC_FEATURES.FAKE_IMPORTS`; defaults to False.


shapiromatron (approved):
LGTM - I added some test for fake mode; ok to merge if you're happy @caseyhans !

PR #915, opened 2023-10-04
dichotomous percent control - check for division by zero

shapiromatron:
Write a suite of tests to ensure that we're not dividing by zero in some calculation (observed with some strangely entered data extractions).


dannypeterson (approved):
lgtm - thanks for fixing the calcs and writing the test

PR #916, opened 2023-10-05
toggle count display in heatmaps

shapiromatron:
In some cases users may not want to display count information in a heatmap; they may want a more qualitative view or it may be hard to describe that the count doesn't refer to # of references but instead # of extracted endpoints, etc. 

This PR adds a new option to users `Show Value Counts`. There are three states:

- 1: show count & show color (default)
- 2: hide count & show color
- 3: hide count & hide color

image

The counts option is available when configuring custom exploratory heatmaps, and is also an option on the "customize" tab of the automatically generated heatmaps.

### Option screenshot (see "Show value counts") :
image

A third option showing no colors:
image


rabstejnek (approved):

shapiromatron:
@rabstejnek re-requesting review - we changed it so that there are three possible states (see updated description)

PR #917, opened 2023-10-05
created new analytics pages

dannypeterson:
Created a single analytics page and 3 partial templates in HAWC to display assessment analytics:

- Analytics (`/mgmt/assessment/:id/analytics/`)
    - Time series
    - Time spent
    - Growth

These will continue to be built out to provide HAWC users with more "meta" information for what's occurring in a given assessment. Its purpose is to keep people motivated, report to management to understand what's currently going on, and show the scale of assessments, where are are questions about errors in extraction and relative magnitude.

Added behind a feature flag `HAWC_FEATURES.ENABLE_ANALYTICS = True` - defaults to False.

assessment analytics


shapiromatron (approved):
I took this a little further than the jira card and made it so that we have one django view that can potentially return different results via htmx if an action is specified. In doing so, I built a few generic base classes to help with this. 

Please give the functionality a run through to make sure it looks ok, and if you're happy with the changes, ok to merge!

PR #888, opened 2023-08-29
Icf create epi v2 client

DookTibs:
Add client operations to CRUD epiv2, and required tests.

PR #918, opened 2023-10-06
epi design api - allow filtering by study id

shapiromatron:
Enable filtering by study identifier on the epiv2 design API.

Example:

- `/epidemiology/api/design/?assessment_id=123`
- `/epidemiology/api/design/?assessment_id=123&study=123456`

This was extracted from #888 in preparation for the quarterly release.


caseyhans (approved):
Works as intended and LGTM!

PR #919, opened 2023-10-06
Add tag deletion/addition filters for conflict resolution page

caseyhans:
Adds additional filtering to conflict resolution page.
- Candidate Tag Additions: filters by references with users seeking to add a specific tag (tag is present for an unresolved UserReferenceTag, but not on reference.tags)
- Candidate Tag Deletions: filters by references with users seeking to remove a specific tag (tag is present on reference.tags, but not in an unresolved UserReferenceTag)

image

- Show tag deletions on Tag Status page
- Show text for empty UserReferenceTag case
image

### DIFF
--- a/hawc/apps/lit/filterset.py
+++ b/hawc/apps/lit/filterset.py
@@ -74,6 +74,32 @@ class ReferenceFilterSet(BaseFilterSet):
         label="Tagged by me",
         help_text="All references that you have tagged",
     )
+    addition_tags = df.ModelMultipleChoiceFilter(
+        queryset=models.ReferenceFilterTag.objects.all(),
+        method="filter_tag_additions",
+        conjoined=True,
+        label="Canditate Tag Additions",
+        help_text="Select a tag to view references with that tag as a candidate tag (but not as a consensus tag). If multiple tags are selected, a reference user tag must include all selected tags.",
+    )
+    include_additiontag_descendants = df.BooleanFilter(
+        method=filter_noop,
+        widget=CheckboxInput(),
+        label="Include tag descendants",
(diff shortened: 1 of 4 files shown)


hawc/apps/lit/migrations/0022_userreferencetag_deleted_tags.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 4.2.6 on 2023-10-30 02:58
+
+import django.contrib.postgres.fields
+from django.contrib.postgres.aggregates import ArrayAgg
+from django.db import migrations, models
+
+
+def add_deletions(apps, schema_editor):
```
shapiromatron, line 8: this was a joy to read; nice job.

hawc/apps/lit/filterset.py
```diff
@@ -188,6 +214,52 @@ def filter_anything_tagged_me(self, queryset, name, value):
         ).filter(my_tag_count__gt=0)
         return queryset.distinct()
 
+    def filter_tag_additions(self, queryset, name, value):
```
shapiromatron, line 217: These methods look good, but they could be slow. I don't think we need to change anything now.

We may need to monitor performance with some real data after its in the wild for a while to determine if we need to optimize.  I'm thinking specifically of a case where we include descendants, and the parent has many children. I wonder if there's a way to see to see if a reference has any of these tags in a single query.

I think you could do it it with the ArrayField using [overlap](https://docs.djangoproject.com/en/4.2/ref/contrib/postgres/fields/#overlap), but that's only for the deletions, not the additions.

shapiromatron (approved):
Really nice job. This was a fun read.

Made some minor changes; if you're ok, feel free to merge!

* ef5786f79 reuse tag queryset for each field instead of multiple queries
* 0cda1b096 use == instead of is for numerical comparison 
* a07c65fa4 use alert template tag; add icon template tag

PR #920, opened 2023-10-10
moved bs4 and hawc template tags to builtins

dannypeterson:
Django templates has a `builtin` option that preloads libraries into every template [re: [Django: Move a template tag library into builtins](https://adamj.eu/tech/2023/09/15/django-move-template-tag-library-builtins/)]. This allows us to no longer need `{% load  templatetag %}` in every template, so we have more DRY code. Also consolidated custom template tags into `hawc` and `bs4`

Two HAWC specific template tag libraries have been registered so they're always available:

- `hawc`
- `bs4`

As well as one external template tag library:

- `crispy_forms_tags`


tests/hawc/apps/common/templatetags/test_hastext.py
```diff
@@ -1,4 +1,4 @@
-from hawc.apps.common.templatetags.hastext import hastext
+from hawc.apps.common.templatetags.hawc import hastext
```
shapiromatron, line 1: rename this file to test_hawc?

hawc/apps/study/templates/study/attachment_form.html
```diff
@@ -1,6 +1,5 @@
 {% extends 'assessment-rooted.html' %}
 
-{% load add_class %}
 {% load crispy_forms_tags %}
```
shapiromatron, line 3: I think you suggested making `crispy_forms_tags` a builtin too - what do you think?

hawc/apps/riskofbias/templates/riskofbias/rob_edit.html
```diff
@@ -1,9 +1,7 @@
 {% extends 'assessment-rooted.html' %}
 
-{% load add_class %}
 {% load get_at_index %}
```
shapiromatron, line 3: I dont think this is used here, or anywhere else? delete `get_at_index` from this template and the `study/templatetags`?

shapiromatron (commented):
Looking good. A few more comments, and a question to you if you like this pattern and if think we should apply to the crispy tag too.

hawc/apps/common/templates/common/dynamic_form.html
```diff
@@ -1,2 +1,2 @@
-
+{% load crispy_forms_tags %}
```
dannypeterson, line 1: need to figure out which `dynamic_form.html` to remove? In hawc there is
`templates/common/dynamic_form.html` and `templates/common/widgets/dynamic_form.html` which are the same

shapiromatron: done! sorry I had a comment I sent to you but realized too late it was "pending" and never sent. I deleted the widgets one.

* fe71fbee8 remove duplicative widget and template

shapiromatron (approved):
Looks great! Thanks for doing this and being so methodical about it; it looks really clean. I made a few very minor changes, ok to merge if you're happy with my tweaks:

* b9a29cd48 modify whitespace; remove unused template
* fe71fbee8 remove duplicative widget and template

PR #921, opened 2023-10-11
rewrite riskofbias exports

shapiromatron:
Rewrite DRF serializer based riskofbias exports with exports using our new QuerySet based approach. Tested on one project, we reduced the number of queries from 750+ to ~5. Looking at just db time on the debug toolbar, from 600ms to around 50ms; this doesn't include the python runtime decrease with converting to Django models and corresponding serializer objects. 

I renamed a few columns in the export just as the prior names seemed incorrect, and since this export isn't used in any downstream views in HAWC, renaming should be harmless.


hawc/apps/riskofbias/exports.py
```diff
@@ -1,92 +1,104 @@
...
+        }
+
+    def prepare_df(self, df):
+        return self.format_time(df)
+
+
+class DomainExport(ModelExport):
+    def get_value_map(self):
+        return {
+            "domain_id": "id",
```
rabstejnek, line 38: Could you move these prefixes from domain and metric into the exporter? Just for consistency

rabstejnek (approved):
Looks good, just one optional comment to address.

hawc/apps/riskofbias/exports.py
shapiromatron: done!

PR #923, opened 2023-10-11
added bootstrap modal to bs4 template

dannypeterson:
Created a common modal template `templates/common/modal_template.html` and converted template based modals for more reusable code:

- `licence.html`
- `_smartTagEditModal.html`


hawc/apps/common/templatetags/bs4.py
```diff
@@ -102,3 +102,35 @@ def plotly(fig: Figure | None, **kw) -> str:
     <script>document.addEventListener("{event}", {func}, false);</script>"""
         )
     )
+
+
+@register.simple_tag
+def render_modal(modal_id, modal_title=""):
```
shapiromatron, line 108: This doesn't look right to me - I think in the jira ticket there were two examples, one for a tag and one for the template used with the tag. I think you can copy/paste the code into hawc, and then just do a quick search for any html templates that have `modal-dialog` and see if you can replace them. I can hop on a call to chat tomorrow if it'd be helpful.

shapiromatron: my comment wasn't correct either - I think we figured it out together this morning. This should be a template, not a template tag

shapiromatron (approved):
nice job!

PR #925, opened 2023-10-13
remove hawc.tools.docx_utils

shapiromatron:
Source and tests removed; no longer used anywhere in application.


dannypeterson (approved):
seems like a straight forward removal - lgtm

PR #928, opened 2023-10-17
update python package versions

shapiromatron:
Update all python package versions except for crispy-forms, which was done in a subsequent PR; see #929. 

- fixed linter updates and rules w/ updated black and ruff
- update plotly.js in npm to sync w/ plotly python version
- fix deprecations and warnings re: unordered querysets
- assorted pandas updates to go from 1.x to 2.x series

BONUS - remove coverage artifact; it can add up to 20s of runtime to github actions and it's rarely used


pyproject.toml
```diff
@@ -68,7 +68,7 @@ exclude = [
 line-length = 100
 target-version = "py311"
 select = ["F", "E", "W", "I", "UP", "S", "B", "T20", "RUF"]
-ignore = ["E501", "B904", "B007", "S308", "S113", "S314", "S603", "S607", "RUF012"]
+ignore = ["E501", "B904", "B007", "S308", "S113", "S314", "S603", "S607", "RUF012", "RUF015"]
```
rabstejnek, line 71: What's your reason for ignoring this rule? Because its slightly less readable to do `next(iter(my_list))`?
https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/

rabstejnek (approved):
Looks good! I am curious about your rule changes, but nothing that blocks this PR from being merged.

pyproject.toml
shapiromatron: Yes, exactly! If I was a normal python developer I'd be very confused by `next(iter(...))` - I'd be fine we made some helper method get_first(...) or something, but I figured in the cases where we slice the whole list, we're not doing anything too computationally expensive.

PR #930, opened 2023-10-18
fix assessment list to make distinct by assessment

shapiromatron:
Fix bug where `/assessment/all/` duplicated assessments based on the number of people assigned to various roles. The previous implementation made duplicative assessments since it was generating a row for each assessment-user combination; to dedup using the distinct() disabled our ability to have a custom sort ordering. 

The new approach breaks adds an `Exists` annotation to the assessment so we prevent the duplicates; query performance is actually faster despite the extra annotations.


dannypeterson (approved):
neat use of djangos built in classes and interesting fix! approved

PR #933, opened 2023-10-19
extraction ready reference layout

shapiromatron:
Convert reference layout on `/lit/assessment/:id/references/extraction-ready/` to use our richer reference list component. Updated the component to selectively show user tags, hide abstract, and use extra tag classes when provided.

Before:
image

After:
image


caseyhans (approved):
Mostly just made style changes as I saw fit. Feel free to change anything back!
LGTM, I like the added flexibility of the reference template fragment.

hawc/apps/lit/models.py
```diff
@@ -1163,14 +1163,15 @@ def extract_dois(cls, qs, logger=None, full_text: bool = False):
             )
 
     @classmethod
-    def annotate_tag_parents(cls, references: list, tags: models.QuerySet):
+    def annotate_tag_parents(cls, references: list, tags: models.QuerySet, user_tags: bool = True):
```
shapiromatron, line 1166: added this option too; reduces number of queries fetching these tags when we dont need them.

PR #935, opened 2023-10-20
api standardization

shapiromatron:
We've developed some good patterns for API permissions and standardization of the past few years, and they've been applied on new API routes but not retrospectively. This PR attempts to standardize API routes throughout HAWC:

* b94afceec more removals of DRF generic viewsets
* 825cdbb79 use BaseAssessmentViewSet where possible
* aea69fbf6 testing - DRY; switch to check_api_json_data


tests/hawc/apps/lit/test_api.py
```diff
@@ -559,7 +549,7 @@ def test_permissions(self, db_keys):
 
     def test_bad_requests(self, db_keys):
         # test bad id
-        url = reverse("lit:api:reference-detail", args=(-1,))
+        url = reverse("lit:api:reference-detail", args=(9999999,))
```
rabstejnek, line 552: This doesn't need to be addressed, but I think `-1` works better here personally, since its an impossible id vs a highly unlikely id.

rabstejnek: Also looking through our tests it looks like we use different variations of `9`s but with different number of digits. It may just be a pet peeve and not an actual code maintainability issue, but if we actually want a big number maybe `123456789` would be better just so that we can maintain consistency with these bad ids. Or make a variable in our `conftest.py`. Just a thought.

rabstejnek (approved):
Looks good! I had one comment that doesn't need to be addressed in this PR, but just a thought about the "bad ids" we use in tests. If you think its worth addressing maybe we can make a card for it?

tests/hawc/apps/lit/test_api.py
shapiromatron: I think I remember changing this because if I remember correctly -1 didn't pass the URL validation anymore since the negative symbol doesn't match `\d+`. I like your idea though; maybe we can use 8675309 :)

shapiromatron: I found it - it's because our regex that we use doesn't allow negative numbers. I changed to 987654321; https://github.com/shapiromatron/hawc/blob/775313d858b75071f65a165a21fdc3c687211d1a/hawc/apps/assessment/api/viewsets.py#L152

PR #936, opened 2023-10-20
fix bug - RiskOfBiasScore.bias_creation via API

shapiromatron:
Fix bug where POST to `/rob/api/reviews/` to create a new score set didn't set the user-defined bias_creation field. Add test to demonstrate success.


hawc/apps/riskofbias/serializers.py
```diff
@@ -321,33 +321,15 @@ def validate(self, data):
...
-            try:
-                score = models.RiskOfBiasScore.objects.create(
-                    riskofbias=rob,
-                    metric=score_to_create["metric"],
-                    is_default=score_to_create["is_default"],
-                    label=score_to_create["label"],
-                    score=score_to_create["score"],
-                    notes=score_to_create["notes"],
-                )
-            except ValidationError as err:
```
rabstejnek, line 342: Is it fine to not include this try/except block? ie will a Django ValidationError automatically be converted to the serializer version?

rabstejnek (approved):
Looks good! I had one question about excluding the try/except block. It looks like custom validation is done on `RiskOfBiasScore`'s `save` method, so we could expect it to actually throw a Django `ValidationError`.

hawc/apps/riskofbias/serializers.py
shapiromatron: I added it back just to be safe. We didn't have a test either so it's probably safer to keep it, though I don't think the `create` method runs the validators.

PR #938, opened 2023-10-20
Reparse DSSTox content and add search index

munnsmunns:
Add a new GeneratedField `search` field to DSSTox model and index it to improve query performance for chemical information. Query the search field instead of individual attributes from model and the JSON field.

Running `python manage.py chemicals --refresh` will refresh the DSStox content using the new API.

References:

* https://www.paulox.net/2023/11/24/database-generated-columns-part-2-django-and-postgresql/
* https://docs.djangoproject.com/en/5.1/ref/contrib/postgres/search/#searchvector


munnsmunns:
@shapiromatron The only thing I haven't done is hook up the search index to the filterset, everything else is ready for review

hawc/apps/assessment/constants.py
```diff
@@ -124,3 +125,6 @@ class UncertaintyChoices(models.IntegerChoices):
     THIRTY_THOUSAND = 30_000, "30,000"
     ONE_HUNDRED_THOUSAND = 100_000, "100,000"
     THREE_HUNDRED_THOUSAND = 300_000, "300,000"
+
+
+DSSTOX_SEARCH_VECTOR = SearchVector("dtxsid", "content", config="english")
```
munnsmunns, line 130: we could overwrite the queryset in the Autocomplete search to use this, maybe we should add a card for that

shapiromatron (approved):
Sorry it took me so long to go through it, I wanted to use the new search index instead of writing it and then not using it. Please see the changes I made and let me know if you're ok with the edits. 

The new GeneratedField in django is really powerful...

PR #940, opened 2023-10-26
Replacing old cache patterns with new cacheable function

dannypeterson:
Replaced some legacy cache code with a new `cacheable()` function


dannypeterson:
@shapiromatron wanted to check if im on the right track before continuing with the rest of the pattern, thanks!

hawc/apps/vocab/views.py
```diff
@@ -12,16 +11,12 @@ class EhvBrowse(TemplateView):
...
-        data = cache.get(key)
-        if data is None:
-            data = WebappConfig(
-                app="animalStartup",
-                page="ehvBrowserStartup",
-                data={"data": models.Term.ehv_dataframe().to_csv(index=False)},
-            ).json()
-            cache.set(key, data, settings.CACHE_10_MIN)
-        return data
+        data = WebappConfig(
```
shapiromatron, line 14: all the others look good, but this will be evaluated every time regardless of the cache status; you could either put this in a lambda or put in a function

shapiromatron (commented):
looking good! one comment for one of the replacements

hawc/apps/animal/api.py
```diff
@@ -93,10 +91,12 @@ def study_heatmap(self, request, pk):
         if unpublished and not self.assessment.user_is_reviewer_or_higher(self.request.user):
             raise PermissionDenied("You must be part of the team to view unpublished data")
         key = f"assessment-{self.assessment.id}-bioassay-study-heatmap-pub-{unpublished}"
-        df = cache.get(key)
-        if df is None:
-            df = models.Endpoint.heatmap_study_df(self.assessment, published_only=not unpublished)
-            cache.set(key, df, settings.CACHE_1_HR)
+        df = cacheable(
+            lambda: models.Endpoint.heatmap_study_df(
```
shapiromatron, line 95: this is super minor, but the indentation is throwing me off. Can we do something like instead?

```python
# one liner
func = lambda: models.Endpoint.heatmap_study_df(self.assessment, published_only=not unpublished)
# or this

def func():
    return models.Endpoint.heatmap_study_df(self.assessment, published_only=not unpublished)
```

dannypeterson:
Changed most but not all occurrences with `cache.set()`, figured this might be a good place to stop; changing the remaining occurrences would have diminishing returns on time imo

hawc/apps/assessment/views.py
```diff
@@ -91,122 +90,118 @@ def dispatch(self, request, *args, **kwargs):
...
-                iv_endpoints_with_data_percent=percentage(iv_endpoints_with_data, iv_endpoints),
-                visuals=visuals,
-                assessments_with_visuals=assessments_with_visuals,
-                assessments_with_visuals_percent=percentage(assessments_with_visuals, assessments),
-            )
-            cache_duration = 60 * 60 * 24  # one day
-            cache.set(key, counts, cache_duration)  # cache for one day
-            logger.info("Setting about-page cache")
-        return counts
+        counts = dict(
```
shapiromatron: I think this is currently always evaluated; we need to put this somewhere else so a cached version could be used. I'll make this edit real quick and merge this in.

shapiromatron (approved):
I made on minor change to get the about page to properly cache - if you're ok w/ my changes, ok to merge!

* 4fe959a9 move calculations outside of the view and cache

shapiromatron:
@dannypeterson - ok to merge?

PR #941, opened 2023-10-27
add new visual - user uploaded image

shapiromatron:
Add new visual type - user uploaded images. Valid import formats can be jpeg, png, and jpg.


rabstejnek (approved):
Looks good! I did add one commit to fix some issues I saw:

- The `image` field needed to be removed from other visual forms.
- The `ClearableFileInput` widget was being used for the `image` field, which we don't want since this is a required field for the image visual form; changed this to `FileInput`.
- Reworded some text so that the periods in file formats didn't look bad

If you are fine with these additional changes feel free to merge, otherwise feel free to revert any of these changes that you don't feel were necessary.

shapiromatron:
Thanks @rabstejnek. Finally got a chance to look of your edits. Great find with the `ClearableFileInput ` -> `FileInput` field; that definitely would have been a bug report in the future.

PR #942, opened 2023-10-28
replace black with ruff format

shapiromatron:
Remove black and replace with ruff, which recently released a formatter. This removes one dependency and should increase performance. 

Two commits (review separately):

* 23548d209 replace black w/ ruff format; update configuration and docs:
* 4be9b8b38 apply ruff formatter

This will require minor updates to VSCode configuration:

```json
"[python]": {
    "editor.defaultFormatter": "charliermarsh.ruff",
    "editor.formatOnSave": true,
    "editor.codeActionsOnSave": {
        "source.fixAll": true
    }
}
```


munnsmunns (approved):
working great in my environment! LGTM

PR #944, opened 2023-10-30
added study data summaries to downloads page

dannypeterson:
### Study level export data available in downloads tab

Added study summary/design data to available exports for animal, epidemiology, and ecological endpoints. These exports only contain published data.
image
image
image


shapiromatron (approved):
LGTM!

PR #946, opened 2023-11-02
visualizations list - show rob heatmap vs study eval heatmap labels

dannypeterson:
Replace 'risk of bias' label with 'study evaluation' in Visualizations tab table and inline filter form if Assessment level field specifies 'Study Evaluation' as opposed to 'Risk of bias'.
Checked [ORD IRIS Assessment PFHxA (2023)](http://127.0.0.1:8000/assessment/100500070/) and saw no extra db queries thanks to `select_related()`.

study eval


shapiromatron (approved):
LGTM!

PR #947, opened 2023-11-02
initial animalv2 model declarations

DookTibs:
First pass at converting Vidhi's wireframes for the animalv2 UI into django models. Bunch of TODO's and placeholder stuff in there still.

Included for visual reference - graph_models output for animalv1 vs. animalv2 models.

### animalv1:
animal

### animalv2:
animalv2


hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+        return self.treatment.get_assessment()
+
+    def get_study(self):
+        return self.treatment.get_study()
+
+    # TODO - flat_complete_header_row / flat_complete_data_row
+
+
+# can't just name it Endpoint; that clashes around baseendpoint_ptr
+class AnimalEndpoint(BaseEndpoint):
```
shapiromatron, line 316: do not integrate with `BaseEndpoint`; we dropped this relation with epiv2; just inherit from models.Model instead. Rename to `Endpoint`, or do we get too many namespace conflicts?

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+    # specifically, DoseGroup.dose?
+    dose = models.ForeignKey(
+        DoseGroup, on_delete=models.CASCADE, related_name="v2_data_extractions"
+    )
+    observation_timepoint = models.ForeignKey(
+        AnimalEndpointTimepoint, on_delete=models.CASCADE, related_name="v2_data_extractions"
+    )
+
+    # specific fields
+    n = models.PositiveSmallIntegerField(blank=True, null=True, validators=[MinValueValidator(0)])
```
shapiromatron, line 453: i think many of these fields may only belong on the `DoseResponseGroupLevelData` - dose, n, variance, treatment_related_effect

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+    objects = managers.DoseResponseGroupLevelDataManager()
+
+    data_extraction = models.ForeignKey(
+        DataExtraction, on_delete=models.CASCADE, related_name="v2_group_level_data"
+    )
+    treatment = models.ForeignKey(
+        Treatment, on_delete=models.CASCADE, related_name="v2_group_level_data"
+    )
+    # specifically, DoseGroup.dose?
+    dose = models.ForeignKey(
```
shapiromatron, line 526: instead of a direct relation - maybe use the same `dose_index` so we can relate a specific response here to doses of different representations

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+    response = models.FloatField()
+    variance = models.FloatField(blank=True, null=True, validators=[MinValueValidator(0)])
+    treatment_related_effect = models.PositiveSmallIntegerField(
+        choices=constants.TreatmentRelatedEffect.choices
+    )
+    statistically_significant = models.PositiveSmallIntegerField(
+        choices=constants.StatisticallySignificant.choices
+    )
+    p_value = models.CharField(max_length=128, blank=True, help_text="TODO")
+    NOAEL = models.SmallIntegerField(default=-999, help_text="No observed effect level")
```
shapiromatron, line 539: remove noel/loel from this table - rename in all cases to noel/loel as the "a" is assessment specific.

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+    def get_assessment(self):
+        return self.data_extraction.get_assessment()
+
+    def get_study(self):
+        return self.data_extraction.get_study()
+
+    # TODO - flat_complete_header_row / flat_complete_data_row
+
+
+# TODO - reversion.register?
```
shapiromatron, line 576: yes - we should register all models to reversion

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+    created = models.DateTimeField(auto_now_add=True)
+    last_updated = models.DateTimeField(auto_now=True)
+
+    def get_assessment(self):
+        return self.data_extraction.get_assessment()
+
+    def get_study(self):
+        return self.data_extraction.get_study()
+
+    # TODO - flat_complete_header_row / flat_complete_data_row
```
shapiromatron, line 573: we can remove these TODO from all models; we're no longer implementing using these methods

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+Make an svg/png out of it (assuming you've installed graphviz):
+
+    dot -Tsvg /path/to/previously_generated_file.dot > converted.svg
+"""
+
+
+class Experiment(models.Model):
+    objects = managers.ExperimentManager()
+
+    # TODO - what does this do?!?!
```
shapiromatron, line 24: it allows those fields to be used in bulk_cleanup methods - ok to remove for now unless it throws an exception somewhere if it's not defined

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+    TEXT_CLEANUP_FIELDS = ()
+
+    study = models.ForeignKey(
+        "study.Study", on_delete=models.CASCADE, related_name="v2_experiments"
+    )
+    name = models.CharField(
+        max_length=80,
+        verbose_name="Experiment name",
+        help_text="Short-text used to describe the experiment "
```
shapiromatron, line 33: annoying request; but make all long help texts a single line triple-quoted string; it's easier to manage a single long string that our editors can format instead of manually formatting breaks; this is consistent with epiv2 - https://github.com/shapiromatron/hawc/blob/main/hawc/apps/epiv2/models.py

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+class DoseGroup(models.Model):
+    objects = managers.DoseGroupManager()
+
+    # TODO
+    TEXT_CLEANUP_FIELDS = ()
+
+    treatment = models.ForeignKey(
+        Treatment, on_delete=models.CASCADE, related_name="v2_dose_groups"
+    )
+    dose = models.FloatField(validators=[MinValueValidator(0)])
```
shapiromatron, line 299: i think we'll need some sort of dose_index here like our old animal model so it's easy to compare that these are the same groups, regardless of the specified dose units. We could also flip to something like this:


```python
class DoseGroup:
    treatment = ForeignKey
    dose_index = PositiveSmallIntegerField()

class DoseRepresentation:
    group = ForeignKey(DoseGroup)
    dose = FloatField()
    units = ForeignKey(DoseUnits)
```

But it may be better to keep it a single model instead of breaking it up...

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+    )
+    # these next 4 are so-called flexible dropdowns. e.g. anatomical is
+    # "left/right/anterior/Posterior/Cranial/Caudal/Lateral/medial" but we want to let
+    # ppl type something else. do we want to store these (and add custom ones) in a table?
+    # Or just build the UI to show options from a list but also let user type custom ones
+    # (in that case just store it as a string)? leave as string for now.
+    effect_modifier_timing = models.CharField(max_length=128, blank=True)
```
shapiromatron, line 378: I'm not sure what to do with these yet... @swatford fyi - I think we need to get more information on how these are supposed to be used... but thanks for taking a first pass...

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
+from django.core.validators import MinValueValidator
+from django.db import models
+
+from ..assessment.models import BaseEndpoint, DSSTox
+from ..vocab.models import Term
+from . import constants, managers
+
+"""
+Generate a GraphViz dot file of this model:
```
shapiromatron, line 9: Move this content to the appropriate location in `docs/docx/database.md`; we can all the static image later after things have stabilized

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+        on_delete=models.CASCADE,
+        related_name="v2_animal_groups",
+        help_text="When adding a new strain, put the stock in parenthesis, e.g., "
+        + '"Sprague-Dawley (Harlan)."',
+    )
+    sex = models.CharField(max_length=1, choices=constants.Sex.choices)
+    animal_source = models.CharField(
+        max_length=128, help_text="Source from where animals were acquired", blank=True
+    )
+    lifestage_at_exposure = models.CharField(
```
shapiromatron, line 172: we have a `animal.constants.Lifestage` enum for this and lifestage_at_assessment - is it ok to use the enum, or was there a reason to use free text?

hawc/apps/animalv2/models.py
```diff
@@ -0,0 +1,576 @@
...
+        return self.name
+
+    def get_assessment(self):
+        return self.experiment.get_assessment()
+
+    def get_study(self):
+        return self.experiment.get_study()
+
+
+class AnimalEndpointTimepoint(models.Model):
```
shapiromatron, line 394: rename model to `ObservationTime`?

shapiromatron (commented):
This is looking great so far! Thanks for taking the first pass. Like you, I'm struggling on how to deal w/ the modifier terms; ok to keep as is for now and we'll internally sync up to provide additional/better guidance.  I made a number of requests, hopefully they're relatively simple to implement. Feel free to start a convo on anything for a deeper discussion.

hawc/apps/animalv2/models.py
DookTibs: I'll circle back with Vidhi and Pam. That was my thought too, that it seemed very odd to have these in both locations, latest guidance I got from them was that it very explicitly should be that way. But like I said, seems weird to me and sounds like it seems weird to you. We'll discuss.

hawc/apps/animalv2/models.py
shapiromatron: There's an assessment-level setting that changes how those values are named, https://github.com/shapiromatron/hawc/blob/213b0a2121bcbbc34ec098bdddcc791aaae717f9/hawc/apps/assessment/models.py#L260-L265.  

I haven't heard of a case where the same assessment needs both loels and loaels, etc.

hawc/apps/animalv2/models.py
DookTibs: per latest guidance from Vidhi, dose on the animal/group level data has been changed from a fk reference to DoseGroup to a free entry field.

hawc/apps/animalv2/models.py
DookTibs: ok think we have this sorted now , stuff has been moved down to the group level data instead of the parent extraction object.

hawc/apps/animalv2/models.py
DookTibs: @shapiromatron is animalv1::DoseGroup::dose_group_id the index you mean?

hawc/apps/animalv2/models.py
DookTibs: ok to use the enum; I've switched both fields to use that.

hawc/apps/animalv2/models.py
DookTibs: confirmed on phone, "yes" that's the one

shapiromatron: Yes the `dose_group_id` is the index I meant. Sorry for not being clear earlier!

hawc/apps/animalv2/models.py
DookTibs: I think in general those were ones I just brought over from animalv1 unchanged; pretty sure I got all of these fixed now, does look nicer that way.

DookTibs:
@shapiromatron ok, I think all feedback items are addressed. Let me know if I missed anything or if there's anything else you'd like on this one.

shapiromatron (approved):
LGTM!

PR #953, opened 2023-11-21
make searches reflect `expanded` prop type

shapiromatron:
We received feedback from some users that the reference search/import information can be useful during tagging.  Therefore, we added back the full search/import list on the detail view of a reference (on reference detail pages as well as reference tagging).

On list views, the collapsed button group is shown instead. 

**Tagging/reference detail view:**
image

**List view (see the source button in the actions menu):**
image


caseyhans (approved):
Removed the label in favor of icons. Good to merge if you're OK with the change!

shapiromatron:
> Removed the label in favor of icons (see updated PR description). Good to merge if you're OK with the change!

Thanks for the review! I ended up reverting the change; scientists requesting restoring the prior UI.

PR #955, opened 2023-11-27
Update AssessmentValue global API

shapiromatron:
Rewrite `AssessmentValue` export to improve performance.

Updated the API to also allow for filtering for basic assessment information, as well as casrn or dtxsid, for example `cas=7681-49-4&dtxsid=DTXSID20220630`. 

This PR also includes a new helper method, `hawc.apps.common.filters.filtered_qs`, which allows you to use a FilterSet in a DRF view and filter the queryset given any user specified query parameters.


hawc/apps/common/exports.py
```diff
@@ -152,7 +152,7 @@ def format_time(self, df: pd.DataFrame) -> pd.DataFrame:
         tz = timezone.get_default_timezone()
         for key in [self.get_column_name("created"), self.get_column_name("last_updated")]:
             if key in df.columns:
-                df.loc[:, key] = df[key].dt.tz_convert(tz).dt.strftime("%Y-%m-%dT%H:%M:%S.%f%z")
+                df[key] = df[key].dt.tz_convert(tz).dt.strftime("%Y-%m-%dT%H:%M:%S.%f%z")
```
dannypeterson, line 155: @shapiromatron discovered the reason `format_time` not working is due to pandas update to 2.1.1 in PR #928. In [this update, ](https://pandas.pydata.org/docs/whatsnew/v2.0.0.html): 'Changed behavior in setting values with df.loc[:, foo] = bar or df.iloc[:, foo] = bar, these now always attempt to set values inplace before falling back to casting ([GH 45333](https://github.com/pandas-dev/pandas/issues/45333))'.

hawc/apps/assessment/exports.py
```diff
@@ -1,9 +1,108 @@
...
+                query_prefix + "peer_review_status", constants.PeerReviewType
+            ),
+        }
+
+
+class AssessmentExporter(Exporter):
+    def build_modules(self) -> list[ModelExport]:
+        return [
+            AssessmentExport("assessment", "assessment"),
+            AssessmentDetailExport("assessment_detail", "assessment__details"),
```
dannypeterson, line 97: Found a bug if a user tries to filter for an assessment with an AssessmentValue but not AssessmentDetail, you cant filter for that assessment: `AttributeError: Can only use .dt accessor with datetimelike values`.  I think because in this join table the AssessmentDetail `Created` field wont exist... looking into it more.

shapiromatron: I think we might need a check before casting the datetimes?

```python
df = pd.DataFrame(data=[[1,None,],[None,None]], columns=['a','b'])
>      a     b
> 0  1.0  None
> 1  NaN  None

df['a'].isnull().all()
> False

df['b'].isnull().all()
> True
```

hawc/apps/common/exports.py
dannypeterson: I can create a ticket for that

PR #958, opened 2023-12-08
Invitro export rewrite

rabstejnek:
Update invitro exports to use ORM values instead of expensive serializers.


hawc/apps/invitro/exports.py
```diff
@@ -1,362 +1,622 @@
...
+        dsstox_cols = [col for col in df.columns if col.startswith("dsstox-")]
+        dsstox_df = df[dsstox_cols]
+        dsstox_df.columns = dsstox_df.columns.str[7:]
+        df["chemical DTXSID"] = dsstox_df.to_dict(orient="records")
+        df["chemical DTXSID"] = df["chemical DTXSID"].apply(
+            lambda x: None if pd.isna(x["dtxsid"]) else x
+        )
+        return df.drop(columns=dsstox_cols)
+
+    def handle_dose_groups(self, df: pd.DataFrame) -> pd.DataFrame:
```
shapiromatron, line 236: wow, this is impressive, and really complicated. I think keeping it as is is good for now, this seems like a conservative conversion from our initial approach. However, when I try it on a large dataset, it's slow. Not super slow, so it may be good enough.

I'll make a card in our backlog to explore more aggressive optimizations. It'll be easier to explore the impact of the optimizations once we have the jupyter notebooks you'll be building for comparing different approaches.

To document some exploration ideas, I got pretty far with this which ran really quickly:

```python
df = pd.DataFrame(IVEndpointGroup.objects.filter(endpoint__assessment=123).order_by('endpoint', 'dose_group_id').values())
col_vals = ['dose','n','response','variance','difference_control','significant_control','cytotoxicity_observed']
df1 = pd.pivot(df.query('dose > 0'), values=col_vals,index=['endpoint_id'],columns=['dose_group_id'])
df1.columns = df1.columns.map(lambda el: f'{el[0].replace("_", " ").title()} {el[1]+1}')
df1.fillna("-")
```

hawc/apps/invitro/exports.py
```diff
@@ -1,362 +1,622 @@
...
-                    ser["observation_time_units"],
-                    getDose(ser, "NOEL"),
-                    getDose(ser, "LOEL"),
-                    ser["monotonicity"],
-                    ser["overall_pattern"],
-                    ser["trend_test"],
-                    doseRange[0],
-                    doseRange[1],
-                    number_doses,
+    def handle_benchmarks(self, df: pd.DataFrame) -> pd.DataFrame:
```
shapiromatron, line 297: A similar approach to EndpointGroup could probably be applied here. Again, in the future, once we've got our comparisons scripts complete.

hawc/apps/invitro/exports.py
```diff
@@ -1,362 +1,622 @@
...
+            StudyExport(
+                "study",
+                "experiment__study",
+                include=("id", "short_citation", "study_identifier", "published"),
+            ),
+            IVChemicalExport(
+                "iv_chemical",
+                "chemical",
+            ),
+            DSSToxExport(
```
shapiromatron, line 427: I think `DSSToxExport` can be removed, and we add `dtxsid_id` to the `IVChemicalExport`.

hawc/apps/invitro/exports.py
```diff
@@ -1,362 +1,622 @@
...
+            return group_df.apply(__func, axis="columns")
+
+        return (
+            df.groupby("iv_endpoint-id", group_keys=False)
+            .apply(_func)
+            .drop(columns="iv_endpoint-data_type")
+            .reset_index(drop=True)
+        )
+
+    def handle_dsstox(self, df: pd.DataFrame) -> pd.DataFrame:
```
shapiromatron, line 511: handle dsstox as above

shapiromatron (approved):
I think this is really close!

We should fix how dtxsid is handled (to fix the prior error in implementation).  Please take a look at my changes here to make sure you're ok with my revisions:

* f6b1449aa add endpoint categories (19 hours ago) {Andy Shapiro}

---

I'll also create a card to look into some more aggressive optimizations in the future, but I think what you did here was perfectly appropriate and it should be simpler in the future to investigate larger changes after we have our comparison scripts online.

rabstejnek:
> I think this is really close!
> 
> We should fix how dtxsid is handled (to fix the prior error in implementation). Please take a look at my changes here to make sure you're ok with my revisions:
> 
> * [f6b1449](https://github.com/shapiromatron/hawc/commit/f6b1449aa69cb9d5b00561a80b4aa02588704490) add endpoint categories (19 hours ago) {Andy Shapiro}
> 
> I'll also create a card to look into some more aggressive optimizations in the future, but I think what you did here was perfectly appropriate and it should be simpler in the future to investigate larger changes after we have our comparison scripts online.

Your changes look good! I agree about optimizing in future PRs, there's definitely some work to do there.

As for the DTXSID changes, I kept the `DSSToxExport` for possible future use but removed it from the current exporters in favor of just the DTXSID:

image

If these changes are fine with you then I'm fine with merging!

shapiromatron (approved):
ok to merge!

PR #959, opened 2023-12-13
migrate to Django 5.0.0

munnsmunns:
Migrates HAWC to Django version 5.0.0

The only thing I could find that broke was wagtail, which required a small update to the next version

I wasn't able to find anything in our codebase that is on the list of [backwards incompatible changes](https://docs.djangoproject.com/en/5.0/releases/5.0/#backwards-incompatible-5-0).

### DIFF
--- a/hawc/apps/animal/managers.py
+++ b/hawc/apps/animal/managers.py
@@ -1,4 +1,3 @@
-from copy import deepcopy
 from typing import Any
 
 import numpy as np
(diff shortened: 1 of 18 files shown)


shapiromatron (changes requested):
Nice, that's fantastic news.

1. Can you run [django-upgrade](https://github.com/adamchainz/django-upgrade)? It should remove all the choices attributes from ChoiceFields; I'm not sure if it'll do anything else
2. Are there any deprecation warnings that we'll need to fix in the tests?

shapiromatron (approved):
LGTM! nice job!

PR #960, opened 2023-12-13
Fix font 404s from EPA styling

caseyhans:
Fix the font 404 errors generated by the relative URLs in the EPA's stylesheet


shapiromatron (approved):
LGTM - I updated our dev docs to add this step in our update checklist.

PR #961, opened 2023-12-14
Animal export rewrite

rabstejnek:
Update animal exports to use ORM values instead of expensive serializers.


hawc/apps/animal/exports.py
```diff
@@ -1,744 +1,1520 @@
...
+                "endpoint-variance_type_name": "endpoint-variance_type",
+                "animal_group-species_name": "species-name",
+                "animal_group-strain_name": "strain-name",
+            }
+        )
+
+        return df
+
+
+class AnimalExporter2(Exporter):
```
rabstejnek, line 473: At the risk of making the `FlatFileExporter` cluttered, maybe these should be inner classes on those? ie:
```
class EndpointGroupFlatDataPivot(FlatFileExporter):
    _Exporter(Exporter):
        ...
```
That way we won't have to have nebulous naming on these, cause I'm not sure `EndpointGroupFlatDataPivotExporter` is any better of a name than `AnimalExporter2`.

hawc/apps/animal/exports.py
```diff
@@ -1,744 +1,1520 @@
...
+            "parents_display": "parents_display",
+            "generation": "generation",
+            "comments": "comments",
+            "diet": "diet",
+            "species_name": "species__name",
+            "strain_name": "strain__name",
+        }
+
+    def get_annotation_map(self, query_prefix):
+        SexSymbol = type(
```
shapiromatron, line 140: these are really confusing; I think there are 3 of them. Can we make a helper method that just makes an Choice field out of these, or even better, update sql_display so that it would accept a dict

hawc/apps/animal/exports.py
```diff
@@ -1,744 +1,1520 @@
...
+            "duration_exposure_text": "duration_exposure_text",
+            "duration_observation": "duration_observation",
+            "num_dose_groups": "num_dose_groups",
+            "positive_control_display": "positive_control_display",
+            "negative_control_display": "negative_control_display",
+            "description": "description",
+        }
+
+    def get_annotation_map(self, query_prefix):
+        PositiveControl = type(
```
shapiromatron, line 174: see above comment

hawc/apps/animal/exports.py
```diff
@@ -1,744 +1,1520 @@
...
+            "trend_result_display": "trend_result_display",
+            "diagnostic": "diagnostic",
+            "power_notes": "power_notes",
+            "results_notes": "results_notes",
+            "endpoint_notes": "endpoint_notes",
+            "additional_fields": "additional_fields",
+        }
+
+    def get_annotation_map(self, query_prefix):
+        VarianceName = type(
```
shapiromatron, line 242: and here

hawc/apps/animal/exports.py
```diff
@@ -1,744 +1,1520 @@
...
-        rows = []
-        identifiers_df = Study.identifiers_df(self.queryset, "animal_group__experiment__study_id")
-        for obj in self.queryset:
-            ser = obj.get_json(json_encode=False)
-            row = []
-            row.extend(
-                Study.flat_complete_data_row(
-                    ser["animal_group"]["experiment"]["study"], identifiers_df
+    def handle_doses(self, df: pd.DataFrame) -> pd.DataFrame:
+        # this is really slow; maybe its the filtering to find matching dose group ids?
```
shapiromatron, line 347: prefix with # TODO so it's easy to find this comment

hawc/apps/animal/exports.py
shapiromatron: I agree; I don't like the AnimalExporter2,AnimalExporter3,AnimalExporter4 names. I think I'd prefer you second convention, which just has long names, but that's ok.

Maybe we can call it `EndpointGroupDataPivotExporter`?

hawc/apps/animal/exports.py
```diff
@@ -1,744 +1,1520 @@
...
+            ):
+                (
+                    row["endpoint_group-lower_ci"],
+                    row["endpoint_group-upper_ci"],
+                ) = dich_ci(incidence, n)
+            return row
+
+        return df.apply(_func, axis="columns")
+
+    def handle_stdev(self, df: pd.DataFrame) -> pd.DataFrame:
```
shapiromatron, line 608: I noticed that some code was almost identical in a few places; we could refactor in the future, but it's fine to keep as is for now, since we're doing our evaluation to make sure it returns identical results.

shapiromatron (approved):
It's too bad that implementing this way nearly doubled the lines of code, and a lot of the code seems really complex and write-only; I had a hard time understanding what it did. That said, its amazing that the exports so far look so close to what they looked like before with an entirely different implementation. I'm looking forward 

A few comments for things to change, but looks good overall.

hawc/apps/animal/exports.py
rabstejnek: I updated the `sql_display` method to accept a `Choice` class or `dict` object.

hawc/apps/animal/exports.py
rabstejnek: The convention I used is `{ClassName}Exporter` for the corresponding exporters.

hawc/apps/animal/exports.py
rabstejnek: At the top of the file I tried to make reusable functions for some of the calculations, but there are some pandas specific code that I didn't refactor, so I agree there's some work to do there. I believe this is better for a future PR as well, since there may be refactoring we can do between export files too which would fall outside the scope of this PR.

shapiromatron (approved):
LGTM - nice job!

PR #962, opened 2023-12-15
Prefilter experiment type

caseyhans:
Adds ability to prefilter bioassay endpoints by experiment type for datapivot visuals.


shapiromatron (approved):
LGTM - I made two other edits to make it fully complete with all the other animal prefilters that we use

* e9d0c6078 add experiment type to study level prefilters as well 
* d7d38da4d filter choices to only show those available in an assessment

Please merge if you're ok with it!

PR #808, opened 2023-04-13
Recreated mgmt views using django templates instead of JS

dannypeterson:
### Recreated frontend/mgmt using django templates by updating views
- Recreated user task list in user_task_list.html and added crispy form filter
user task list

- Recreated user_assessment_task_list.html in templates and added crispy filter
- Recreated assessment_dashboard.html page in templates
- Recreated task list page in assessment_details.html and added crispy filter
task details

- Used HTMX to recreate update forms
- Deleted JavaScript counterparts in frontend/mgmt folder

PR #964, opened 2023-12-19
Risk Of Bias review tasks dashboard

shapiromatron:
A prior PR, #808, removed what we thought was an unused list of study evaluations a user is currently assigned but have not yet completed. We heard from users that this feature was used, so we've added it back in this PR.

- Add a list of incomplete study evaluations at the assessment mgmt level, as well as the overall user level
    - `/mgmt/assessment/:id/tasks/`
    - `/mgmt/tasks/`

image

--- 

In addition, updated the navbar to only show the "about" and "resources" pages if the user is not logged in and the new documentation tab is not enabled. I found that even with a "wide" laptop monitor of 1440px, with the EPA flavor of the styles, it wouldn't fit on the a single line with the HAWC title as well.

### DIFF
--- a/hawc/apps/mgmt/templates/mgmt/fragments/rob_reviews.html
+++ b/hawc/apps/mgmt/templates/mgmt/fragments/rob_reviews.html
@@ -1,5 +1,5 @@
 <p class="text-muted my-2">This lists shows all {{rob_name|lower}} reviews which are currently assigned to you that require completion.</p>
-<div class="list-group">
+<div class="list-group mb-3">
     {% for review in rob_reviews %}
     <a class="list-group-item list-group-item-action dark-link" href="{{review.get_edit_url}}">{% icon 'fa-edit' %}&nbsp;{% if show_assessment %}{{review.study.assessment}}&nbsp;|&nbsp;{% endif %}{{review.study.short_citation}}&nbsp;({{review.final|yesno:"final,individual"}} review)</a>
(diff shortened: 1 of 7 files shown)


caseyhans:
This list includes tasks for published assessments, which are tasks that users can't work on/submit. I'm not sure if this is as intended but it's worth noting. Overall, it looks good to me, but I did make some minor style tweaks:
-  Added a `.dark-link` class so all links in an list-item-action list are the same color, regardless of whether they've been visited or not. Abstracted to a separate class since it might be useful in other locations
-  Removed bottom margins for the new subheaders
- Fixed the 'logout' button active state color; it's bothered me for a while, and we're touching nearby code anyway

caseyhans (approved):
LGTM, see above comment.

shapiromatron:
@caseyhans we could filter for `study__assessment__editable=True` to only show those studies which are on editable assessments. I think our test fixture may just be a bad example; seeing all the stuff that's on your queue might be ok with real world projects. Are you ok keeping it as is, or should we add this additional filter?

caseyhans:
> @caseyhans we could filter for `study__assessment__editable=True` to only show those studies which are on editable assessments. I think our test fixture may just be a bad example; seeing all the stuff that's on your queue might be ok with real world projects. Are you ok keeping it as is, or should we add this additional filter?


I'm good with keeping it as is!

PR #968, opened 2024-01-13
fix flaky test `tests/integration/test_lit::test_workflows.py`

shapiromatron:
The lit integration tests were not reliably finding the selectors which area needed for the tests. This PR makes a few minor updates to make the selectors more resilient:

- use text based selectors instead of ID based selectors
- use a regex for the count match instead of an exact count; the count can change depending on which tests run before


shapiromatron:
Currently getting an error locally when repeating 10x:

```python
        expect(page.get_by_text("in Title/Abstract 2")).to_be_visible()
        expect(page.get_by_role("link", name="1  needs tagging")).to_be_visible()
>       expect(page.get_by_role("link", name="1 ≠ with conflicts")).to_be_visible()
E       AssertionError: Locator expected to be visible
E       Actual value: None
E       Call log:
E       LocatorAssertions.to_be_visible with timeout 5000ms
E         - waiting for get_by_role("link", name="1 ≠ with conflicts")

tests/integration/test_lit.py:136: AssertionError
```

caseyhans (approved):
LGTM!

PR #972, opened 2024-01-25
Add pagination to the references list view

rabstejnek:
The references list view used to render all references for a given tag. However, if there were hundreds or thousands of references for a given tag, it became slow to load and hard to navigate for a user.

This pull request adds pagination to this view. Counts in the tag tree and histogram remain the same, however only 25 references are loaded at a time, and the select box on the left only contains the references on the page.

A more intrusive solution that could speed it up further would be to make an API call that returns a paginated response. However, both the tag tree and year histogram rely on the full list, and the tag tree and histogram are used in other places besides this view so it would be a much bigger change. A future PR could involve making API endpoints that retrieve counts for the tag tree and counts for the year histogram, and then using a paginated API call for the reference list.

### DIFF
--- a/frontend/lit/ReferenceTreeBrowse/Main.js
+++ b/frontend/lit/ReferenceTreeBrowse/Main.js
@@ -4,6 +4,7 @@ import {toJS} from "mobx";
 import {inject, observer} from "mobx-react";
 import PropTypes from "prop-types";
 import React, {Component} from "react";
+import HelpText from "shared/components/HelpText";
 import Loading from "shared/components/Loading";
 import TextInput from "shared/components/TextInput";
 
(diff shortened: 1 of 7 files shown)


shapiromatron (approved):
I think this looks great, and this was a smart way to do this initially by just dealing with the rendering component instead of doing the frontend and backend all at once. Hopefully this is good enough for most situations.

I made a number of minor edits, and there was a little scope creep with what I did, but ok to merge if you're happy with my changes.


* f9dd26d8c make modal title a clickable link to reference browse (2 minutes ago) {Andy Shapiro}
* 8fad75c0c camelCase javascript variables (8 minutes ago) {Andy Shapiro}
* ec176c11f page top header h3 instead of h4 (9 minutes ago) {Andy Shapiro}
* e38361b97 simplify code; remove autorun (15 minutes ago) {Andy Shapiro}
* 8f2498943 use label; add second paginator (18 minutes ago) {Andy Shapiro}
* 10bf2c39d use a label instead of help text (38 minutes ago) {Andy Shapiro}
* e75d27e14 tighten up types (39 minutes ago) {Andy Shapiro}
* 2c6b76b35 show 50 items per page (60 minutes ago) {Andy Shapiro}

rabstejnek:
@shapiromatron Changes look good! I did add back `autorun` so that the page number resets to 1 whenever the filters change, just to make sure the page number never falls out of bounds of the current filters. I added a comment in the code to explain why we need this as well.

PR #974, opened 2024-01-30
change minus to times

caseyhans:
Change the symbol for 'author contacted' from a minus to a times. This changes the symbol in other locations too, but it's prety minor.

image
image


shapiromatron (commented):
Ahh yes, this is a much better icon. I searched the codebase and saw it was used in some other places too - do you think it should be replaced in these cases too, or is there ever a case for the minus?

image

caseyhans:
> Ahh yes, this is a much better icon. I searched the codebase and saw it was used in some other places too - do you think it should be replaced in these cases too, or is there ever a case for the minus?
> 
> image

Most of these cases are using minus as a minimize symbol, which is a good use case for that symbol.
The exception is in the study list, where minus indicates that a study is not of a particular type (i.e., bioassay, epidemiology, etc.). I tried swapping for the times symbol here but I felt that the similarity to the check mark made it harder to determine the study type(s) at a glance.

Example: 
image

image

shapiromatron (approved):

shapiromatron:
> > Ahh yes, this is a much better icon. I searched the codebase and saw it was used in some other places too - do you think it should be replaced in these cases too, or is there ever a case for the minus?
> > image
> 
> Most of these cases are using minus as a minimize symbol, which is a good use case for that symbol. The exception is in the study list, where minus indicates that a study is not of a particular type (i.e., bioassay, epidemiology, etc.). I tried swapping for the times symbol here but I felt that the similarity to the check mark made it harder to determine the study type(s) at a glance.
> 
> Example: image
> 
> image

Completely agree. Nice job!

PR #975, opened 2024-01-31
Remove try/except now that django-filters has a fix

munnsmunns:
Removes an unnecessary try/except now that the issue has been fixed in django-filters.


shapiromatron (approved):
nice!

PR #976, opened 2024-01-31
Format Django HTML templates

munnsmunns:


munnsmunns:
@shapiromatron After doing a bit of research/testing, I think we should consider djlint instead of djhtml since it's easier to set up in everyone's environment. Let me know what you think, I've laid out some pros/cons here

djhtml
- no dependencies, runs faster
- no VSCode extension, and the [proposed workaround](https://github.com/rtts/djhtml/issues/108) would require manually adding a keybinding rather than being executed on save. There might be a way to configure it to act on save, but I haven't found it yet after a little digging

djlint
- has a bunch of dependencies, runs slower but still reasonably fast in my testing. a bit bloated as it supports a lot of file types that we don't use
- has a vscode extension which makes it easy to set up auto formatting

shapiromatron:
@scott-8 - did you do anycomparison of [djhtml](https://github.com/rtts/djhtml) vs [djlint](https://github.com/djlint/djlint)? They both have around the same number of stars on github.

PR #977, opened 2024-01-31
Candidate tag filtering bug

caseyhans:
Fixes a bug where candidate tag addition filtering shows extraneous references. The filterset now uses similar filtering to the bulk merge tag conflicts view.

Moves the 'View Workflows' dropdown link so team members can see the link. This lets team members see how workflows are defined. Editing/creating workflows is still restricted to project managers.


shapiromatron (approved):
LGTM

PR #978, opened 2024-02-01
Handle and display HTMX 400/500 errors

munnsmunns:
Ports over HERO's handling of HTMX errors. Now when a 400/500 error occurs in an HTMX request, a message is displayed to the user.


shapiromatron (approved):
nice job!

PR #982, opened 2024-02-02
Fix epi orm exports

rabstejnek:
There are two issues in the epi exports that are addressed in this PR:
- If an empty queryset is used in one of the exporters it would break when attempting to access properties on the first object in the queryset (ie None).
- If `groupby` results in no groups then `apply` calls would just return an empty dataframe with all the original columns; this behavior was interfering with later `join` calls in one of the exporters.


shapiromatron (approved):
LGTM

PR #983, opened 2024-02-05
hotfix - fix generational animal group form

shapiromatron:
Fix change in how Django 5 handles choice fields so that the choices are sliced directly instead of the form model field choices, which no longer seems to work. I did a quick keyword search in the codebase and didn't see any other cases where this may have been an issue.

Also wrote a test to test the behavior on the views, and raised code coverage for the forms and views:

image


rabstejnek (approved):
Tested it out, looks good!

PR #985, opened 2024-02-06
Add formatting for HTML templates

munnsmunns:
Adds formatting for HTML templates via djHTML. Formatting is applied via a pre-commit, and can also be manually applied using `make format-html`.

We had to manually update an SVG element which was really long since the formatter allows a maximum length of 10,000 (see the header.html SVG):

```python
import textwrap
long_attr = "..."
print("\n".join(textwrap.wrap(long_attr, 5000)))
```

HTML allows for line breaks within an attribute; the page appears to render fine on modern browsers.

---

When using pre-commit, if you try to commit and the validation fails, it will throw an exception and try to format the files properly, but not commit anything automatically. As an example, in this example, I had a badly formatted html template:

```
(venv) ► git commit -m "test"
DjHTML...................................................................Failed
- hook id: djhtml
- files were modified by this hook

reindented hawc/apps/animal/templates/animal/_animal_group_list_tbl.html
1 template has been reindented.

ruff.................................................(no files to check)Skipped
ruff-format..........................................(no files to check)Skipped
```

It fixes the issue, but leaves the changes unstaged, so I'll need to manually stage and then try to commit again.


make.bat
```diff
@@ -40,6 +40,7 @@ echo.  lint-py           check python formatting issues
 echo.  format-py         fix python formatting issues where possible
 echo.  lint-js           check javascript formatting issues
 echo.  format-js         fix javascript formatting issues where possible
+echo.  format-html       fix html formatting issues where possible
```
shapiromatron, line 43: can you add this to the `format` command as well?

shapiromatron (commented):
Looking good! Would you mind adding an update to the `development.md` docs file to just give a quick description of how to install or run pre commit? I think you have to run an init command or something to get it linked with the git repo, is that right?

shapiromatron:
@caseyhans @rabstejnek - please check this one out and give it a test drive, if we're all ok with how it works, I think we can merge!

shapiromatron (approved):
nice, excited to add this!

rabstejnek (approved):

caseyhans (approved):

PR #986, opened 2024-02-06
Custom filter logic helper

rabstejnek:
Added a helpful representation of a user's custom filter logic directly below their input on the exploratory heatmap forms.

image

It is indented and the grouping simplified in the representation, making it easier for a user to visually see what logic they have written. It also informs the user in real time if the logic string is non-parseable.


shapiromatron (approved):
Great idea to use the parser output; I just edited what the query component looks like, but didn't touch the logic:


```
(
  (
    "study id" ≥ 123 {1}
    AND
    "system" contains "hepatic" {2}
  )
  OR
  "sex" contains "male" {3}
)
```

PR #987, opened 2024-02-06
data pivot - add columns to the right of the figure

shapiromatron:
Add new option to add content to the right of the plot in a data pivot.

Check an item "to right":

Screenshot 2024-02-06 at 6 02 09 PM

image

image

Related work in #991 adds calculated columns as a new option, to compose multiple columns in a dataset to a new column.


rabstejnek (approved):
Looks good! I did notice some edge cases though that aren't rendering as you would expect; I'll leave it up to you if they need to be addressed before merging, they are relatively minor.

- When text background is highlighted but not behind the plot, the columns on the right aren't highlighted.
  image

- When all columns are to the right and text background is highlighted, there is some leftover highlighting on the left of the plot.
  image

PR #988, opened 2024-02-07
Improvements to client visual downloader

munnsmunns:
Improve resilience of the visual downloader after trying it on a suite of real world assessments. This required making some edits to both the frontend and the backend to properly prompt our client package when a visualization error occurs.

- Adds argument to set the default timeout in the interactive client
- Adds handling for additional visual types that were originally overlooked
- Adds ability for script to detect error messages to improve speed when downloading multiple images


shapiromatron (approved):
working pretty well, though hopefully we'll figure out new ways to improve in the future as we continue to use.

PR #990, opened 2024-02-09
fix automatic scrolling with htmx requests

caseyhans:
Fixes the automatic scrolling on the workflows list page while CRUDing, and adds an animation for smoothness.
Previously, updating or creating a workflow would snap focus to the name input field of the first form present on the page (which creates a bug if multiple forms are present). After cancelling or saving the form, users would be left looking at the page footer, since the form is so long and the save/cancel buttons are at the bottom. This PR saves the scroll position before a form is brought onto the page and then reinstates that scroll position once form operations are complete.
- List is now always ordered by creation time, descending (newest first)
- Create form moved to top of the list
- Fixes script placement for help text popups


hawc/apps/lit/templates/lit/workflows.html
```diff
@@ -16,4 +16,44 @@ <h2 class="mb-2">Workflows</h2>
...
+        if ($(workflowForm).length) {
+            // if the form exists..
+          if ($(targetHTML).hasClass("workflow-edit-row")) {
+                // if both source and target are a form (failed update/create), maintain scroll attr between elements
+            $(workflowForm).attr("scrolltop", $(targetHTML).attr("scrolltop"));
+          } else {
+                // otherwise, set scroll attr to current scroll position so we can return later
+            $(workflowForm).attr("scrolltop", $("body,html").scrollTop());
+          }
+          $("body,html").animate({scrollTop: $(workflowForm).offset().top - 20}, 400); // now animate a scroll to the form
```
shapiromatron, line 38: this seems like a reusable method we should add to to our HAWCUtils

hawc/apps/lit/templates/lit/fragments/workflow_row.html
```diff
@@ -89,6 +89,6 @@ <h4 class="mb-3">{{object.title}}&nbsp;<i class="fa fa-spinner fa-spin htmx-indi
     </div>
   </div>
   <div class="mt-2 ml-1 d-flex justify-content-center text-center">
-    <i class="m-0 text-muted" style="max-width: 800px;">{{object.description}}</i>
+    <i class="m-0 text-muted" style="max-width: 75%;">{{object.description}}</i>
```
shapiromatron, line 92: does this look weird on mobile or small widths screens?

hawc/apps/lit/templates/lit/fragments/workflow_edit_row.html
```diff
@@ -62,19 +62,21 @@ <h3 class="py-2">Create Workflow</h3>
...
+    </div>
   {% endif %}
-</div>
-</div>
-{% if form %}
-  <script type="text/javascript">
-    document.querySelector('#id_title').focus();
+  <script>
+    function delRow(elt) {
+      var scrollVal = $(elt).attr('scrolltop');
```
shapiromatron, line 77: can this be a helper method too? honestly I think we'd want this version of a deletion everywhere

shapiromatron (changes requested):
Changes look good, but I think we should refactor the code to put most of this into the HAWCUtils.js file so we can reuse in other HTMX places. The same will be true on the epiv2 and eco forms, and TBH, they're used a lot more than the workflow edit forms, so we should move the goodness over there too.

hawc/apps/lit/templates/lit/fragments/workflow_row.html
caseyhans: The whole page looks a bit off on smaller screens--this specific change isn't causing further problems. 
But overall the page looks good if the phone is rotated horizontally!

shapiromatron (approved):
LGTM - I did mostly some minor code formatting changes:

* ab436f68a delete row (3 minutes ago) 
* 8dbdf754a update naming and a few minor logic changes (8 minutes ago) 
* b0d46b7d8 rename to better reflect what it does (58 minutes ago) 
* 9961645d0 rename addScrollHTMX to addScrollHtmx (4 hours ago) 
* 4edbd8472 rename delRow to deleteRow (4 hours ago)

caseyhans:
@shapiromatron had to add back delete for workflow edit rows, or the create button scroll doesn't work after the first time

PR #991, opened 2024-02-13
data pivot - calculated columns

shapiromatron:
Enable the ability to add calculated fields in a data pivot whose output is derived from other fields in the dataset. Coupled with #987, this allows for users to mix and match columns to present to the right of a data pivot using information presented in the visual.

For example:

- The endpoints column: `${endpoint name} (${response units}) @ ${observation time text}`
- The dose column: `match(dose,0)?-:${dose} ${dose units}`
- The response column: `${response} ± round(stdev,2) (round(percent control mean,1)%)`

Renders:

image


frontend/summary/dataPivot/DataPivotVisualization.js
```diff
@@ -196,6 +196,28 @@ class DataPivotVisualization extends D3Plot {
...
+            });
+
+            // add data-pivot row-level key and index
+            row._dp_y = i;
+            row._dp_pk = row["key"] || i;
+
+            settings.calculated_columns
+                .filter(d => d.name && d.formula)
+                .forEach(d => {
+                    row[d.name] = d.formula;
```
shapiromatron, line 214: @rabstejnek - could start incorporating here?

rabstejnek:
@shapiromatron Ready for review!
- You can include a column value by doing `${column}`
  - ie `This is a ${species}.` with `{species: "rat"}` returns `This is a rat.`
- You can use ternaries by doing `match(column,str/int)?Truthy string:Falsey string`
  - ie `match(species,"cat")?This is a cat.:This is not a cat.` with `{species: "rat"}` returns `This is not a cat.`
- You can escape characters using `\`
  - ie `\${species}` would return `${species}`

rabstejnek:
@shapiromatron I have updated the parser language so that it can handle nested ternaries correctly; an example is in the test suite. I also added the `exists` condition with some tests. The logic for `exists` is as follows:
```
return options.getValue(id) != null & options.getValue(id) != "";
```

There is also an npm script that generates the `query` and `format` parsers:
```
npm run generate-parsers
```

shapiromatron (commented):
@rabstejnek I think you have to do the final review b/c I opened the PR initially, but I think this is great and will be a I think a highlight of our next release.

I made the following edits:

* 782feb498 fix docs (facepalm) (17 seconds ago) {Andy Shapiro}
* 19433ecf8 add some text to describe how to use (50 minutes ago) {Andy Shapiro}
* b3d90723d fix test case (80 minutes ago) {Andy Shapiro}
* 3be80dbd9 add round (2 hours ago) {Andy Shapiro}
* 99aef96ad add "to_right" to fixture; should have added in #987 (3 hours ago) {Andy Shapiro}
* 9d17e63da use latest peggy version (6 hours ago) {Andy Shapiro}
* 11883b88d fix migration conflict (6 hours ago) {Andy Shapiro}

If you're ok w/ the changes, please approve.

---

To document our sidebar chats, but it'd be good to get this in use initially to determine the need first:

- we may change `${}` to `show()` in the future
- we may rewrite the ternary expression `a?b:c` to `ifelse(a,b,c)`

rabstejnek (approved):

PR #992, opened 2024-02-13
Htmx viewset changes

rabstejnek:
Refactored htmx viewset paths so that the action is embedded into the path instead of defining a path for each action. The only `HtmxViewset` that wasn't refactored was `hawc.apps.assessment.views::PublishedItemsChecklist`, since this viewset had paths with a different number of args depending on the action.


shapiromatron (approved):
LGTM! I fixed a bug where the cancel buttons no longer works; this is because in the old way, the action was read, but the url name was detail, so we had to change the routes to use read instead of detail.  In the `urls.py`: 

```python
# this was the old routes that were deleted
path(
    "chemical/<int:pk>/",
    views.ChemicalViewSet.as_view(),
    {"action": "read"},
    name="chemical-detail",
)
```

Ok to merge!

PR #994, opened 2024-02-15
Flush and rebuild vcr, replace pytest vcr with pytest recording

dannypeterson:
Flushed and rebuilt VCR. `pytest-vcr` package is no longer supported, so replaced with `pytest-recording`.  Removed urllib3 because it is a dependency of the requests package. A list of changes:
  - Updated requirements/dev.txt 
  - Updated pyproject.toml
  - Updated github actions workflow
  - Deleted old cassettes and rebuilt with pytest recording, checked with wifi disabled


shapiromatron (commented):
nice job!

shapiromatron (approved):

PR #996, opened 2024-02-23
hotfix - add a delay for adding time spent

shapiromatron:
Add a 10s delay for adding time spent to ensure existing rows have been created and foreign key relations will be valid. This uses [countdown](https://docs.celeryq.dev/en/stable/userguide/calling.html#eta-and-countdown) to schedule 10 seconds in the future.

Using countdown with a redis broker does have a warning in the celery docs:

> Tasks with eta or countdown are immediately fetched by the worker and until the scheduled time passes, they reside in the worker’s memory. When using those options to schedule lots of tasks for a distant future, those tasks may accumulate in the worker and make a significant impact on the RAM usage.

> Moreover, tasks are not acknowledged until the worker starts executing them. If using Redis as a broker, task will get redelivered when countdown exceeds visibility_timeout (see [Caveats](https://docs.celeryq.dev/en/stable/getting-started/backends-and-brokers/redis.html#redis-caveats)).

We're using a `visbility_timeout` of 1hr w/ redis (default), so it seems unlikely that an additional 10s wait will make it too close (https://docs.celeryq.dev/en/stable/getting-started/backends-and-brokers/redis.html#visibility-timeout).


munnsmunns (approved):
Looks good! It makes sense that we shouldn't have any impact from those caveats, since tasks are only existing in ram for an extra 10s

PR #997, opened 2024-02-23
bugfix- restrict EHV tag API to accept only integer for detail ID

shapiromatron:
Restrict the EHV API to only accept an integer as an ID for the EHV API views. Previously any slug was allowed; this tightens up the API and improves error messaging.


munnsmunns (approved):
Did a quick test, it's good to go!

PR #998, opened 2024-02-23
expand options for assessment details; selectively show in template

shapiromatron:
Expand the options which are available in the assessment details to include some fields which did not previously exist; selectively show these fields on the assessment details page.


munnsmunns (approved):
Looks good!

PR #1001, opened 2024-02-26
Assessment Search Duplicate Fix

caseyhans:
Adds a .distinct() call to the assessment filterset search field to remove duplicates. Also added the debug badge to assessment list pages to more rapidly view/copy assessment IDs when debug information is enabled.


shapiromatron (approved):
LGTM - I added the debug badge to these views as well so we can see/copy the IDs more quickly.

PR #1003, opened 2024-02-26
check ID type validation

shapiromatron:
Add an additional check to ensure the ID is valid. This was written generically, as it could be useful in a number of other bulk update situations via API.


munnsmunns (approved):
Looks good to merge!

PR #1004, opened 2024-03-01
turnstile improvements

shapiromatron:


caseyhans (approved):
LGTM!
I changed the render response for the Turnstile object to be None instead of an empty string. The empty string was causing the django crispy fields renderer to look for a "" field, which raised a console error since the "" field doesn't exist on the form. [90da45f](https://github.com/shapiromatron/hawc/pull/1004/commits/90da45fc1d099b61615ca8f64935b9d2aabb2924)

PR #1005, opened 2024-03-05
add additional fields to data pivot list API

shapiromatron:
Add a few additional fields to the data pivot list API, most notably the published flag, to make it comparable to visual list API.


munnsmunns (approved):
LGTM!

PR #1009, opened 2024-03-13
bugfix - fix "needs further tagging" after consensus tags applied

shapiromatron:
Fix "needs further tagging" count on literature overview page with conflict resolution enabled:

- Before, it was the count of references w/ <2  unresolved user reviews
- Now, it is the count of references w/ <2 unresolved user reviews AND no consensus tags

With the old way, after tags were resolved and applied, the user reviews were unresolved, so it went back into the needs tagging bucket.

@caseyhans - please give it a good review if you don't mind to make sure the logic is sound. I did some spot checking and it seemed to work ok for me...


caseyhans (approved):
Made a couple changes to the filtering logic so the "needs tagging" calculation still works for advanced workflows. I also changed the conflict resolution "[number] needs tagging" tile  to "[number] untagged and need further tagging", which is a bit clunkier but better describes the statistic.

LGTM

PR #1016, opened 2024-04-17
new permission classes

shapiromatron:
Add new permission class options for views.  Previously, the project manager and team member classes always checked not only that the user is a project manager or team-member, respectively, but also that the content they're trying to view is editable. This is ideal for views that edit content, but is not ideal for read-only views where even if the content is not editable, the team should be able to view.

This work adds two new options, `TEAM_MEMBER_EDITABLE`, and `PROJECT_MANAGER_EDITABLE`.  These options work just as `TEAM_MEMBER` and `PROJECT_MANAGER` did before. The existing classes were updated to only check for role, and not for edibility status for object.

```py
class AssessmentViewPermissions(models.IntegerChoices):
    PROJECT_MANAGER = 1
    TEAM_MEMBER = 2
    VIEWER = 3
    PROJECT_MANAGER_EDITABLE = 4  # NEW
    TEAM_MEMBER_EDITABLE = 5  # NEW
```

Views were also updated to use the new permissions.

### DIFF
--- a/docs/docs/development.md
+++ b/docs/docs/development.md
@@ -251,6 +251,16 @@ This makes the test database useful when writing new features. There are multipl
  **Team member**     | team@hawcproject.org     | pw
  **Reviewer**        | reviewer@hawcproject.org | pw
 
+
+There are currently four assessments in the database:
+
+ID | Name | Editable | Public
+---|---|---|---
+1 | Chemical Z | Yes | No
+2 | Chemical X | No | Yes
+3 | Chemical Y | Yes | Yes
+4 | Chemical A | Yes | No
+
 As new features are added, adding and changing content in the test database will be required to test these features. Instructions for loading and dumping are described below.
 
 #### Generating the test database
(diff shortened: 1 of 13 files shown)


rabstejnek (approved):
Looks good! Just a few comments and questions.
1) The test that's failing is because assessment 2 is not editable, so that needs to be fixed.
2) The logic here is that a project manager or higher is still able to edit a locked study. Is this the correct permissions?
3) Should `user_can_edit_assessment` on the `Assessment` model be updated to check if the assessment is editable?

Once the test is fixed its good to merge!

shapiromatron:
> Looks good! Just a few comments and questions.
> 
> 1. The test that's failing is because assessment 2 is not editable, so that needs to be fixed.
> 2. The logic here is that a project manager or higher is still able to edit a locked study. Is this the correct permissions?
> 3. Should `user_can_edit_assessment` on the `Assessment` model be updated to check if the assessment is editable?
> 
> Once the test is fixed its good to merge!

For #1, thanks, will do!
For #2 and #3, I completely agree that it is unclear what the correct behavior as is, and the current behavior is confusing and should be changed. I added to the roadmap for this quarter a refactoring of this mixin and all surrounding permissions logic; this PR was supposed to be a hotfix to fix one issue, but it opened up a bit of a pandora's box of other changes and things that should be done.  I'll add your comment above to our ticket, and will be addressed in a future PR.

PR #1020, opened 2024-04-22
Add PRISMA visual type view and JS stub

munnsmunns:
Add the building blocks for the PRISMA visual type. Adds a JS view for the PRISMA form that loads when it is selected from the dropdown list, which is a stub for future development of the form.


hawc/apps/summary/migrations/0052_visual_image_alter_visual_visual_type.py
```diff
@@ -33,6 +33,7 @@ class Migration(migrations.Migration):
                     (6, "exploratory heatmap"),
                     (7, "plotly"),
                     (8, "static image"),
+                    (9, "PRISMA"),
```
shapiromatron, line 36: I had to update a migration since the database choices were altered; I didn't have to create one since this doesn't change the database

shapiromatron (approved):
Looks good - I did a major refactoring to use the full JS form lifecycle, instead of the django template based once you originally implemented, this is what we'll need for this diagram since it will be heavily interactive. It took a look of code, a lot of it redundant with the exploratory heatmap JS, so it may be ripe for refactoring in the future.

That said, this PR will be a good template for how to create a new JS summary figure.

PR #1021, opened 2024-04-24
clone name and ensure under size limit

shapiromatron:
Users were having issues where when a clone was created and " (2)" was appended, it exceeded the limit for a character field. This PR adds a helper function to always create a new name that will not exceed the character limit. This will create a (2), (3), (4) ..., (9), (10), (11), etc., and will remove characters from the left-side to ensure it's under the field limit.


caseyhans (approved):
Love this!

PR #1022, opened 2024-04-24
Standardize 'User Defined Fields' for UDFs

munnsmunns:
While setting up my demo for UDFs I noticed a lot of areas that still refer to UDFs as User Defined Forms, so I did a quick search and replace. All UI should now refer to them as User Defined Fields.


shapiromatron (approved):
nice job!

PR #1023, opened 2024-05-06
Add UDF logs

caseyhans:
Adds logging for all UDF functionality, including UserDefinedForm, ModelBinding, TagBinding, ModelUDFContent, and TagUDFContent

A number of larger changers were made to make it to facilitate this task:

* Added Assessment.Log logging for UDF create/update
* Added read-only reversion admin pages, these were helpful for debugging but will also be useful in production
* Added new middleware and helper methods which allow us to get the request (and therefore the user) from anywhere in the application. This middleware should work in all known deployments on HAWC, but require the reasonable assumption that each request is on it's own thread, which may not be the case in future deployments


shapiromatron (approved):
Looks great, though I made a number of edits. I added some new admin views to make it easier to debug, but also wanted to add the user to the logs created so I ended up creating some new middleware to make that possible.  Please review and let me know if it works ok and performs as expected; if you're happy we can merge!

caseyhans:
LGTM! I fixed one bug I found; in the UDF update view, previewing a schema with required fields would trip up the actual form. The preview schema div is now emptied prior to form submission.

hawc/apps/udf/forms.py
shapiromatron, line 1: @caseyhans I did a minor update to make it consistent with how we handled this in HERO - we just disable the fieldset so that they're removed from the validation

PR #1026, opened 2024-05-07
Display Reference UDF Data

caseyhans:
Display a reference's tag binding UDF data on the reference detail and reference search pages. The reference browse page requires changing API endpoints so it does not yet show reference UDF data

List view
image

List view expanded
image

Detail View
image

Detail view expanded
image


shapiromatron (approved):
Nice job, as always. I made a few minor edits, mostly stylistic. Ok to merge if you're ok with the changes. I did have one thought which was to add a few comments on complicated functions like `_get_tag_binding_contents` - I had to read it closely to see it did all the things that it needed to do; it'd be nice as a reader to be prompted of what it does.

* 4c6b76d85 rewrite JS using a few more modern techniques (34 seconds ago) {Andy Shapiro}
* 87077d8d3 add a comment that describes it does the consensus tag check (66 seconds ago)
* 24fe3ebc4 don't show deprecated UDF in binding form (2 hours ago)                                                 
* 28aa1e86a Merge branch 'main' into display-reference-udf-content (18 hours ago)

shapiromatron:
And I fixed a regression in my changes 😅 

* 762e2e8a1 make an empty object instead of null (3 minutes ago)

PR #1027, opened 2024-05-08
Add UDF Export API

caseyhans:
Add UDF exports and endpoints for model and tag binding contents.


shapiromatron (approved):
Nice job. I made a number of changes, some editorial, some a little larger:

* 10cd51131 fix integration test
* 689e73830 only export resolved tags or children by default
* aeedd3aa3 add admin views
* 1a5b7ef70 add serializer checks
* 65a07a866 refactor apis to split "view" code from "form" code
* 461e272fa add method to validate a dictionary of data for a UDF
* c2ebff3a5 remove commented out code and add type annotation
* 9ad970938 add back field expansion to exports

There were three larger changes:

1. add back the UDF data export to show one row per field.
2. Split API 'action code' from 'form validation code'. I created a new design pattern using pydantic as an API form validation layer; I think it will work well. You use pydantic to check the REST API JSON types, and then a `model_validate` attribute to check django db models. LMK what you think.
3. Fix UDF tag data export to only show UDF data on consensus tags. If `all=1` is specified or anything else truthy, then it would show all UDF tag data.

Bonus:

- add UDF content admin views

caseyhans:
Nice changes! LGTM!

PR #1028, opened 2024-05-14
Assessment Detail/Value adjustments

munnsmunns:
2 small changes to Assessment details and values:

- change max length for assessment_details.qa_id from 16 to 32
- remove validation for assessment_value.value. Now it will display a message on the detail page for the value if the validation doesn't pass

image


hawc/apps/assessment/templates/assessment/assessmentvalue_detail.html
```diff
@@ -65,6 +65,15 @@ <h3>Assessment Value for {{object.assessment}}</h3>
       {% endfor %}
     </tbody>
   </table>
+  {% if pod_message %}
```
shapiromatron, line 68: we don't need to invite this again. Use `{% alert warning %}` and `{% endalert %}`

hawc/apps/assessment/templates/assessment/assessmentvalue_detail.html
```diff
@@ -65,6 +65,15 @@ <h3>Assessment Value for {{object.assessment}}</h3>
...
+    <div class="usa-alert usa-alert--info usa-alert--slim">
+      <div class="usa-alert__body">
+        <p class="usa-alert__text">
+          {{pod_message}}
```
shapiromatron, line 72: write the message here; and make `pod_message` the values you need to add the text

hawc/apps/assessment/views.py
```diff
@@ -419,6 +420,20 @@ class AssessmentValueDetail(BaseDetail):
...
+        # add a message if POD/UF is not equal to the value
+        obj = self.object
+        if obj.value and obj.pod_value and obj.uncertainty:
+            calc_value = obj.pod_value / obj.uncertainty
+            if not np.isclose(
+                obj.value,
+                calc_value,
+                rtol=0.01,
+            ):
+                context["pod_message"] = f"""The POD/UF = Value is different than what is reported.
```
shapiromatron, line 432: the 0.364 is hard coded; it should be `calc_value`. Make `pod_message` a dictionary and then we can edit the text in the template instead of the view. 

We may want to also add to the message something like "Note: this message is only shown to team members and higher" (and make sure that is the case).

hawc/apps/assessment/views.py
```diff
@@ -419,6 +420,20 @@ class AssessmentValueDetail(BaseDetail):
...
         context = super().get_context_data(**kwargs)
         context["adaf_footnote"] = constants.ADAF_FOOTNOTE
+        # add a message if POD/UF is not equal to the value
+        obj = self.object
+        if obj.value and obj.pod_value and obj.uncertainty:
+            calc_value = obj.pod_value / obj.uncertainty
+            if not np.isclose(
+                obj.value,
+                calc_value,
+                rtol=0.01,
```
shapiromatron, line 430: make this a single line (remove the final comma)

shapiromatron (changes requested):
Looking good; a few minor edits for cleanup

hawc/apps/assessment/templates/assessment/assessmentvalue_detail.html
```diff
@@ -65,6 +65,14 @@ <h3>Assessment Value for {{object.assessment}}</h3>
       {% endfor %}
     </tbody>
   </table>
+  {% if obj_perms.edit and value_warning %}
```
shapiromatron, line 68: @swatford any changes to the language in this message?

hawc/apps/assessment/templates/assessment/assessmentvalue_detail.html
```diff
@@ -65,6 +65,14 @@ <h3>Assessment Value for {{object.assessment}}</h3>
...
+    {% alert warning %}
+      The POD/UF = Value is different than what is reported.
+      ({{object.pod_value|floatformat:'-3'}}/{{object.uncertainty|floatformat:'-3'}} = {{value_warning|floatformat:'-3'}}, which is different than
```
shapiromatron: I didn't know about this one until now, but if you use [stringformat](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#stringformat) instead of `floatformat`, you can use a lot more options; honestly the floatformat seems pretty limited for our scientific use case https://docs.python.org/3/library/string.html#format-string-syntax

shapiromatron (approved):
LGTM - I made some edits; please see my comments, as I think you could have caught a few of these w/o me (eg., the migration)

PR #1034, opened 2024-05-15
Exploratory heatmap "count column" setting

rabstejnek:
This PR adds a "Count column" setting to exploratory heatmaps where a user can explicitly set the column to get unique counts of.

> image
> image

The same setting was also integrated into the exploratory heatmap templates.

> image
> image


shapiromatron (approved):
Great job. Made some very minor code edits during the review; please merge if you're happy:

* a321bd537 minor code edits
* 2af5596e0 update help text language

PR #1036, opened 2024-05-22
make EHV download public

caseyhans:
Unauthenticated users previously got a 405 error when trying to download the EHV data. This fixes that bug!


shapiromatron (approved):
perfecto!

PR #1038, opened 2024-05-23
fix identifier DOI dupe checks

caseyhans:
Use our DOI regex to check for duplicate DOIs when validating HERO identifiers

image


shapiromatron (approved):
LGTM - I made one minor tweak (08c77eb36d2ff9eb6524ca72ea8e58de17d3e210) to keep they types consistent with what they were before.

I think there still may be duplicates that this check doesn't catch, since this only checks HERO IDs which don't already exist in HAWC, and there could be cases where a different HERO ID has the same DOI or something to that effect, so we may need to revisit again in the future, but this is definitely an incremental improvement!

PR #1039, opened 2024-05-24
Add RoB data to EpiV2 exports

munnsmunns:
Adds RoB scores to EpiV2 data exports.


shapiromatron (approved):
I had a little fun rewriting to build the JSON w/ SQL instead of in python. Ok to merge if you're ok with my changes @munnsmunns.

PR #1012, opened 2024-04-09
Refactor HTMX Viewsets

shapiromatron:
Refactor our lit.workflows and assessment.attachments HTMX ViewSets to use new `.show-only-child` class and remove nearly-identical templates between eco and epiv2 viewsets. This uses the pattern originally used in #989.

- `show-only-child` hides an element unless it is the only child of a parent element
- `hide-empty-tbody` will hide an element unless it has a child `<tbody>` containing a `<tr>` element
- Fix deletion script for eco class; cascading deletions are now properly displayed in the eco update page (i.e., deleting a Cause will immediately remove related Results from the table)

PR #1048, opened 2024-06-17
Update Animal V2 HTMX viewsets

munnsmunns:
Updates the Animal V2 templates and viewsets for the Experiment update page to match the updated patterns in #992 and #1012.


munnsmunns:
@shapiromatron There are some utility functions and css classes that the new htmx pattern uses but that aren't in the animal v2 branch yet. should we pull from main into this branch or should that happen in the animalv2 branch first? or just copy/paste those functions into this branch?

shapiromatron:
> @shapiromatron There are some utility functions and css classes that the new htmx pattern uses but that aren't in the animal v2 branch yet. should we pull from main into this branch or should that happen in the animalv2 branch first? or just copy/paste those functions into this branch?

Whatever would be quicker!

shapiromatron (approved):

PR #1049, opened 2024-06-20
Replace <colgroup> and <thead> with bs4 tags

munnsmunns:
Replaces use of the `<thead>` and `<colgroup>` html tags with the more condensed bs4 template tags where possible. Also replaces alerts with the bs4 alert tag where possible.

Notes:
- Alerts with additional css classes were left as-is since adding those classes would require changes to the bs4 function. You can find them by searching `class="alert` in the VSCode search.


shapiromatron (approved):
Nice job, I found a few more cleanups but looks good.

PR #1050, opened 2024-06-20
Refresh VCRs for tests

munnsmunns:


munnsmunns:
On my machine I've been getting Decode Errors when running certain tests with VCRs. Refreshing the VCRs allows the tests to succeed on my machine, but they cause the CI checks to fail with XML parsing errors. I'll keep these VCRs on my local machine until I can figure out how to fix my environment.

PR #1051, opened 2024-06-21
Improve test coverage

munnsmunns:
Improves unit test coverage in these areas:

- Epi v1 client CRUD
- Summary client CRUD
- Reference list API
- Epi v1 `Result` update view with formset
- Assessment enpoints list API
- Lit tagging with UDF data
- Admin actions: 
  - `delete_orphan_tags`
  - `migrate_terms`
- `order_rob` API
 
Overall coverage up to 92%
image


client/hawc_client/summary.py
```diff
@@ -73,7 +73,7 @@ def get_visual(self, visual_id: int):
             visual_id (int): ID of the visual to read
 
         Returns:
-            dict: The result object, if get was successful
```
munnsmunns, line 76: unlike other `get_<model>` functions in the client, visual and datapivot return the resopnse object instead of a dict. Should I make it consistent and return the dict here? Seems like an oversight

shapiromatron: great idea - let's add a jira card and do next quarter, or after the testing work is complete!

tests/hawc/apps/hawc_admin/test_api.py
```diff
@@ -81,3 +84,24 @@ def test_assessment_values(self):
         assert resp.status_code == 200
         df = pd.read_json(StringIO(resp.content.decode()))
         assert df.shape == (3, 33)
+
+
+@pytest.mark.django_db
+class TestAdminActions:
+    def test_delete_orphan_tags(self, db_keys):
```
munnsmunns, line 91: Is there a quick way to produce an orphan tag? I tried just deleting a parent (e.g. Inclusion) but that seemed to clean up the children too

shapiromatron: you could probably make one in SQL using the django run sql command and set the depth to 2 and the path to something like 99990001.  You can add that as a todo and mark as xfail and we can fix later, i wouldn't spend a lot more time on it https://docs.pytest.org/en/8.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail

shapiromatron: I ended up removing this admin action since it wasn't used as much as I thought we needed it; and added a different test to check the clean_orphans implemented in `hawc.apps.common.models.AssessmentRootMixin`

shapiromatron (approved):
Nice! That 2% is a big deal when we have 30k LOC, thanks! Some of these tests looked really time consuming to write, glad they're here....

PR #1052, opened 2024-06-21
Remove  `summary.Visual.sort_order`

shapiromatron:
Remove the `summary.Visual.sort_order` database model field. This is a configuration for a single visualization type (the risk of bias heatmap), and therefore it was migrated to the settings JSON for that case. This simplifies the logic elsewhere in the app, since it shouldn't be defined globally on all Visuals.

In addition, the field didn't actually work :(. We fixed the implementation so now if you choose to sort by overall study judgment, it actually sorts as intended.  It defaults to alphabetical.

Sort by name alphabetically:
image

Sort by overall study judgment:
image


rabstejnek (approved):
Looks good!

PR #1053, opened 2024-06-21
add form helpers instead of repeating code

shapiromatron:
In a few recent code reviews I noticed some boilerplate code that sets the textarea height for different fields in django forms. I rewrote the current forms by adding a helper method `set_textarea_height` that sets the height for all textarea or a specified list. 

I also removed some code that set `form-control` or not to checkboxes; it didn't appear to have any impact, so I wonder if it was related to when we used twitter bootstrap 2 or 3 instead of 4, but please double-check to make sure I'm not missing something.


caseyhans (approved):
So many forms! LGTM, always nice to cut some lines of code out.

PR #1054, opened 2024-06-22
Exports v2

shapiromatron:
Convert all our tabular exports to build from ORM exports cast into pandas data frames instead of lists of nested JSON objects. This should improve export performance significantly in many cases, and should make the code simpler and easier to maintain in the future.

This work was originally started October 12, 2023 (260 days ago). Nice job @rabstejnek!!!

This original work was more of a 1:1 conversion; future work can be added to dramatically improve performance if we find any performance bottlenecks in the future.


rabstejnek (approved):
🎊 🎊 🎊

PR #1056, opened 2024-06-24
Eco updates from review

caseyhans:
- Update and add new nested terms
- Replace habitats


shapiromatron (approved):
LGTM - please double check @caseyhans! Here's the notebook I used to process the data and design the migrations: https://github.com/USEPA/cphea-assessments/blob/ajs/projects/hawc/eco/nested-terms-update.ipynb

caseyhans:
Works for me!

PR #1057, opened 2024-06-24
add dev-header to full width of body instead of as a tab

shapiromatron:
The tabbed development header was changed to a full width header at the top. This is useful for screenshots so it's possible to take a screenshot of the page without showing the development environment; it also adds a little more prominence to the tag.

Old:
image

New:
image


caseyhans (approved):
This is great! Never liked it in the screenshots.

PR #1058, opened 2024-06-25
exports-v2 - Updates from image review

shapiromatron:
I reviewed around 3000 visuals and found very minor differences in <1% of them. This PR addresses fixing the issues in that 1%, along with a few other bugs I found when investigating these differences. Here's a detailed list of changes I made:

- The data pivot html view previously didn't appropriately handle a 500 error on a dataset and treated it like a valid response; this is now fixed and it throws an error message correctly
- The animal `treatment_period` transformation method didn't appropriately transform text in a few cases; this was rewritten
- In some cases, if a rare case where the study evaluation metrics have duplicate names, an export would fail because you can not assign multiple columns (`df.loc[list_of_strings]`) if columns are not unique; a get unique name method was added, and the study evaluation fetching code was updated to be reused across exports
- If data is correct and upper and lower CI are not provided, epi can calculate upper and lower ci values. A new method was added to do so
- Fixed invitro category addition to a datafarme where an endpoint has None as a category ID
- Remove `iv_endpoint-category_id` from invitro data pivot exports

Unrelated change:

- Fix deprecation warning when running pytest (in conftest.py):
    - Marks applied to fixtures have no effect
- Add UDF API to the openapi yml specification


hawc/apps/animal/exports.py
```diff
@@ -688,21 +688,18 @@ def _func(group_df: pd.DataFrame) -> pd.Series:
...
-                result_type="reduce",
-            )
-        ).astype(str)
-        df["treatment period"] = (
-            txt_updated + " (" + df["dosing_regime-duration_exposure_text"]
-        ).where(df["dosing_regime-duration_exposure_text"].str.len() > 0) + ")"
+    def handle_treatment_period(self, df: pd.DataFrame) -> pd.DataFrame:
+        def _calc(row):
+            txt = row["experiment-type_display"].lower()
+            if txt.find("(") >= 0:  # TODO - remove extra space after confirming changes
```
rabstejnek, line 694: If this TODO was addressed we should remove it.

rabstejnek (approved):
Changes look good! I had just one comment about a TODO left in the code, after that is addressed I think its good to merge.

hawc/apps/animal/exports.py
shapiromatron: resolved in dcd0d0c8f1857de6c2de9eade1ff00cb6ac9be46

PR #1060, opened 2024-06-28
eco - add deprecated badge on search view

shapiromatron:
If a deprecated term is an ancestor or descendant of a valid term, add badge to show it's deprecated:

image


caseyhans (approved):
Good call! LGTM

PR #1061, opened 2024-06-28
Fix bugs from lit demo

caseyhans:
Fixes overflow literature reference IDs and UDF HTMX conditionals


shapiromatron (approved):
well done, works great!

PR #1062, opened 2024-06-28
add more tests

shapiromatron:
Using the new class and function level coverage reports, target a few more unit tests to test largely untested code:

- Test visibility of DSSTox list admin
- Test assessment attachment ViewSet methods
- Test admin dashboard functionality
- Test myuser model forms
- Test vocab term manager methods

In addition:

- Fix bug in the admin user form; previously it was not possible to save if email_verified was unset, and the text-based datetime widgets failed transforming on a POST
- Remove `get_tags_in_assessment` method; it was unused
- Improve the `get_client` method to optionally turn on htmx headers
- Skip coverage on all management methods; they're not available directly in the web application which is our primary focus of test coverage
- Update a calculation to add a parameter for days, otherwise the test would fail since all data in our text fixture is older than the default of 180 days


munnsmunns (approved):
Nice! Those were some tricky spots to test

PR #1064, opened 2024-07-01
UDF & Reference Filterset Change

caseyhans:
Two changes:
- Fixed a UDF binding list bug that occurred if the tag binding and model binding had the same ID
- Removed the `user_tags__is_resolved=False` query check for the reference filterset when filtering by "my tags" or "anything tagged me"; clicking "2 reviews done by you" on the overview page and seeing an empty list felt incorrect, and there are additional filtering options available (needs tagging) if users want to filter down further.


shapiromatron (approved):
LGTM - changes make sense where you added them. This does change how the filterset works in other places I think (like the tagging view), but it seems reasonable I think, or at least internally consistent.

PR #1066, opened 2024-07-03
CPHEAPM-26 - Add ToxRefDB vocab fixture and display

ZindahFarhaICF:
## Changes
- Added vocab fixture and migration
- Updated tests to include new vocab namespace and page: {baseurl}/vocab/toxref


frontend/animal/index.js
DookTibs, line 1: need to address the ci/cd formatting issues; want to make sure your dev environment is running such checks so we don't run into this

frontend/animal/ToxrefBrowser/App.js
DookTibs, line 1: in general ToxrefBrowser and the existing EhvBrowser are like 99% the same so there is a lot of copied code between the two (EhvBrowser/App.js and index.js are exactly equivalent to their new ToxrefBrowser counterparts for instance). Could we refactor these so we're not duplicating so much code?

scripts/load-toxref-vocab.py
```diff
@@ -0,0 +1,158 @@
+import json
+
+# import os
```
DookTibs, line 3: remove the commented out code, e.g. "# import os", "# os.environ.setdefault", etc. (obv good to keep actual explanatory comments like "# Generate the data fixture...")

frontend/animal/ToxrefBrowser/App.js
DookTibs: true in a couple other places too; hard to say if it's worth refactoring or not. Could also see if Andy would prefer it that way; like if they anticipate adding additional vocabularies over time it is worth the effort; if it is a real rarity than maybe it isn't.

frontend/animal/VocabBrowser/App.js
```diff
@@ -16,7 +17,13 @@ class App extends Component {
             query: "",
         };
         this.updateQuery = _.debounce(() => props.store.updateQuery(this.state.query), 100);
+
+        this.tables = {
```
ZindahFarhaICF, line 21: To reduce duplicate code, I added a more generic VocabBrowser folder that imports the correct term table based on the vocabulary invoked. To render the correct table, I had to explicitly define all possible options: I don't love this as a setup, but couldn't figure out how to avoid this, and am open to any changes.

frontend/animal/VocabBrowser/App.js
```diff
@@ -3,7 +3,8 @@ import {inject, observer} from "mobx-react";
 import PropTypes from "prop-types";
 import React, {Component} from "react";
 
-import Table from "./Table";
+import EhvTable from "./EhvBrowser/Table";
```
ZindahFarhaICF, line 6: As of now, the term tables are very similar. I went off of the assumption that tables may vary more widely in the future, and didn't want any logic to get muddy, so I'm passing in the full table instead of creating smaller custom components. Again, open to feedback!

hawc/apps/vocab/api.py
```diff
@@ -92,6 +93,18 @@ def related_entity(self, request: Request, pk: int | None = None) -> Response:
         return Response(serializer.data, status=status.HTTP_201_CREATED)
 
 
+class EhvTermViewSet(VocabTermViewSet):
```
ZindahFarhaICF, line 96: I had the option of invoking the parent, VocabTermViewSet, with arguments, but I created each subclass to make calling the vocabulary cleaner, which seems to be the method used in other parts of hawc.

hawc/apps/vocab/templates/vocab/toxref_browse.html
```diff
@@ -0,0 +1,21 @@
+{% extends 'crumbless.html' %}
+
+{% block content %}
+  <a class="pull-right btn btn-primary" href="{% url 'vocab:api:toxref-nested' %}?format=xlsx">Download</a>
+  <h2>ToxRef Database Vocabulary</h2>
+  <div>
+    <p>
+      Experimental data extracted into HAWC is designed to enable filtering, sorting, and
```
ZindahFarhaICF, line 8: I left this template in case we're changing the text for the toxref vocab: if the text is consistent for all vocabularies, further refactoring is possible

frontend/animal/VocabBrowser/App.js
ZindahFarhaICF: I also just implemented a version of this which uses dynamic loading (to import the table I need instead of every table). To update a vocab in this version, you would still need to update a new "Tables.js" file to include the new vocab import. If that might be a better way, I can commit it to this PR as well.

DookTibs: I think this looks better yea. I personally like the dynamic loading idea but @shapiromatron I believe I've submitted code that does that in the past and after feedback went back to doing it explicitly (this may have been in LLR not HAWC now that I think about it). Any preference?

shapiromatron: sure, this seems fine!  I generally like things more explicit if there are securiy implications, but client side, seems fine.  The only thing I don't like about this pattern is it doesn't really handle the case where the vocab doesn't match, you'd get a silent undefined. maybe something like this?

```js

const tables = {ehv: EhvTable, toxref: ToxrefTable},
  Table = tables[this.props.store.config.vocab];

return (
  ...
  {_.isUndefined(Table)?<p>ERROR - unknown vocabulary</p>:<Table/>}
  ...
)
```

ZindahFarhaICF:
Hi @shapiromatron, marking this as ready for review.

frontend/animal/VocabBrowser/App.js
```diff
@@ -0,0 +1,50 @@
...
+        this.state = {
+            query: "",
+        };
+        this.updateQuery = _.debounce(() => props.store.updateQuery(this.state.query), 100);
+    }
+
+    render() {
+        const tables = {ehv: EhvTable, toxrefdb: ToxRefDBTable},
+            Table = tables[this.props.store.config.vocab];
+        if (_.isUndefined(Table)) {
```
shapiromatron, line 26: I changed this to guard (return w/ a different condition if we're not on the happy path)

hawc/apps/vocab/api.py
```diff
@@ -12,15 +13,17 @@
 from . import constants, models, serializers
 
 
-class EhvTermViewSet(viewsets.GenericViewSet):
+class VocabTermViewSet(viewsets.GenericViewSet):
```
shapiromatron, line 16: nice job generalizing this. With your initial approach, the data was loaded every time the server restarted, which is all the time in development mode, since it was a class-level property on the class. I changed it to be a callable method, so it's only fetched when needed. I did the same for the view too.

shapiromatron (approved):
Fantastic job! I made some edits, but nothing too major.  See my individual comments and then the diffs for the commits below; I tried to separate independent changes so you can hopefully follow along...

* 7754a217 cleanup abstract class (9 seconds ago)
* b6c44ddd switch to guard check to return bad case, instead of check in component (20 minutes ago)
* f8cf5bbf a few updates from testing (23 minutes ago)
* b9b742bd remove script (move to another repo) (29 minutes ago)
* 89f0cff6 rename Toxref to ToxRefDB (2 hours ago)
* 238ca3c3 update migrations to be consistent with current main (2 hours ago)
* 868d8ac0 lazily generate large vocab dataframes instead of on app startup (2 hours ago)
* f95099d5 Merge branch 'main' into ICF-CPHEAPM26-toxref-vocab (4 hours ago)

PR #1070, opened 2024-07-12
Add path for new HERO API in HERO-ID based import

munnsmunns:
Adds a new path in the HEROFetch class that uses the new HERO api. To use the new path, set the feature flag `HAWC_FEATURES.ENABLE_NEW_HERO = True` and set `HERO_API_KEY` to a valid api key from your HERO profile page.


hawc/services/epa/hero.py
```diff
@@ -29,20 +29,37 @@ def _force_int(val, default=None) -> int | None:
...
+        return dict(
+            json=content,
+            HEROID=_force_int(content.get("id")),
+            PMID=_force_int(content.get("accession_number")),
+            doi=try_get_doi(content.get("doi", "")),
+            title=content.get("title"),
+            abstract=content.get("abstract"),
+            source=content.get(
+                "type_of_reference"
+            ),  # TODO: need to construct source from journal/issue/pages/etc.
```
munnsmunns, line 43: The new API doesn't return the source like the old one does, but it does return the journal title/issue #/pages separately. However I don't see anywhere that we actually use the source from HEROFetch, should I just leave it like this?

shapiromatron: thanks for the heads up; I'll take a look when I do the detailed review. Is this one complete from your end?

hawc/services/epa/hero.py
```diff
@@ -90,24 +107,53 @@ def get_content(self):
...
-                r = requests.get(url, timeout=30.0)
-                if r.status_code == 200:
-                    data = json.loads(r.text)
-                    for ref in data["results"]:
-                        self.content.append(parse_article(ref))
-                else:
+            results = []
+            if settings.HAWC_FEATURES.ENABLE_NEW_HERO:
+                url = "https://heronetnext.epa.gov/api/reference/export/json"
+                params = {"id": request_ids, "type": "hero"}
```
shapiromatron, line 130: switch to post for this b/c the request_ids could be long in a URL as url params

hawc/services/epa/hero.py
```diff
@@ -129,7 +129,7 @@ def get_content(self):
                 url = "https://heronetnext.epa.gov/api/reference/export/json"
                 params = {"id": request_ids, "type": "hero"}
                 try:
-                    r = requests.get(url, params, headers=headers, timeout=30.0)
+                    r = requests.post(url, params, headers=headers, timeout=30.0)
```
shapiromatron, line 132: Does the cassette need to be regenerated?

hawc/services/epa/hero.py
```diff
@@ -90,24 +124,47 @@ def get_content(self):
             self.failures = []
             return dict(success=self.content, failure=self.failures)
 
+        if settings.HAWC_FEATURES.ENABLE_NEW_HERO:
```
shapiromatron, line 127: I removed the key validation check if it fails, we'll get an error, but we send fewer requests to HERO

hawc/services/epa/hero.py
```diff
@@ -90,24 +124,47 @@ def get_content(self):
...
+                        results = data["results"]
+                    else:
+                        logger.info(f"HERO request failure: {url}")
+                except requests.exceptions.Timeout:
+                    logger.info(f"HERO request timeout: {url}")
+                except json.JSONDecodeError:
+                    logger.info(f"HERO request failure: {url}")
+
+            for ref in results:
+                self.content.append(parse_func(ref))
```
shapiromatron, line 167: instead of adding an if statement to determine which parser to use, we select a `parse_func` and then use it everywhere. I just made a new function w/ our custom logic.

hawc/services/epa/hero.py
```diff
@@ -28,6 +28,40 @@ def _force_int(val, default=None) -> int | None:
         return default
 
 
+def format_source(content: dict) -> str:
```
shapiromatron, line 31: I added this func to address to TODO you had, since the old HERO API previously gave us a "{journal} {volume}:{pages}" like source string but the new one doesn't, so we can selectively construct it here.  I added a test, I figure we can expand that test overtime if there are some edge cases that aren't handled well by the current implementation.

shapiromatron (approved):
LGTM - please merge if you're ok with my edits!

PR #1074, opened 2024-07-24
Alert tag rewrite

caseyhans:
Add args + kwargs functionality to the alert template tag, and use the tag where possible.


rabstejnek (changes requested):
Have you or could you look into making the alert template tag more like `with` or the like? ie where you can construct it like `{% alert type="danger" %}` instead of `{% alert "type=danger" %}`, just for consistency's sake.
https://github.com/django/django/blob/main/django/template/defaulttags.py#L1509
In that link you will also see there are token parsers provided by django, so we may be able to cut down on some code as well.

rabstejnek (approved):

PR #1076, opened 2024-08-02
DRAFT PR: refining a search when a similar search already exist - ref…

matt-PiR2:
Draft code to finalize.

### DIFF
--- a/hawc/apps/lit/models.py
+++ b/hawc/apps/lit/models.py
@@ -402,31 +402,61 @@ class Search(models.Model):
         3. do not have any tags applied
         4. do not have any studies
         """
-        if results["removed"]:
-            ids = [str(id) for id in results["removed"]]
-            # filter removed references with no tags
-            no_tags = list(
-                self.references.filter(identifiers__unique_id__in=ids)
-                .annotate(ntags=models.Count("tags"))
-                .filter(ntags=0)
-                .values_list("id", flat=True)
-            )
-            # filter untagged references with only one search (this one)
-            no_searches = (
-                Reference.objects.filter(id__in=no_tags)
-                .annotate(nsearches=models.Count("searches"))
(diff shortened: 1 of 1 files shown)


matt-PiR2:
The code I though to fix this issue is as follow :

-------------------------------------------------------------
```
if  (date of the last search query [is before] date of applying the seach algorithm fix):
          (apply a different search query filter to fix the old search query, my suggestion is to make it as if we are starting a brand new search here, using the seach algorithm fix.)
else:
          (use the seach algorithm fix which consist of using a rolling filter)
```

hawc/apps/lit/models.py
```diff
@@ -402,31 +402,61 @@ def delete_old_references(self, results):
...
-            _ids = no_searches.values_list("id", flat=True)
-            Study = apps.get_model("study", "Study")
-            no_studies = no_searches.exclude(id__in=Study.objects.filter(id__in=_ids))
-
-            # remove candidate deletions
-            n = no_studies.count()
-            if n > 0:
-                logger.info(f"Removing {n} references from search {self.id}")
-                no_studies.delete()
+        if #(date of the last search query [is before] date of applying the seach algorithm fix):
```
shapiromatron, line 405: I'm a little confused - the code looks identical in the if and else statements for this pull request - what's different in the two approaches?

From the comment, it looks like you're saying we need a different search query filter - what would that new search query filter be?

shapiromatron: @matt-PiR2 - I sent an email last week with a possible idea that made this an option for the user at the form level instead of something hard-coded based on date - how did that one sound? If that approach works, can you try to implement, or would you need help from my team?

shapiromatron:
Closing; chatted with @matt-PiR2 over email and there's not a clear path forward on what is desired, so we will make no change for now.

PR #1078, opened 2024-08-13
Assessment tag schema & CRUD

rabstejnek:


hawc/apps/common/models.py
```diff
@@ -277,8 +277,11 @@ def create_tag(cls, assessment_id, parent_id=None, **kwargs):
         else:
             parent = cls.get_assessment_root(assessment_id)
 
+        # TODO make this a regex check to prevent adding a name that blocks another assessment
```
shapiromatron, line 280: is that needed? we could just update the queryset to only check for a depth of 1, which users couldn't edit I think

hawc/apps/assessment/forms.py
```diff
@@ -667,3 +667,66 @@ class Meta:
...
+        else:
+            default_choice = (self._queryset[0].pk, "-----")
+            self.choices = [
+                default_choice,
+                *[(obj.pk, obj.nested_name) for obj in self._queryset[1:]],
+            ]
+
+
+class TagForm(forms.ModelForm):
+    # TODO make htmx not allow multiple open updates,
```
shapiromatron, line 694: make this a TODO for future work? Would this happen frequently, it seems like a possible rare edge case

hawc/apps/common/models.py
rabstejnek: Good call! I included that in the newest PR

rabstejnek:
Closing this in favor of #1085

PR #1080, opened 2024-08-14
Update Ruff Rules

caseyhans:
Update code base to accept previously ignored ruff rules.
- Return errors when caught and raised
- Unused iterable variable names prefixed with underscores
- Add a 15 second timeout to any GET and POST requests 

Notes on other ignored rules:
S308: using mark_safe
	- we use this a lot in template tags
	- HERO uses a custom HTML sanitizer that calls mark_safe, which is then flagged as safe inline
	- We could use this approach, or flag all our mark_safes inline

S314: raised when parsing XML input; I believe it'd require a new package to solve (defusedxml)
	- 4 cases; could flag inline & remove from ignore list
	- 3 cases are from parsing pubmed imports, one other is in a lit migration (0012) that parses HERO/RIS

S603: executing subprocess calls on "untrusted" inputs
	- 4 cases; could be flagged as safe inline
	- two of these cases are in the assessment_db_dump.py command for getting the db name from settings
	- two of these are in the git.py commit model that returns last commit info; runs "git log" and grabs output

RUF012: 825 errors. too many to fix. related to mutable class attribute annotations

RUF015: iteration check; 3 failing cases, could be flagged safe inline

### DIFF
--- a/client/hawc_client/interactive.py
+++ b/client/hawc_client/interactive.py
@@ -49,8 +49,8 @@ async def get_svg_image(page, timeout_ms) -> BytesIO:
 
     try:
         await expect(download_button.or_(error_alert)).to_be_visible(timeout=timeout_ms)
-    except AssertionError:
-        raise TimeoutError(f"Timeout - exceeded {timeout_ms}ms") from None
+    except AssertionError as err:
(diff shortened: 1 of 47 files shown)


munnsmunns (approved):
Looks good to merge, nice job!

PR #1082, opened 2024-08-14
Study filter for exports

munnsmunns:
Add a filter for the animal and epiv2 full exports to limit results to specific study ids using the `study_ids` param.

### DIFF
--- a/hawc/apps/animal/api.py
+++ b/hawc/apps/animal/api.py
@@ -29,10 +29,10 @@ class AnimalAssessmentViewSet(BaseAssessmentViewSet):
 
     def get_endpoint_queryset(self, request):
         published_only = get_published_only(self.assessment, request)
-        study_ids = [int(study_id) for study_id in self.request.query_params.get("study_id", [])]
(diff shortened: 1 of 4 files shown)


caseyhans (approved):

hawc/apps/animal/api.py
```diff
@@ -29,7 +29,11 @@ class AnimalAssessmentViewSet(BaseAssessmentViewSet):
 
     def get_endpoint_queryset(self, request):
         published_only = get_published_only(self.assessment, request)
-        return models.Endpoint.objects.get_qs(self.assessment).published_only(published_only)
+        study_ids = [int(study_id) for study_id in self.request.query_params.get("study_ids", [])]
```
shapiromatron: this would fail with a 500 error if the value couldn't be cast to an int. I changed to: 

`study_ids = try_parse_list_ints(request.query_params.get("study_ids"))`

shapiromatron (approved):
See my change below - I modified so it could support "bad" requests

PR #1083, opened 2024-08-14
Upgrade django to 5.1

munnsmunns:
Update Django from 5.0 to 5.1


rabstejnek (approved):
I looked at the list of known backwards compatibility issues and tested out the new version locally and I didn't see any problems!
https://docs.djangoproject.com/en/5.1/releases/5.1/#backwards-incompatible-5-1

PR #1086, opened 2024-08-21
Remove Executive Summary feature

munnsmunns:
Removes the unused SummaryText model and all related frontend and backend code.


rabstejnek (approved):
I took a pass to see if I could find any remnants but I couldn't; looks good!

shapiromatron (approved):
nice job! sorry to see this go but since no one was using it, it's less code to maintain!

PR #1088, opened 2024-08-27
Catch FieldErrors and ValidationErrors in autocomplete

munnsmunns:
Prevents autocomplete from causing 500 errors when users send a bad request.


caseyhans (approved):

shapiromatron (approved):
nice job, good tests!

PR #1089, opened 2024-08-28
Fix missing popover on assessment details page

munnsmunns:
Fixes an issue where Confidence popover would not appear if ADAF was not checked for an assesment value.


shapiromatron:
I think I already fixed it in https://github.com/shapiromatron/hawc/pull/1077

PR #1090, opened 2024-08-30
Add exports for Prisma plot

munnsmunns:
Add Prisma data exports to visual form view, if Prisma is the selected visual type.


shapiromatron (approved):
LGTM; I made a few edits and fixed an error which would have broken all visual create forms that didn't have prisma data (there was a dangling comma in the JSON)

* 47d138bae add to update page in addition to create page
* ef29d587f model logic to model instead of view; add more metadata
* 3434de3af fix dangling comma for non-prisma visuals

PR #1093, opened 2024-09-04
Make AssessmentValue.study M2M

munnsmunns:
Changes `study` foreign key field to be a many-to-many relationship instead. Also modifies the help text for `value` and `pod_value` to make it clearer that entering scientific notation is valid.


hawc/apps/assessment/exports.py
```diff
@@ -103,6 +103,6 @@ def build_modules(self) -> list[ModelExport]:
             AssessmentDetailExport("assessment_detail", "assessment__details"),
             AssessmentValueExport("assessment_value", ""),
             StudyExport(
-                "study", "study", include=("id", "short_citation", "hero_id", "pubmed_id", "doi")
+                "study", "studies", include=("id", "short_citation", "hero_id", "pubmed_id", "doi")
```
shapiromatron, line 106: does this work if you have 2 studies for an assessment value?

hawc/apps/assessment/exports.py
```diff
@@ -39,6 +39,7 @@ def get_annotation_map(self, query_prefix: str) -> dict:
             "uncertainty_display": sql_display(
                 query_prefix + "uncertainty", constants.UncertaintyChoices
             ),
+            "studies__short_citation": str_m2m(query_prefix + "studies__short_citation"),
```
munnsmunns, line 42: is this enough study information for this export? It's tricky I think to add the identifiers like pmid because it's multiple m2m relations. I also tried to add the hawc ID but str_m2m doesn't work on the ID field because it's an integer

hawc/apps/assessment/migrations/0041_assessmentvalue_studies_and_more.py
```diff
@@ -0,0 +1,39 @@
+# Generated by Django 5.1 on 2024-09-04 18:41
```
rabstejnek, line 1: Would it be too much trouble to merge the other migration files into this one? Just to cut down on the number we're introducing.

rabstejnek (commented):
Looks good! I just had one comment about merging the migrations introduced in this PR.

Also, I noticed a comment that was marked as resolved regarding some of our m2m helper methods. While we normally pass them a field string, we can also pass them expressions, so we may be able to build each type of identifier (ie pmids, hero ids, etc) if we use a subquery. Just food for thought if we plan on implementing it.

hawc/apps/assessment/migrations/0041_assessmentvalue_studies_and_more.py
shapiromatron: Good find Matt. I did some digging and that's confirmed:

https://docs.djangoproject.com/en/5.1/ref/migration-operations/#runpython

> On databases that do support DDL transactions (SQLite and PostgreSQL), RunPython operations do not have any transactions automatically added besides the transactions created for each migration. Thus, on PostgreSQL, for example, you should avoid combining schema changes and RunPython operations in the same migration or you may hit errors like OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events.

https://stackoverflow.com/a/66799446/906385

shapiromatron (approved):
Looks good. There were larger two changes I had to fix during testing

1) The study autocomplete showed ALL studies across all assessments, not just those for one particular assessment.
2) The migration failed if a study is None and and it was trying to write a null study in the studies many to many relation

Diffs are here: 

* 8b2342740 reduce queries on detail view
* 72bbe60ac fix autocomplete; it previously didn't filter to only show studies in assessment
* ed6d346cb don't build a helper method for this one
* 4acac5bf4 rename migrations; don't add studies when study is None

PR #1094, opened 2024-09-05
remove unused scripts

shapiromatron:
Remove old scripts that haven't been run in ages and may not work or be relevant.


munnsmunns (approved):
Looks good!

PR #1095, opened 2024-09-05
make all hawc imports relative instead of absolute

shapiromatron:
Within the hawc package, make all imports relative instead of absolute


munnsmunns (approved):
Looks good! I'll make sure to do that with future imports

PR #1096, opened 2024-09-05
add a user agent for external requests

shapiromatron:
Add a `User-Agent` for all external http requests.  This makes it easier for 3rd party tools to identify our software; the user agent can be configured by changing the `HAWC_USER_AGENT` environment variable. 

In addition, standardize how we make external requests using a custom request session.

Finally, refactor the ancient `pubmed.py` file slightly to make the logic a little bit easier to follow. It's still messy, but we get the pubmed API key directly from the django config settings instead of an old singleton pattern (I think I just learned about that design pattern in school when I wrote this 🤣 ).

I regenerated all the VCR cassettes to confirm that everything did still work as expected.


munnsmunns (approved):
LGTM! Nice to see the Pubmed bit refactored as well.

PR #1098, opened 2024-09-05
Long reference export

shapiromatron:
Add a new option to download the literature reference export in long format instead of wide format.


munnsmunns (approved):
Works great Andy, looks good to merge!

caseyhans (approved):
LGTM!

PR #1099, opened 2024-09-06
Animal export optimization

rabstejnek:
This PR addresses some performance issues found in the full animal export. Some profiling revealed that there were two major bottlenecks for larger assessments:
- Dosage
  - Originally we used a `groupby` call to add the necessary dose unit columns and relevant doses. However, the calls made within the `groupby` had some overhead, so it was taking a long time on larger assessments.
  - The fix was to use a `pivot` call instead; this created a pivot table with all of the dose units as columns and doses as values, which we then just merged back into the dataframe.
- HTML cleaning
  - HTML cleaning was performed on multiple fields of multiple models. However, this was done using an `apply` call, so essentially it was an operation iterated over every record in the dataframe, which takes time.
  - The fix was to use a vectorized version of HTML cleaning, which sped this export up dramatically. This has the added benefit of speeding up other exports, since it was applied at the model level.

Below is a table of execution times of the full animal export on the top 10 largest assessments in terms of bioassay endpoints; there were no data differences noted after the changes. However, note that one assessment resulted in an execution time of `None` due to a `MemoryError`, which may just be a limitation of my resources when running it locally. This happened both before and after the updates, so it should be looked into separately from this PR.

Export time in seconds vs # endpoints with new approach (around 30 seconds for 10k endpoints):
8287c578-ff80-430d-adde-e8110001c562

Export time in seconds comparison main vs this PR:
1e2a53d9-8575-415e-b5a8-a8236abda748


caseyhans (approved):

shapiromatron (approved):
LGTM, but I had to some reworking in my PR after some extensive testing; the good news is I think I got the speed much faster, around 30s for an export w/ 10k endpoints.

The major change was not fetching any dose-group information in our initial query, and fetching that with a second query instead.  The dramatically reduced the # duplicative rows on the initial query returned, which increased the memory footprint and computation time.

@rabstejnek - please review my code and if you're with the changes, ok to merge!

rabstejnek:
Good changes! We should also move the dose group changes over to the other animal exports, but that can happen in another PR.

PR #1100, opened 2024-09-06
Add form prefixes to HTMX forms

caseyhans:
Our HTMX views allow for multiple identical forms to be rendered; this can cause some annoying bugs (notably with checkboxes).

This PR adds a [prefix](https://docs.djangoproject.com/en/5.1/ref/forms/api/#prefixes-for-forms) to the form instances based off the object's ID, which addresses the issue for the most part.

The bug is still present if a user creates multiple new forms (i.e. hits 'Create New' multiple times--since there are no model instances, these forms are again identical). A more complete solution requires refactoring our htmx more broadly; probably by only allowing one blank 'create' form at a time.


hawc/apps/assessment/forms.py
```diff
@@ -341,7 +341,8 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         obj = kwargs.pop("parent", None)
-        super().__init__(*args, **kwargs)
+        prefix = f"attachment-{kwargs.get("instance").pk if "instance" in kwargs else "new"}"
```
munnsmunns, line 344: Maybe add a random string or number to "new" as a quick fix here?

munnsmunns (commented):
Looks great! I had an idea for fixing the issue with multiple new objects. I don't think it would be too much work if you decide to go ahead with it.

hawc/apps/assessment/forms.py
caseyhans: This would require rewriting all the create tests to "get" the form, isolate the random number, and insert it into the data before POSTing. Since it's not an instant quick fix, I think tweaking our htmx viewsets next quarter is a better solution

munnsmunns (approved):

shapiromatron (approved):
Looks good. I had to figure out a different pattern for dynamic show/hide forms. We could use alpine.js but that's too big for this PR. 

* d023b534e fix dynamic JS using the data attribute instead of the name

PR #1104, opened 2024-09-11
Add UDF to animal groups and datapivot exports

caseyhans:
Add UDFs to animal groups, and add model UDF content from studies, endpoints, and animal groups to endpoint/endpoint group data pivot exports

- Need to get display names for choice fields


rabstejnek (approved):
Looks good!

shapiromatron (approved):
Nice job. I added a little more complexity to how we rename UDF columns, just to make sure we don't overwrite existing column names.

* 01a6a52b7 selectively rename and add test
* e48b7ab79 rename generic relations
* 0780a80a9 order models

PR #1109, opened 2024-09-23
Change pandas pattern for empty dataframe

caseyhans:
Change df.shape[0] == 0 and df.shape[0] > 0 checks to use df.empty


shapiromatron (approved):
looks good!

PR #1110, opened 2024-09-24
ADAF updates for Assessment values

munnsmunns:
- Changes the text and meaning of the ADAF footnote for assessment values. Now, checking the ADAF checkbox indicates that the adjustment has been applied to the value.
image

- Adds a field called Value Type Qualifier, which adds qualifying text to the value type (e.g. Adult  OSF or Adult Overall RfD)
image
image


shapiromatron:
@munnsmunns I think the other part of the jira task was to add the "Value Type Qualifier" - will that be done in this or a separate request

shapiromatron (approved):
Looks great. Made some minor name and text revisions:

* 8f4650c57 add to db fixture
* 2968b41a7 Update models.py
* 9fa2ebf5f naming revisions
* d228a0b2f rename migration

PR #1111, opened 2024-09-25
Fix for conflict between SET url setting and url validation

rabstejnek:
This fix adds a trailing forward-slash to the short citation links that SET tables create by default for that column, so that if a user ever customized them they won't fail validation due to a missing trailing slash.


shapiromatron (approved):
good find!

PR #1113, opened 2024-09-28
multiple worksheet epiv2 export

shapiromatron:
Add new tabular export for epiv2 where there is one Excel tab (worksheet) per database table. While a more complicated export, it will allow for more flexibility in building data downstream:

image

The export is Excel-only.

image


caseyhans (approved):
Added the study internal communications if unpublished studies are included.

PR #1115, opened 2024-10-01
Bulk tagging JS fix

rabstejnek:
If the IDs in an excel file were parsed as strings, then the bulk tagging form would not be able to map them to integer based IDs. This PR addresses this issue by casting all incoming excel IDs explicitly as strings and the keys in the reference map also as strings; since these "stringified" values are only used for matching and because some IDs (ie doi) are already strings this seemed like the most unified approach.

Whitespace is also trimmed from incoming excel IDs, since identifiers stored in HAWC should not have leading or trailing whitespace; this also aids in a better match between excel IDs and the reference map.


shapiromatron (approved):
Nice job, a simple fix that should work well.

PR #1116, opened 2024-10-03
update epiv2 study design visual fields

caseyhans:
Update the epiv2 study summary visual heatmap options


shapiromatron (approved):
LGTM - I had to add a few more things to make it so it was one row per study design, but it works well.

PR #1117, opened 2024-10-09
Icf cpheapm70 deep clone study, study evaluation, bioassay, epiv2

BerkowitzMaxICF:
Add the ability for a user to clone a study, including study evaluation and animal bioassay information, from one assessment to another assessment. This will do a deep clone that copies all related information.

The new page is listed under the Study List page:
image

You select the source assessment to copy from, and then which studies to clone. If study evaluation is selected, you must also map metrics from the from destination to source:
image


hawc/apps/study/views.py
```diff
@@ -6,6 +6,8 @@
 from django.urls import reverse_lazy
 from django.views.generic import RedirectView
```
shapiromatron, line 8: use relative not absolute imports

hawc/apps/study/forms.py
```diff
@@ -221,6 +221,46 @@ def save(self):
...
+            help_text=f"Copy {rob_name} metrics and domains from an existing HAWC assessment which you have access to.",
+            cancel_url=reverse("riskofbias:arob_update", args=(self.assessment.id,)),
+            submit_text="Copy from assessment",
+        )
+        helper.layout.insert(3, cfl.Div(css_id="approach"))
+        helper.layout.insert(2, cfl.Div(css_id="extra_content_insertion"))
+        return helper
+    """
+
+    def evaluate(self):
```
shapiromatron, line 259: I'd put the code here in `hawc.apps.study.actions.clone` or something like that, and actions can be a folder with an empty `__init__.py` file.  We can wrap a `@transaction.atomic` around this method perhaps.

hawc/apps/study/views.py
```diff
@@ -116,6 +118,43 @@ def get_context_data(self, **kwargs):
...
+    # parent_template_name = "assessment"
+    # parent_model = Assessment
+    model = Assessment
+    form_class = forms.StudyCloneForm
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        context["study_data"] = {}
+        assessments = Assessment.objects.all().user_can_view(
```
shapiromatron, line 136: This list will be massive in many cases, suggest making it smaller by adding a src assessment and only showing references from the src assessment.

hawc/apps/study/templates/study/study_clone.html
```diff
@@ -0,0 +1,62 @@
...
+                <input disabled name="rob" type="checkbox" value="{{study.rob}}">
+              {% endif %}
+            </td>
+          </tr>
+        {% endfor %}
+      </tbody>
+    </table>
+  {% endfor %}
+  <br>
+  <h4>Map metrics*</h4>
```
shapiromatron, line 45: to map the metrics, I think you'll need to select the src assessment at a prior stage, we dont want to map the metrics for all possible assessments here. It may make more sense to make two pages (or an htmx command) where you'd select your src assessment first, and then this populates with just the info for the src assessment.

For now, you could just update the url to do something like `assessment/<int:pk>/clone-from/<int:src_pk>/` and then if someone visited `assessment/1/clone-from/2/`, the src would be assessment 2 and dst would be assessment 1.

shapiromatron (commented):
Looking good so far, I added a few comments...

shapiromatron (approved):
Great start! I made a number of changes to address lots of edge cases, and hopefully improved error messaging to users when things didn't behave as expected. The initial work was a good template for reworking.

shapiromatron:
Adding @rabstejnek to the review; I made so many changes I think it needs another set of eyes.

hawc/apps/study/models.py
```diff
@@ -179,46 +178,13 @@ def get_json(self, json_encode=True):
     def get_attachments_dict(self) -> list[dict]:
         return [attachment.get_dict() for attachment in self.attachments.all()]
 
-    def get_bioassay_endpoints(self):
```
rabstejnek, line 182: Nice finds in this file!

rabstejnek (changes requested):
Looks good! Small changes requested, or just some clarification in some cases.

hawc/apps/study/actions/clone.py
```diff
@@ -0,0 +1,349 @@
...
+
+    # copy identifiers
+    dst_study.identifiers.set(identifiers)
+
+    return study_map, dst_study
+
+
+def clone_animal_bioassay(src_study: Study, dst_study: Study) -> StudyAppMapping:
+    animal_map = defaultdict(dict)
+    experiments = list(src_study.experiments.all().order_by("id"))
```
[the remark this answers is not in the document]
shapiromatron: We also need the order by because in some cases there parent/child relations in the same database table, so if we go by the order that they were created we should be able to resolve all the parent child relations (the AnimalGroup generational model)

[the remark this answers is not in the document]
shapiromatron: But I will double check for consistency as well, thank you!

hawc/apps/study/actions/clone.py
```diff
@@ -0,0 +1,349 @@
...
+    def validate_after(self):
+        if self.include_rob is False and len(self.study_rob) > 0:
+            raise ValueError("Cannot include RoB without a study selected for RoB")
+        elif self.include_rob and (len(self.metric_map) == 0):
+            raise ValueError("Cannot include RoB without a RoB mapping")
+        elif self.include_rob and (self.copy_mode is None):
+            raise ValueError("Cannot include RoB without a copy mode specified")
+        return self
+
+    def clone(self, user, context: dict) -> dict[Study, tuple[Study, dict]]:
```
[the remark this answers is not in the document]
shapiromatron: It was wrapped in a transaction at the view level, but I agree, I moved the decorator to this function to make it clearer it's being used (see 8cc38c2c1c61e9818171dc0c09ace51c12e9dcb6).   I 

I also made the other functions "private" w/ an underscore prefix, since they're not wrapped in transactions; I never really intended them to be used outside of testing in other systems so this was a good suggestion to clarify (see 8c78023eaf92f08717c4b1d0d8258d0fa6f30ad9).

rabstejnek (approved):

PR #1118, opened 2024-10-11
Icf cpheapm71 72 add guidelines

ZindahFarhaICF:
71
- Upload guideline table as json in the vocab app 'fixtures' folder.
- Update the experiment model to have a "guideline profile" (charfield) with validation from above guidelines.
72
- Create a GuidelineProfiles model and fixture with current hawc toxref endpoints


hawc/apps/vocab/constants.py
```diff
@@ -70,3 +70,15 @@ class Ontology(IntegerChoices):
...
+class ObservationStatus(TextChoices):
+    """
+    Guideline profile observation status
+    """
+
+    NM = "NM", "NM"
+    NR = "not required", "not required"
+    REC = "recommended", "recommended"
+    REQ = "required", "required"
+    TR = "triggered", "triggered"
```
ZindahFarhaICF, line 84: Added type_checking for GuidelineProfile observation status, can be changed/removed as needed

scripts/client/load_guideline_profile.ipynb
```diff
@@ -0,0 +1,177 @@
...
+    "    endpoint_target.name AS endpoint_target\n",
+    "FROM\n",
+    "    public.vocab_term AS endpoint_target\n",
+    "LEFT JOIN\n",
+    "    public.vocab_term AS endpoint_type ON endpoint_target.parent_id = endpoint_type.id\n",
+    "LEFT JOIN\n",
+    "    public.vocab_term AS endpoint_category ON endpoint_type.parent_id = endpoint_category.id\n",
+    "WHERE\n",
+    "    endpoint_target.type = 4 AND endpoint_target.namespace=2;\n",
+    "\"\"\""
```
ZindahFarhaICF, line 50: This notebook will probably be moved - mainly to demonstrate how guideline profiles were loaded (for review)

To load the model, I updated the endpoint ids in the given guideline_profiles spreadsheet to be the "endpoint target" (effect_subtype) id in the equivalent hawc vocab, and included the SQL logic for getting that here. Hopefully I didn't misunderstand the endpoints/goal here, but def let me know if I did!

hawc/apps/vocab/migrations/0009_guideline_profile.py
```diff
@@ -0,0 +1,54 @@
...
+                (
+                    "endpoint",
+                    models.ForeignKey(
+                        blank=True,
+                        null=True,
+                        on_delete=django.db.models.deletion.CASCADE,
+                        to="vocab.term",
+                    ),
+                ),
+                ("endpoint_category", models.CharField(max_length=256)),
```
ZindahFarhaICF, line 32: Guideline_profile_id is the primary key, and I kept the 3 vocab name fields that make up the endpoint. If they were mainly for my benefit to load in the vocab ids, I can remove them!

shapiromatron (changes requested):
This looks good generally, unfortunately I think we wanted to add it to the bioassay-v2 schema, not the bioassy-v1 schema.

To do that, I think this PR may need to be closed and reworked and then submitted as new PR that should be merged into the `bioassay-v2` branch, not the main branch.

I think a lot of the code in the `hawc.apps.animal` will need to be modified and adapted to `hawc.apps.animalv2`. Hopefully that won't be too painful?

scripts/client/load_guideline_profile.ipynb
shapiromatron, line 1: Is this script needed? What's the purpose of it, and when would we call it?

PR #1119, opened 2024-10-16
Refactor HTMX Implementation to avoid duplicate create forms

caseyhans:
Refactors our HTMX views to avoid having multiple create forms at once, since this introduced bugginess.

The general pattern is this (refer to lit workflows as a good example):

- In the template where you want the create form to exist, add a div with the classes `create-row hidden`
- On the create button, add `hx-target=".create-row" hx-swap="outerHTML"`
- In the edit form template, add the class `create-row` if this is a create and not an update (so that duplicate button presses will replace this item). Don't delete the form if a user hits cancel, just hide.


rabstejnek (changes requested):
There were a few bugs I witnessed when testing this out:
- The cancel button doesn't work on label create forms in the HAWC Prime flavor.

Other optional changes:
- I noticed usage of classes for the create row htmx trigger. In practice this should only be a single element, so I'd personally prefer using an id, but it doesn't really make a difference with functionality. This change may be a bit branching so I'd be fine if it wasn't addressed.

*Edited this comment after some bugs turned out to be local.*

caseyhans:
@rabstejnek I can't recreate any of these bugs. The first two are css-related, so a hard refresh might fix it. Not sure about the last bug. I'll look into using IDs for create rows, that's a good call

caseyhans:
Ok, I tried swapping the `create-row` htmx target to IDs instead of classes, but it was introducing some weird bugginess to the autoscroll script that I couldn't pin down. Since it's not really a functional change anyways, I think we're OK to leave it as a class. If the change is worth the time, I keep working on it.

shapiromatron:
> Ok, I tried swapping the `create-row` htmx target to IDs instead of classes, but it was introducing some weird bugginess to the autoscroll script that I couldn't pin down. Since it's not really a functional change anyways, I think we're OK to leave it as a class. If the change is worth the time, I can keep working on it.

Seems reasonable to me, I think this looks like it's ready for my review. Thanks yall!!

rabstejnek (approved):

shapiromatron (approved):
Functionally, it works great, and fixes the bugs I introduced with my alternative approach. Since the selectors on pages with multiple creation forms use a dual selector, `.foo.bar`, I think this should work fine in practice. https://stackoverflow.com/a/3772305/906385

Nice job!

PR #1121, opened 2024-10-21
fix display choices for swagger display /admin/api/swagger/

shapiromatron:
Swagger display failed b/c the choices were a callable which converted to executable yaml, which fails in our swagger display. Fix so it's no longer a callable, which fixes the openapi rendering on `/admin/api/swagger/`


munnsmunns (approved):
Looks good!

PR #1122, opened 2024-10-22
data pivot - fix conditional formatting on numeric data

shapiromatron:
Fixed a possible regression in #993.

On data pivots, for all conditional formatting when using the "discrete-style" type of conditional formatting, always cast to a string for both the conditional formatting rule, and the value for the observation in question.

Previously, we used to cast the observation, but not the rule, which could break in some cases.  This should be much more robust, and a string comparison should always be valid for "discrete-style" conditional formatting. The other styles such as "point-size" and "point-color" need to be numeric for scaling and shading gradients, but these discrete rules can use string matches.

As an example, this will fix the control formatting for `/summary/data-pivot/assessment/100500243/DCS-IHD-case-controlcohort-ratios-water-categ-9182/`.


caseyhans (approved):
LGTM!

PR #1124, opened 2024-10-23
github publish hawc client

shapiromatron:
Automatically publish [hawc-client](https://pypi.org/project/hawc-client/) to pypi when a branch is tagged with a `hawc-client` tag.

In addition to changes in this PR:

* Create a new github environment named `publish`, and restrict to the `main` branch
* Add a secret to the `PUBLISH_PYPI` environment variable

As an example, see [this action](https://github.com/shapiromatron/hawc/actions/runs/11485959210/job/31971386217) which resulted in a complete push to test pypi; the final PR made some [changes](https://github.com/shapiromatron/hawc/pull/1124/commits/b57a8d66e9e638015bc1ed120a36b27760685f69) to limit when this can be executed, and only push from the main branch. 

---

Reference:

* https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/
* https://github.com/pypa/gh-action-pypi-publish
* https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -184,17 +184,3 @@ jobs:
         else
           exit 1
         fi
-    - name: build wheels
-      if: startsWith(github.event.ref, 'refs/tags/') || endsWith(github.event.ref, 'refs/heads/main')
-      run: |
-        python manage.py set_git_commit
-        flit build --format wheel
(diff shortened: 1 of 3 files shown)


rabstejnek (approved):
Looks good!

PR #1125, opened 2024-10-24
github action - switch to uv

shapiromatron:
Use [uv](https://docs.astral.sh/uv/). in github actions instead of pip to speed up installs. This speeds up installs by around a minute (30s for the backend job, and 30s for integration).

Also removed the wheel building and artifacts, since that is handled by the github action.


munnsmunns (approved):
Wow, incredible! That's a lot of time saved across all of our CI checks

PR #1126, opened 2024-10-28
Refactor Prisma data fetch

munnsmunns:
The main scope of this PR is to pass the data for prisma visuals in via API instead of being passed directly by a django template, to make it more consistent with other visual types.

This will be a part of a broader effort to standardize and refactor how data is passed to the visuals for rendering on the frontend to follow this pattern introduced in the PR.

A few concepts were introduced.

At the API level:

* Added new `/summary/api/assessment/:id/json_data/` endpoint to return data for create/update a visual. Must be team-member or higher.
* Added new `/summary/api/visual/:id/json_data/` endpoint to return data for read-only views of a visual. Has standard permissions tied to the visual.

At the model level, added three new methods to the `Visual` model; they are currently only implemented for the Prisma visual, but we expect that refactoring will take place to implement for other visual types in the future: 
* `get_data`: returns the data for the visual (this can be a visual saved to a database, or a visual created but not yet saved)
* `update_config`: returns configuration required to for create/update of a visual (this can be a visual saved to the database, or a visual created but not yet saved)
* `read_config`: returns configuration required to view a visual (this is intended for a visual which has been saved to the database)

In the future, we may essentially have two different templates for visuals, one for visuals which can be fully handled by django and django templates, and one for visuals that require a JS app. The work in this PR is intended to support visuals which require a JS app.


hawc/apps/summary/api.py
```diff
@@ -130,6 +130,24 @@ def data(self, request, pk):
             )
         return FlatExport.api_response(df, obj.slug)
 
+    @action(
+        detail=False, methods=("post",), action_perms=AssessmentViewSetPermissions.CAN_VIEW_OBJECT
+    )
+    def data_json(self, request):
```
shapiromatron, line 136: add a test?

hawc/apps/summary/models.py
```diff
@@ -417,6 +418,24 @@ def get_prisma_data(cls, assessment: Assessment) -> str:
             }
         )
 
+    @classmethod
+    def get_data_from_config(cls, config):
```
shapiromatron, line 422: perhaps use a pydantic schema here to ensure config matches? we could create a `schema.py` if that'd be helpful.

hawc/apps/summary/schemas.py
```diff
@@ -0,0 +1,22 @@
+from pydantic import BaseModel
+
+from . import constants
+
+
+class VisualConfig(BaseModel):
```
shapiromatron, line 6: I mean a schema for the data coming in as a request for a dataset, not the data going out. You don't have to add all fields, just the things like that may be required to get the correct dataset for filtering. In the example above, youd need an assessment which is an int, and a visual type, which is an int too. I'd keep it simple, we can always add more.

Docs on how extra is ignored: https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra

hawc/apps/summary/models.py
```diff
@@ -417,6 +418,22 @@ def get_prisma_data(cls, assessment: Assessment) -> str:
             }
         )
 
+    @classmethod
+    def get_data_from_config(cls, config: schemas.VisualConfig):
```
shapiromatron, line 422: also, we'll always know the assessment ID, external to the config. I'd take that as an input argument instead of getting it from the config; if we relied on config a bad actor could submit a different assessment in their config from a visual assessment or assessment.

hawc/apps/summary/views.py
```diff
@@ -445,7 +445,28 @@ def get_context_data(self, **kwargs):
             len(context["breadcrumbs"]) - 1, get_visual_list_crumb(self.assessment)
         )
         if context["visual_type"] == constants.VisualType.PRISMA:
-            context.update(prisma_data=models.Visual.get_prisma_data(self.assessment))
+            context.update(api_url=reverse("summary:api:visual-data-json"))
+        context["config"] = schemas.VisualConfig(
```
shapiromatron, line 449: the config that's being sent to the template is completely different than the config that's being passed to get the dataset, I would not overload the two.  For this PR, I'd focus exclusively on the config required to get a dataset.

shapiromatron (approved):
@munnsmunns  

I rewrote a large amount of this PR so that it worked for the create, update, and read views, since the work as you implemented didn't modify the read-only views, and passing the data in via API for the visuals was the purpose of this PR. 

I didn't fix for the delete view.  I'm thinking that this refactoring may be really nice for any case where we have extensive JS, but we may have a separate route for rendering and displaying visuals if it can all be handled with just standard django templates. 

Please review in detail and ensure I didn't break anything with these changes, and it's ok to merge if you're happy with it.

PR #1129, opened 2024-10-30
Move UDF table on animal group detail page

caseyhans:
Move the UDF table on the animal group detail page to above the endpoint list

image


shapiromatron (approved):
LGTM

PR #1130, opened 2024-11-02
Add label reordering

caseyhans:
Add the ability to optionally reorder labels "after" other sibling labels.
- Checks that 'parent' field matches the parent of the 'after' sibling field, if given
- Returns and refreshes entire label list on successful updates, creates, and deletes, so the list is always accurate
- Removes scroll HTMX animation--unnecessary in this case since the forms are so small, and it feels a bit annoying if you're trying to see the entire tree while creating a new label

image

### DIFF
--- a/hawc/apps/assessment/views.py
+++ b/hawc/apps/assessment/views.py
@@ -968,6 +968,10 @@ class LabelViewSet(HtmxViewSet):
     model = models.Label
     form_fragment = "assessment/fragments/label_edit_row.html"
     detail_fragment = "assessment/fragments/label_row.html"
+    list_fragment = "assessment/fragments/label_list.html"
+
+    def get_queryset(self, request):
+        return models.Label.get_assessment_qs(request.item.assessment.pk)
 
     @action(permission=can_view, htmx_only=False)
     def read(self, request: HttpRequest, *args, **kwargs):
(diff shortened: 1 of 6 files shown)


rabstejnek (approved):
Looks good!

I wonder though if it would be a good idea to add both a "Before" and "After" field, with validation that you can only provide one or the other. Technically you could do everything using "After", but if you want a fresh tag to be at the top you'd have to move all the other tags to be after it, so a tag list with 10 tags would require 9 operations to move a fresh tag to the top. This may be something we can wait on user feedback about to see if its really necessary.

hawc/apps/assessment/views.py
```diff
@@ -983,36 +987,57 @@ def read(self, request: HttpRequest, *args, **kwargs):
...
+            else:
+                retarget_form = True
         else:
             form = forms.LabelForm(assessment=request.item.assessment)
-        context = self.get_context_data(form=form)
-        return render(request, template, context)
+        context = self.get_context_data(form=form, **kwargs)
+        response = render(request, template, context)
+        if retarget_form:
+            response["HX-Retarget"] = "#label-edit-row-"
```
shapiromatron: didn't know about this, thanks, so basically on success it uses the target in the template, but if there's a form validation error it retargets and replaces itself. 

https://htmx.org/reference/

> a CSS selector that updates the target of the content update to a different element on the page

shapiromatron (approved):
Looks great. I really like the hx-retarget trick where we design the html templates for the happy case but fallback to form replacement on error, really nice work.

I learned something new today!

shapiromatron:
> Looks good!
> 
> I wonder though if it would be a good idea to add both a "Before" and "After" field, with validation that you can only provide one or the other. Technically you could do everything using "After", but if you want a fresh tag to be at the top you'd have to move all the other tags to be after it, so a tag list with 10 tags would require 9 operations to move a fresh tag to the top. This may be something we can wait on user feedback about to see if its really necessary.

I tried switching to after and it worked, but then I had a hard time with the wording for the form, I felt like the current approach was clearer to the user, even it it make require more edits. That's ok, I'd imagine that sorting is done rarely.

PR #1132, opened 2024-11-05
Update app to use Node v22

munnsmunns:
Update CI workflow file to use Node LTS version 22, and indicate in our dev docs to install that version.

Update to the newer version of Node here https://nodejs.org/en, or in CLI by running `npm install -g node@22`


rabstejnek (approved):
Looks good!

PR #1135, opened 2024-11-06
Fix 500 errors

munnsmunns:
This PR tracks down multiple 500 errors found in API endpoints:

- ObjectDoesNotExist error in study endpoints
- Catch ValueError when unable to parse DTXSID in Epi endpoints
- Error thrown when HEAD request made to lit tagtree endpoint

Other errors in the card related to the autocomplete endpoint were fixed in the latest release by #1088


hawc/apps/lit/api.py
```diff
@@ -63,7 +63,7 @@ def tagtree(self, request, pk, *args, **kwargs):
                 "Updated (tagtree replace)", assessment, assessment.id, self.request.user.id
             )
         else:
-            raise ValueError()
+            raise ValidationError(f"Request method {self.request.method} is not allowed.")
```
munnsmunns, line 66: The ValueError was being tripped by a HEAD request, which apparently is [treated like a GET request](https://http.dev/head), but won't match any if statements here. Does this seem like the right way to handle it or should we add "HEAD" to the if statement with "GET"?

shapiromatron: maybe we just add another if clause for head, and return a 200 response w/ no data?

caseyhans (approved):

PR #1136, opened 2024-11-21
fix empty database migrations

shapiromatron:
It's been a long time since we started a fresh version of hawc, and we had few issues where if you run django database migrations from scratch, it didn't work.

1. Fixed an error where a `batch_size` is now required when using prefetch_related and iterator
2. Fixed an issue where a vocab fixture attempted to load a deprecated field, before the deprecated field was created.

To check, update your `local.py` settings to use a database named `foo`, and then:

```bash
dropdb -U hawc foo
createdb -U hawc foo
python manage.py migrate
```


caseyhans (approved):
Nice! Works like a charm.

PR #1139, opened 2024-11-22
data pivot - default conditional format --- (passthrough)

shapiromatron:
In data pivots, make the default conditional format style ---, aka passthrough. Previously it defaulted to the first option.  Generally we see with conditional formatting many passthrough rules, so it makes more sense to pick that option instead of something else.

Also update the modal layout to make it more space efficient, including change the discrete values to a table, and fixing the gradient preview:

NEW:
image

OLD:
image


rabstejnek (approved):
Looks good!

PR #1140, opened 2024-11-22
Set order for label displays

caseyhans:
Applies an order to labeled items based on the label path.
Before:
image

After:
image

Tree:
image


shapiromatron:
See #1148 as a proposed alternative to this PR

PR #1142, opened 2024-11-24
fix test deprecation warnings

shapiromatron:
Running pytest shows a list of pending deprecations. This PR removes a number of the deprecation warnings with the exception of a those from pandas and a few other libaries.

Before:

`769 passed, 28 skipped, 728 warnings in 34.90s`

After: 

`769 passed, 28 skipped, 101 warnings in 33.66s`

The majority of the remaining warnings are pandas related; these will be fixed in a subsequent PR.

**References**

- https://pandas.pydata.org/docs/whatsnew/v2.2.0.html#deprecate-aliases-m-q-y-etc-in-favour-of-me-qe-ye-etc-for-offsets
- https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.html.format_html


rabstejnek (approved):
600 down, 100 more to go! Looks good!

PR #1143, opened 2024-12-02
global assessment search

shapiromatron:
We've long had a request to be able to search globally in HAWC across assessments. This PR adds that ability by creating a new `/search/` page. 

* You can currently search for studies, visuals, or data pivots. In the future, we could add references and summary tables. In the future we should merge visuals and data pivots, because  conceptually there shouldn't really be a distinction there. I don't think it's worth trying to do it now, until the database models are merged, then it'll be much simpler.
* If a user is not authenticated, then you can search all assessments (published, visible), or you can select some assessments. 
* If a user is authenticated, you can search for all internal assessments, or you can select a subset

For visuals and data pivots, you can also search for assessment labels.

image

With this initial release, we didn't add a link anywhere else on the website that links to this page; you have to know that it exists. After we get a little more feedback from users, we will add links to the page so you can get there from elsewhere.


rabstejnek (approved):
Looks good!

caseyhans (approved):
LGTM

PR #1144, opened 2024-12-03
tagtree - show references on root node

shapiromatron:
Previously we didn't show references on the root node, now that we added pagination #426, show on the root node as well.

URLS to check:

- `/lit/assessment/:assessment-id/references/visualization/`
- `/summary/visual/assessment/:assessment-id/:visual-id/` (custom tagtree visual)


rabstejnek (approved):
Good change, especially with the custom visuals maybe having a root that users would be interested in inspecting. Tested it out with no problem!

PR #1145, opened 2024-12-03
implement bmds modeling using pybmds

shapiromatron:
Implement pybmds in hawc with latest release of version [24.1](https://pypi.org/project/pybmds/).

This initial release doesn't have all the customization that's available in bmdsonline.epa.gov, and hopefully it won't need that level of customization. That said, the execution of bmds models is currently configured to only be available if the user is in our `beta tester` group. Individual models cannot be selected; by default we run all standard models with the standard parameter restrictions. Users can modify the BMR type and value, the dose units to use, and, the number of doses to drop (starting with the highest dose).  We hope to get feedback from a small pool of users initially before expanding more broadly.

You can currently create a BMD analysis for an endpoint, view existing analyses (you may want to have multiple for different dose units), update an existing analysis, or delete an analysis. You can also create a Microsoft Word report documenting the analysis. To access the BMD modeling, you'll navigate to an `animal.Endpoint` instance that has dose-response data extracted so it's available to conduct BMD modeling, and under the actions menu there's an option to create a BMD analysis (again, for now, if you're an admin or `beta tester`).

More detailed changes include:

* Remove `bmds` package and some of the dll requirements for execution, switch to `pybmds` package.
* Rewrite all `bmds3` javascript components from scratch (around 1000 LOC). There weren't any `bmds3` analyses in production systems of HAWC to our knowledge, so it was safe to rewrite from scratch without any backwards compatibility (preserving compatibility with `bmds2` was needed however).
    * Use new plotly components for rendering the dose response and curve. We used the existing d3 based views on the `animal.Endpoint` pages, but this may change in the future. The plotly plots, borrowed heavily from BMDS Online, are easier to develop and maintain, and objectively look better.
* Remove `bmd.AssessmentSettings`. Previously you'd select a particular version of BMDS and all runs in your assessment would use that version, and recommendation logic would be set globally for that assessment. This has all been removed; HAWC will now only run the latest version of pybmds, so users will need to manage and potentially rerun analyses if they need them consistent for an assessment. Recommendation logic is no longer customizable.
* Added new `bmd.Session` api to return a Word report.
* Remove `FeatureFlags.ENABLE_BMDS_33`
* Update test fixture to remove BMDS 3.3 examples and use pybmds 24.1 examples
* Add integration tests to ensure you can view a bmds 2.7 run, and execute a continuous and dichotomous analysis in pybmds

In addition (tangentially related changes):

* Bump test coverage minimum from 91.8% to 92.0%. Our current test suite is around 92.2% covered in CI.
* Added a new javascript `WaitLoad` React component that delays rendering for a specified time. This was used for Plotly plots to make sure the rest of the page is laid out so when the plot renders, it uses 100% of the height/width. Using a timing delay isn't ideal, but it seemed to work well for most situations tested and could be improved in the future if it proves to be an issue.
* Fixed a bug with the SerializerHelper in testing where in some cases tests would fail b/c models may not have been loaded and registered during unit tests.
* Fix typing annotation to remove pydantic constrained int in favor of using Annotated Fields (fix type hinting in VSCode)

## Screenshots

**Settings:**
image

**Results:**
image
image

**Individual Model Result Modal:**
image


caseyhans (approved):
Ok, I poked around quite a bit and it all seemed to work well! I like it. I did a minor margin and alignment tweak for an alert, but that was it. I'd like it if there was a loading spinner when you click 'save selection', but it's not a big deal and works well as-is. This is a big one, great stuff!

rabstejnek (approved):
Looks good! I had some old sessions too that I checked out, and it looks like the backwards compatibility works well!

PR #1148, opened 2024-12-09
Set order for label displays (2)

rabstejnek:
Applies an order to labeled items based on the label path.

This applies the ordering at the model level, which I find less intrusive then the solution at #1140 .


caseyhans (approved):
Great find, works for me! #1140 can be deleting pending final approval of this change

PR #1153, opened 2025-01-08
python package updates

shapiromatron:
Quarterly python package update. Most packages updated without any issues. There were a few modifications that were made in addition:

* With `ruff`, a few new formatting rules were applied and configured. This includes automated formatting of ipynb notebooks, which was new.
* With upgrading `numpy` to v2, when numpy values are printed, they now show things like `np.int64(-1)` where they previously just showed -1, This can make serializer error messages look strange. I fixed by changing the print format to use the older version, https://numpy.org/doc/2.1/reference/generated/numpy.set_printoptions.html#
* I updated `plotly` in python and javascript. The javascript update made a number of other changes to the yarn lock file, likely because I got a new computer and the stability of lockfiles is a known issue w/ yarn v1


rabstejnek (approved):
Don't see any issues!

PR #1158, opened 2025-01-17
Increase lit test coverage

caseyhans:
- Increase test coverage for the lit module:
  - Test reference filterset 
  - Test Workflows
- Fix labeled items search bug
- Fix invalid literature workflow in the fixture
- Add test for serializer FilterReferences


shapiromatron (approved):
Looks good. I added a test for the `PydanticDrfSerializer` to better understand the changes you made to that class, I guess we should have had tests there in the first place.

PR #1162, opened 2025-01-28
rewrite visual forms using react instead of jQuery

shapiromatron:
This PR should be reviewed after #1114

There were a few legacy jQuery applications for building custom visuals in HAWC. These user-interfaces did not perform well and had some bugs, but because they were so complicated, we never were able to fix them and they remained in this painful state. Because of the inability to change the applications, we had a really complicated django view and form layout too, since we had multiple different approaches depending on the frontend technology being used. This PR removes the jQuery applications and fully switch to a react frontend. This should dramatically improve stability of the application.

This rewrites the form frontend for Crossview, RobHeatmap, and RobBarchart. 

In addition:

* Refactor and standardize visual form startup javascript
* Rewrite how the django-form panel is being used in React. Instead of recreating the django form with React, we have move the form in the react tab, and then out of the react tab, so that we're not duplicating state.
* Rewrite the dose units selector widget to use React instead of the jQuery custom solution. See `PreferredWidget`.
* Added tests to improve test coverage for the diversity of visual types and evidence types
* Remove `VisualizationCreateTester`
* Refactor how data for preview of visualizations is generated and passed to the application.


caseyhans (changes requested):
I'm having trouble getting the Crossview form to work. I'm looking at the fixture crossview (summary/visual/assessment/2/crossview/) and once I navigate to Settings > Crossview Settings or Settings > References, the frontend crashes; looks like it's trying to call to_string on a null integer value.

Opening the Settings > Legend Settings tab on the RoB barchart (summary/visual/assessment/2/rob-barchart/update/) also produces a failed prop type error on a checkbox input. Checking 'Show NR in Legend' produces an error re: controlled/uncontrolled inputs. No such issues in the RoB heatmap.

The first one is my main concern. Let me know if you have trouble reproducing these errors!

rabstejnek (commented):
Looks good. I did not find any additional problems in this PR aside from the ones Casey identified (which I was able to reproduce). I did notice though that the smart tag form is not working correctly, but this bug seems to predate this PR. It may be worth fixing in this PR though if its a simple fix (looks like maybe jQuery just needs to be imported before select2).

shapiromatron:
Great comments @caseyhans! Re-requesting review based on your feedback. In addition to fixing the issues you identified:

* 7f476c27e fix bug in crossview configuration name (restore original name "column -> columns")
* f7cf91ce7 update existing data in fixture
* 467e5e238 migrate data where previously we allowed for null ints, now we set default values
* 546629fae fix logic error where a select may have no options if there are no data
* e311f106d fix tests

PR #1167, opened 2025-02-07
CATS Views

caseyhans:
Basic implementation of CATS views and forms. This PR creates the basic views and forms, but does not implement full modeling functionality.

To do:
- Consider permissions/visibility beyond staff
- Consider moving models/views to their own app/somewhere else than lit (since many of these models are site-wide)
- Run model in task
- Add htmx form for user to manually approve predictions
- Consider security implications of loading random uploaded pickled files
- Tests


caseyhans:
OK, it's technically ready for review, but this should wait to be merged until I can get in model functionality

